PR #5, opened 2020-10-06
Add option to not verify ssl

daviddavis:
fixes #7658


bmbouter (approved):
I tested this and it works great for me!

mdellweg (approved):
So this was a default true flag that could be set?
I agree that was rather stupid.

PR #6, opened 2020-10-07
Have the CI run every night

daviddavis:
I figure that changes in pulpcore and plugins could potentially break our code so it's probably a good idea to run every night.

[noissue]

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '15 3 * * *'
 
 env:
   COLORTERM: 'yes'


ipanova (approved):

.github/workflows/main.yml
```diff
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull:
+  schedule:
+    - cron: '15 3 * * *'
```
ipanova, line 7: :+1:

ipanova: if you would like to be super picky we can exclude sat and sun
'15 3 * * 1,2,3,4,5' i think this is the syntax

mdellweg (approved):
At some point we might want to test against containers of different versions of pulp in the "nightlies". But ATM this seems to be just about right.
Thank you!

PR #9, opened 2020-10-08
Use docker when running on Github Actions

daviddavis:
Currently, podman on Github Actions is using vfs which is very
inefficient. Use docker instead until we figure out how to get
fuse-overlayfs set up.


daviddavis:
This seems to also be faster: 3 min vs 8 min.

mdellweg (approved):
As much as I preferred using (not docker), lets do it that way.
Do you think calling it ’CONTAINER_RUNTIME’ would be better?

PR #11, opened 2020-10-12
Add support for configuring the CLI

daviddavis:
fixes #6607

Tasks
- [x] Add .netrc information
- [x] Add tests

### DIFF
--- a/pulpcore/cli/common.py
+++ b/pulpcore/cli/common.py
@@ -1,7 +1,10 @@
 import click
 import datetime
 import json
+import os
 import time
+
+import toml
 
 from pulpcore.cli.openapi import OpenAPI
 
@@ -45,12 +48,35 @@ class PulpContext:
         raise click.ClickException("Task timed out")
 
 
+def _config_callback(ctx, param, value):
+    if ctx.default_map:
+        return
+
+    if value:
+        ctx.default_map = toml.load(value)["cli"]
+    else:
+        xdg_config_home = os.path.expanduser(os.environ.get("XDG_CACHE_HOME") or "~/.config")
+        default_config_path = os.path.join(xdg_config_home, "pulp", "settings.toml")
+
+        try:
+            ctx.default_map = toml.load(default_config_path)["cli"]
+        except FileNotFoundError:
+            pass
+
+
 @click.group()
 @click.version_option(prog_name="pulp3 command line interface")
+@click.option(
+    "--config",
+    type=click.Path(resolve_path=True),
+    help="Path to the Pulp settings.toml file",
+    callback=_config_callback,
+    expose_value=False,
+)
 @click.option("--base-url", default="https://localhost", help="Api base url")
-@click.option("--user", default="admin", help="Username on pulp server")
+@click.option("--user", help="Username on pulp server")
 # @click.option("--password", prompt=True, hide_input=True, help="Password on pulp server")
-@click.option("--password", default="password", help="Password on pulp server")
+@click.option("--password", help="Password on pulp server")
 @click.option("--verify-ssl/--no-verify-ssl", default=True, help="Verify SSL connection")
 @click.option("--format", type=click.Choice(["json"], case_sensitive=False), default="json")
 @click.pass_context
--- a/pulpcore/cli/openapi.py
+++ b/pulpcore/cli/openapi.py
@@ -32,7 +32,12 @@ class OpenAPI:
             "Accept": "application/json",
         }
         self._session = requests.session()
-        self._session.auth = (username, password)
+        if username and password:
+            self._session.auth = (username, password)
+        elif username:
+            raise Exception("Password is required if username is set.")
+        elif password:
+            raise Exception("Username is required if password is set.")
         self._session.headers.update(headers)
         self._session.verify = validate_certs
 
--- a/tests/scripts/test_config.sh
+++ b/tests/scripts/test_config.sh
@@ -1,0 +1,47 @@
+#!/bin/sh
+
+. "$(dirname "$(realpath $0)")/config.sh"
+
+good_settings="good_settings.toml"
+bad_settings="bad_settings.toml"
+good_cli="$PULP_CLI --config $good_settings"
+bad_cli="$PULP_CLI --config $bad_settings"
+
+cleanup() {
+  rm $good_settings || true
+  rm $bad_settings || true
+}
+trap cleanup EXIT
+
+cat >$good_settings <<EOL
+[cli]
+base_url = "$PULP_BASE_URL"
+user = "$PULP_USER"
+password = "$PULP_PASSWORD"
+EOL
+
+cat >$bad_settings <<EOL
+[cli]
+base_url = "http://badurl"
+user = "$PULP_USER"
+password = "$PULP_PASSWORD"
+EOL
+
+$good_cli file repository list
+$bad_cli --base-url $PULP_BASE_URL file repository list
+
+if $bad_cli file repository list 2> /dev/null; then
+  echo "'$bad_cli ...' succeeded unexpectedly"
+  exit 1
+fi
+
+if $good_cli --base-url "http://badurl" file repository list 2> /dev/null; then
+  echo "'$good_cli --base-url http://badurl ...' succeeded unexpectedly"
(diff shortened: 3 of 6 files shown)


pulpcore/cli/common.py
```diff
@@ -47,25 +71,24 @@ def wait_for_task(self, task_href, timeout=120):
...
+@click.option("--format", type=click.Choice(["json"], case_sensitive=False))
 @click.pass_context
-def main(ctx, base_url, user, password, verify_ssl, format):
+def main(ctx, config, base_url, user, password, verify_ssl, format):
     ctx.ensure_object(PulpContext)
+
+    ctx.obj.config = PulpConfig(config)
     ctx.obj.api = OpenAPI(
-        base_url=base_url,
+        base_url=(base_url or ctx.obj.config.base_url or "https://localhost"),
```
mdellweg, line 87: I would prefer to see this logic in the Config object.
If we provide overrides to the __init__ and provide defaults there, too, it's just `ctx.obj.config.base_url` here.

pulpcore/cli/common.py
```diff
@@ -47,25 +71,24 @@ def wait_for_task(self, task_href, timeout=120):
...
         doc_path="/pulp/api/v3/docs/api.json",
         username=user,
```
mdellweg, line 89: The name should also be part of the config, wdyt?
from name and host, netrc can provide the password

mdellweg (commented):
I like the approach.
As said, i would put all the config option mangling into the config object.

pulpcore/cli/common.py
daviddavis: I couldn't figure out how to get this to work. I tried to set auth on the requests session a few different ways (`self._session.auth = (username, None)`, `self._session.auth = (username, '')`, etc) but it seems that the only way to have it use `.netrc` is not to set `_auth` at all so I am not using defaults for username and password and that seems to work.

pulpcore/cli/config.py
```diff
@@ -0,0 +1,32 @@
+import os
+
+import click
+import toml
+
+DEFAULT_PATH = os.path.expanduser("~/.config/pulp/settings.toml")
```
mdellweg, line 6: Can we try `$XDG_CONFIG_HOME` first?

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

pulpcore/cli/config.py
```diff
@@ -0,0 +1,35 @@
...
+
+def config_file():
+    def decorator(f):
+        options = {
+            "type": click.Path(resolve_path=True),
+            "help": "Path to the toml config file",
+            "callback": _config_callback,
+            "expose_value": False,
+        }
+        return click.option("--config", **options)(f)
```
mdellweg, line 33: Why assemble the options into a dict instead of just passing them?
Also something like
`config_file = click.option("--config", ...)`
should be the decorator you are looking for.

daviddavis: Originally I was going to add options to config_file(). But yes your suggestion would be simpler. My other thought though is perhaps we don’t even need a decorator. Maybe I should just add a config option to main. Wdyt?

daviddavis: I've updated this PR to not use a decorator but am happy to define the `config_file` decorator if you think that's a better solution.

mdellweg: Just adding config_file to main seems to be the most simple solution.

pulpcore/cli/common.py
```diff
@@ -45,19 +48,42 @@ def wait_for_task(self, task_href, timeout=120):
         raise click.ClickException("Task timed out")
 
 
+def _config_callback(ctx, param, value):
+    xdg_config_home = os.path.expanduser(os.environ.get("XDG_CACHE_HOME") or "~/.config")
```
mdellweg, line 52: Will this only be called if `--config` is specified?

pulpcore/cli/common.py
```diff
@@ -45,19 +48,42 @@ def wait_for_task(self, task_href, timeout=120):
...
-@click.option("--password", default="password", help="Password on pulp server")
+@click.option("--password", help="Password on pulp server")
 @click.option("--verify-ssl/--no-verify-ssl", default=True, help="Verify SSL connection")
 @click.option("--format", type=click.Choice(["json"], case_sensitive=False), default="json")
 @click.pass_context
 def main(ctx, base_url, user, password, verify_ssl, format):
     ctx.ensure_object(PulpContext)
     ctx.obj.api = OpenAPI(
-        base_url=base_url,
+        base_url=(base_url),
```
mdellweg, line 86: Why is this change?

pulpcore/cli/openapi.py
```diff
@@ -32,7 +32,8 @@ def __init__(
             "Accept": "application/json",
         }
         self._session = requests.session()
-        self._session.auth = (username, password)
+        if username or password:
```
mdellweg, line 35:
```suggestion
        if username and password:
```

pulpcore/cli/common.py
daviddavis: It's always called. I'll move it down to the `else`.

pulpcore/cli/openapi.py
daviddavis: My only concern here is that this will silence errors in situations where a user accidentally defines username or password but not both. wdyt?

pulpcore/cli/common.py
mdellweg: Sorry, i meant the whole function. And i was wondering if the defaults from the default location are read if you didn't use the `--config` parameter.

pulpcore/cli/openapi.py
mdellweg: Then we need more if/else.
If either of them is None and the other one isn't, MyPy complains.

pulpcore/cli/common.py
daviddavis: Ah yea, that's correct. This is always called and if there's no value, it'll use the default location.

pulpcore/cli/openapi.py
daviddavis: In that case, I can switch it to `and`. We should probably add this validation elsewhere.

daviddavis:
@mdellweg added some validation around username and password. openapi.py seemed like the logical place. also added a test.

pulpcore/cli/openapi.py
```diff
@@ -32,7 +32,12 @@ def __init__(
             "Accept": "application/json",
         }
         self._session = requests.session()
-        self._session.auth = (username, password)
+        if username and password:
+            self._session.auth = (username, password)
+        elif username and not password:
+            raise Exception("Password is required if username is set.")
+        elif password and not username:
```
mdellweg, line 39:
```suggestion
        elif password:
```
Technically we don't need it...

pulpcore/cli/openapi.py
```diff
@@ -32,7 +32,12 @@ def __init__(
             "Accept": "application/json",
         }
         self._session = requests.session()
-        self._session.auth = (username, password)
+        if username and password:
+            self._session.auth = (username, password)
+        elif username and not password:
```
mdellweg, line 37:
```suggestion
        elif username:
```

tests/scripts/test_config.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+. "$(dirname "$(realpath $0)")/config.sh"
+
+good_settings="good_settings.toml"
+bad_settings="bad_settings.toml"
+good_cli="$PULP_CLI --config $good_settings"
```
mdellweg, line 7: We can now provide the settings via giving a `XDG_CONFIG_HOME` pointing in the right place.
So we can get rid of this `$PULP_CLI` variable. WDYT?

daviddavis: I don't know. Right now my pain point is that I have to supply `--base_url` every time I use the CLI since this PR isn't merged.

daviddavis: I created a task for your suggestion: https://pulp.plan.io/issues/7727

mdellweg (approved):

PR #15, opened 2020-10-18
Add type annotations

mdellweg:
Started to add type annotations.

This already revealed small inconsistencies (mostly around str vs bytes) and one actual bug (the `super` call).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ options include:
 ## Configuration
 
 The CLI can be configured by using a toml file.
-By default the location of this file is `~/.config/pulp/settings.toml`. 
+By default the location of this file is `~/.config/pulp/settings.toml`.
 However, this can be customized by using the `--config` option.
 Any settings supplied as options to a command will override these settings.
 
@@ -64,3 +64,8 @@ password password
 Tests are run using `pytest`.
 
 Tests are shell scripts in `tests/scripts` with names like `test_*.sh`.
+
+## Contributing
+
+`pulp-cli` comes with python type annotations and black code formatting.
+To verify your code please run `black`, `flake8` and `mypy`.
--- a/pulpcore/cli/common.py
+++ b/pulpcore/cli/common.py
@@ -1,3 +1,5 @@
+from typing import Any, Optional
+
 import click
 import datetime
 import json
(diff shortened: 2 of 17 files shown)


daviddavis:
I was just thinking about doing this.

mdellweg:
It's incompatible with python2. Is that a problem?

Well, it could be done with py2, but it's a mess.

pulpcore/cli/common.py
```diff
@@ -7,19 +9,27 @@
 
 
 class PulpJSONEncoder(json.JSONEncoder):
-    def default(self, obj):
+    def default(self, obj: Any) -> str:
         if isinstance(obj, datetime.datetime):
             return obj.isoformat()
         else:
-            super(json.JSONEncoder).default(obj)
+            return super().default(obj)
```
daviddavis, line 16: Nice that you were able to catch this using typing hints.

daviddavis:
> It's incompatible with python2. Is that a problem?

That's a good question. I don't think it is but let me confirm.

mdellweg:
now with `--strict`. :grinning:

mdellweg:
> > It's incompatible with python2. Is that a problem?
> 
> That's a good question. I don't think it is but let me confirm.

FWIW We don't support py2 anyway to this point:

https://github.com/pulp/pulp-cli/blob/c7ca274c947015064d295d6fc2dc299c055c4b93/.github/workflows/main.yml#L23

README.md
```diff
@@ -15,7 +15,7 @@ options include:
 ## Configuration
 
 The CLI can be configured by using a toml file.
-By default the location of this file is `~/.config/pulp/settings.toml`. 
+By default the location of this file is `~/.config/pulp/settings.toml`.
```
daviddavis, line 18: :+1:

pulpcore/cli/core/artifact.py
```diff
@@ -1,29 +1,32 @@
...
     result = ctx.obj.call("artifacts_list", parameters=params)
     ctx.obj.output_result(result)
 
 
 @artifact.command()
 @click.option("--file", type=click.File("rb"), required=True)
-@click.option("--chunk-size", default=1000000, help="Chunk size in bytes (default is 1 MB)")
+@click.option(
+    "--chunk-size", default=1000000, type=int, help="Chunk size in bytes (default is 1 MB)"
+)
```
daviddavis, line 27: Why the change here?

pulpcore/cli/core_cli.py
```diff
@@ -1,16 +1,17 @@
...
 
 @main.command()
 @click.pass_context
-def status(ctx):
-    result = ctx.obj.call("status_read")
-    ctx.obj.output_result(result)
+def status(ctx: click.Context) -> None:
+    pulp_ctx: PulpContext = ctx.find_object(PulpContext)
+    result = pulp_ctx.call("status_read")
+    pulp_ctx.output_result(result)
```
daviddavis, line 14: I'm curious why this changed as compared to say `orphans_delete()`?

pulpcore/cli/common.py
```diff
@@ -42,13 +55,11 @@ def wait_for_task(self, task_href, timeout=120):
...
             if task["state"] == "canceled":
                 raise click.ClickException("Task canceled")
-            time.sleep(1)
-            timeout -= 1
             click.echo(".", nl=False, err=True)
         raise click.ClickException("Task timed out")
 
 
-def _config_callback(ctx, param, value):
+def _config_callback(ctx: click.Context, param: Any, value: str) -> None:
```
daviddavis, line 62: I think param could maybe be a str but I guess it doesn't make a huge difference since we're not using it.

pulpcore/cli/core/artifact.py
mdellweg: It is adding `type=int` and i hope that will add input validation on the click side.
Not sure if the decorator will add type annotations to the variable passed to the command, but that would be super cool.

pulpcore/cli/core_cli.py
mdellweg: I didn't know about `find_object` but it looks like you can embed more than one object in the context that way.
I think, this is more explicit and forgot about `orphan_delete` it seems.

pulpcore/cli/common.py
mdellweg: I think, `str` didn't work...

daviddavis (approved):

PR #16, opened 2020-10-22
Provide api calls as debugging output

mdellweg:
fixes #7729
https://pulp.plan.io/issues/7729

~Just a first draft. It should probably go to stderr and may be coloured?~
Edit: It is on stderr and bold now.

### DIFF
--- a/pulpcore/cli/common.py
+++ b/pulpcore/cli/common.py
@@ -32,6 +32,8 @@ class PulpContext:
     def output_result(self, result: Any) -> None:
         if self.format == "json":
             click.echo(json.dumps(result, cls=PulpJSONEncoder, indent=2))
+        elif self.format == "none":
+            pass
         else:
             raise NotImplementedError(f"Format '{self.format}' not implemented.")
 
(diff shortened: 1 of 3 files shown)


pulpcore/cli/common.py
```diff
@@ -75,12 +75,12 @@ def _config_callback(ctx, param, value):
 )
 @click.option("--base-url", default="https://localhost", help="Api base url")
 @click.option("--user", help="Username on pulp server")
-# @click.option("--password", prompt=True, hide_input=True, help="Password on pulp server")
 @click.option("--password", help="Password on pulp server")
 @click.option("--verify-ssl/--no-verify-ssl", default=True, help="Verify SSL connection")
-@click.option("--format", type=click.Choice(["json"], case_sensitive=False), default="json")
+@click.option("--format", type=click.Choice(["json", "none"], case_sensitive=False), default="json")
```
daviddavis, line 80: What does `none` do here?

daviddavis:
What about maybe having a verbosity flag?

https://click.palletsprojects.com/en/7.x/options/#counting

pulpcore/cli/common.py
mdellweg: Oh, is't the best feature you get for 8 added characters. Nothing. (It supresses normal output instead of json-rendering it.)

mdellweg:
> What about maybe having a verbosity flag?
> 
> https://click.palletsprojects.com/en/7.x/options/#counting

Instead of debug? Sure.

daviddavis (approved):
Tested and this works great :+1:

daviddavis (approved):

PR #17, opened 2020-10-23
Prompt for a password if not provided

mdellweg:


mdellweg:
@daviddavis Some people might not want to put their password in files...

If we can get a good (restful) handle on the login/auth endpoint, we might even come up with a login mechanism.

daviddavis (approved):

PR #18, opened 2020-10-24
Add command to show and join tasks

mdellweg:


daviddavis:
This code change looks good to me. I might suggest another word other than "join". Maybe "watch", "monitor", "wait", etc?

mdellweg:
We use "wait_for_task" in several places. We should just stick to it. I'll call it `--wait`.
("join" was taken from threads. And in the context of weaving it seems appropriate...)

tests/scripts/test_task.sh
```diff
@@ -10,14 +10,22 @@ cleanup() {
...
+# Test canceling a task
 task=$(pulp_cli file repository sync --background --name "cli_test_file_repository" 2>&1 | egrep -o "/pulp/api/v3/tasks/[-[:xdigit:]]*/")
 pulp_cli task cancel --href "$task"
-test $(pulp_cli task list --name $sync_task --state canceled | jq -r .count) -eq $(($count + 1))
+test $(pulp_cli task list --name $sync_task --state canceled | jq -r length) -eq $(($count + 1))
+test $(pulp_cli task show --href $task | jq -r '.state') = "canceled"
+
+# Test waiting for a task
+task=$(pulp_cli file repository sync --background --name "cli_test_file_repository" 2>&1 | egrep -o "/pulp/api/v3/tasks/[-[:xdigit:]]*/")
+test $(pulp_cli task show --wait --href $task | jq -r '.state') = "completed"
```
mdellweg, line 27: Not quite sure if this is a good idea, because this sync task takes forever...

daviddavis:
I checked out this PR and tested it by running "pulp task list" which ran for a couple minutes and made something like 30 calls to my pulp server. Maybe we should set a default limit?

tests/scripts/test_task.sh
daviddavis: I tried testing this on my system but I am getting a strange error:

```
sync_task="pulp_file.app.tasks.synchronizing.synchronize"                                                                                                                           
count=$(pulp_cli task list --name $sync_task --state canceled | jq -r length)                                                                                                       
parse error: Invalid numeric literal at line 1, column 2                                                                                                                            
```

daviddavis: Actually, I had my format set to "yaml". Need to figure out how to handle settings for tests.

I ran this locally and the entire test took less than one min so I think that's ok.

mdellweg:
Sure, what do you recommend? 10 calls in batches of 10?
And a warning if there are more?

pulpcore/cli/core/task.py
```diff
@@ -1,41 +1,99 @@
...
+                limit = 0
+            result: Dict[str, Any] = self.pulp_ctx.call("tasks_list", parameters=payload)
+            count = result["count"]
+            entities.extend(result["results"])
+            if result["next"] is None:
+                break
+            payload["offset"] += payload["limit"]
+        else:
+            click.echo(f"Not all {count} entries were shown.", err=True)
+        return entities
```
mdellweg, line 32: As you can see, this code is (hopefully) generic enough to be reused by other entities.

bmbouter:
> I checked out this PR and tested it by running "pulp task list" which ran for a couple minutes and made something like 30 calls to my pulp server. Maybe we should set a default limit?

If there was a limit would it be like a page? How would "older ones" be listed?

mdellweg:
We could add an offset option too.

bmbouter:
I'm not sure, but if we're limiting the number of responses we need a way for users to see those too. Whatever you all work out is ok w/ me. I'm also ok with "all of them" being listed even though it could be a tremendous number of calls.

pulpcore/cli/common.py
```diff
@@ -76,16 +75,53 @@ def wait_for_task(self, task_href: str, timeout: int = 120) -> Any:
...
+                click.echo(".", nl=False, err=True)
+            else:
+                raise NotImplementedError(f"Unknown task state: {task['state']}")
         raise click.ClickException("Task timed out")
 
 
+class PulpEntityContext:
+    HREF: str = "entity"
+    LIST_ID: str = "tasks_list"
+    READ_ID: str = "tasks_read"
```
daviddavis, line 94: Did you mean to hardcode these?

mdellweg: No, but i needed to provide something.

mdellweg: Should have been "entities_list" which will produce an error. I would have liked to make it an abstract class.

daviddavis (approved):

PR #20, opened 2020-10-24
Adding shellcheck and cleaning up code

daviddavis:
[noissue]


.ci/run_container.sh
```diff
@@ -15,18 +15,17 @@ fi
...
-for i in $(seq 10)
+for _ in $(seq 10)
 do
   sleep 3
   if curl --fail http://localhost:8080/pulp/api/v3/status/ > /dev/null 2>&1
   then
     echo "SUCCESS."
     break
   fi
-  echo -n "."
```
daviddavis, line 26: Wasn't sure what to do about this line. Here's the error: `SC2039: In POSIX sh, echo flags are not supported.`.

.github/workflows/main.yml
```diff
@@ -29,8 +29,10 @@ jobs:
       - name: Install Test Dependencies
         run: |
           pip install black pytest flake8 mypy .
+          sudo apt-get install shellcheck
       - name: Lint code
         run: |
+          find . -name '*.sh' -print0 | xargs -0 shellcheck -e SC1090
```
daviddavis, line 35: More info about this check https://github.com/koalaman/shellcheck/wiki/SC1090

daviddavis: I skipped it mostly because it's a pain to fix. I'm not opposed to checking though.

.ci/run_container.sh
mdellweg: No problem. It's for nice output only.

.ci/run_container.sh
```diff
@@ -15,18 +15,17 @@ fi
...
 done
 
-trap "${CONTAINER_RUNTIME} stop pulp" EXIT
+trap '${CONTAINER_RUNTIME} stop pulp' EXIT
```
mdellweg, line 28: Was this really problematic? We knew the CONTAINER_RUNTIME at installing the trap. This makes it lazy evaluated, i think.

tests/scripts/config.sh.example
```diff
@@ -1,8 +1,10 @@
+#!/bin/sh
```
mdellweg, line 1: Nope.
This file is meant to be sourced.
It should not be executable to begin with. And hence no shebang.

tests/scripts/config.sh.example
```diff
@@ -1,8 +1,10 @@
...
+
 # these variables must be defined
-PULP_CLI="pulp"
-PULP_BASE_URL="http://localhost:8080"
-PULP_USER="admin"
-PULP_PASSWORD="password"
+export PULP_CLI="pulp"
+export PULP_BASE_URL="http://localhost:8080"
+export PULP_USER="admin"
+export PULP_PASSWORD="password"
```
mdellweg, line 7: Also, here it was intended to _not_ export the variables. We don't want the cli to load the values from the environment (yet).

tests/scripts/test_artifact.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-. "$(dirname "$(realpath $0)")/config.sh"
+. "$(dirname "$(realpath "$0")")/config.sh"
```
mdellweg, line 3: :+1:

tests/scripts/test_task.sh
```diff
@@ -16,8 +16,8 @@ pulp_cli file remote create --name "cli_test_file_remote" \
   --url "https://fixtures.pulpproject.org/file-large/PULP_MANIFEST"
 pulp_cli file repository create --name "cli_test_file_repository" --remote "cli_test_file_remote"
 
-task=$(pulp_cli file repository sync --background --name "cli_test_file_repository" 2>&1 | egrep -o "/pulp/api/v3/tasks/[-[:xdigit:]]*/")
+task=$(pulp_cli file repository sync --background --name "cli_test_file_repository" 2>&1 | grep -E -o "/pulp/api/v3/tasks/[-[:xdigit:]]*/")
 pulp_cli task cancel --href "$task"
-test $(pulp_cli task list --name $sync_task --state canceled | jq -r .count) -eq $(($count + 1))
+test "$(pulp_cli task list --name $sync_task --state canceled | jq -r .count)" -eq $((count + 1))
```
mdellweg, line 21: Nice find! This was an actual bug i believe?

mdellweg (changes requested):
At one point shellceck seems to be too clever. Maybe we should not name the include files *.sh.

tests/scripts/test_task.sh
daviddavis: The use of $count? Here was the warning:

```
In ./tests/scripts/test_task.sh line 21:
test $(pulp_cli task list --name $sync_task --state canceled | jq -r .count) -eq $(($count + 1))
     ^-- SC2046: Quote this to prevent word splitting.
                                                                                    ^----^ SC2004: $/${} is unnecessary on arithmetic variables.
```

tests/scripts/constants.sh
```diff
@@ -1 +1,3 @@
-FILE_REMOTE_URL="https://fixtures.pulpproject.org/file/PULP_MANIFEST"
+#!/bin/sh
+
+export FILE_REMOTE_URL="https://fixtures.pulpproject.org/file/PULP_MANIFEST"
```
mdellweg, line 3: I think, the same as for config.sh applies here.

tests/scripts/config.sh.example
```diff
@@ -1,3 +1,5 @@
+# shellcheck disable=SC2148
```
mdellweg, line 1: :+1:

mdellweg:
With the sourced files, why don't we apply this pattern and rename `config.sh` and `constants.sh` to `config.source` and `constants.source` ?

https://stackoverflow.com/questions/58587756/why-does-shellcheck-fail-when-a-source-file-is-representing-the-variables-at-the

mdellweg (approved):

PR #21, opened 2020-10-25
Improve error reporting from openapi calls

mdellweg:
In case the library can identify the error, there is no use in dumping a
complete stack trace. Introduced OpenAPIError exception.

### DIFF
--- a/pulpcore/cli/common.py
+++ b/pulpcore/cli/common.py
@@ -8,7 +8,7 @@ import time
 
 import toml
 
-from pulpcore.cli.openapi import OpenAPI
+from pulpcore.cli.openapi import OpenAPI, OpenAPIError
 
 
 class PulpJSONEncoder(json.JSONEncoder):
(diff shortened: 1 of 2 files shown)


daviddavis (approved):

daviddavis:
FYI, I started looking at the failures. They're also happening on develop. I couldn't reproduce locally though with my dev server. I think there might be something in the single container that's causing the failures...?

mdellweg:
I added more error output on failed tasks. So in a way, we were lucky here. XD
https://github.com/pulp/pulp-cli/pull/21/checks?check_run_id=1304812030#step:6:224

daviddavis:
Here we go: 

https://github.com/aio-libs/aiohttp/issues/5110 
https://github.com/aio-libs/aiohttp/pull/5118

daviddavis:
I opened https://github.com/pulp/pulpcore/pull/984 so we're not completely unpinned.

I also kicked off a new pulp-oci-images publish job.

mdellweg:
So it has been fixed upstream... ?!

PR #23, opened 2020-10-28
Refactor

mdellweg:


mdellweg:
Let's get this out of draft. Sure there are more places to adjust to the new concepts, i experimented with here, but i also want feedback and we can care for them later. ;)

daviddavis (approved):
:+1:

PR #27, opened 2020-11-04
Add container commands

mdellweg:


mypy.ini
```diff
@@ -1,5 +1,7 @@
 [mypy]
 strict = True
+show_error_codes = True
+disable_error_code = misc
```
mdellweg, line 4: I really don't want to do that, but it prevents us from having the two files
`file/remote.py`
`container/remote.py`

Should we rename them to `file_remote.py` and `container_remote.py`? (without the subdirectories)

daviddavis: Adding in `__init__.py` files fixed it for me. Is that not an option?

pulpcore/cli/core/task.py
```diff
@@ -26,21 +27,14 @@ def task(ctx: click.Context) -> None:
...
-    result = task_ctx.list(limit=limit, offset=offset, parameters=params)
-    pulp_ctx.output_result(result)
+# deepcopy to not effect other list subcommands
+list_tasks = deepcopy(list_entities)
+click.option("--name", help="List only tasks with this name.")(list_tasks)
+click.option("--name-contains", "name__contains", help="List only tasks whose name contains this.")(
+    list_tasks
+)
+click.option("--state", help="List only tasks in this state.")(list_tasks)
+task.add_command(list_tasks)
```
mdellweg, line 37: I'm not really sure if this still justifies reusing the common `list_entries` command.

mdellweg:
@daviddavis I decided to move the refactoring to another PR #33

PR #28, opened 2020-11-04
Support PulpExport/Exporter

ggainey:


pulpcore/cli/core/export.py
```diff
@@ -0,0 +1,131 @@
...
+@click.group()
+@click.pass_context
+def export(ctx: click.Context) -> None:
+    pulp_ctx: PulpContext = ctx.find_object(PulpContext)
+    ctx.obj = ExportContext(pulp_ctx)
+
+
+@export.group()
+@click.pass_context
+def pulp(ctx: click.Context) -> None:
```
mdellweg, line 44: Is this level of subcommand needed? As i read it, it will make all the commands look like `pulp export pulp <the real thing>`. Are we ever going to have something else than "pulp" exports?

tests/scripts/test_pulpexporter.sh
```diff
@@ -0,0 +1,65 @@
...
+expect_succ pulp export pulp run --name $NAME2 --full False --versions $REPO1 file 0
+expect_succ pulp export pulp run --name $NAME2 --full False --start-versions $REPO1 file 0 --versions $REPO1 file 1
+expect_succ pulp export pulp run --name $NAME2 --full False --start-versions $REPO1 file 0 --versions $REPO1 file 1 --chunk-size 5KB
+
+# Test list exports
+expect_succ pulp export pulp list --name $NAME1
+
+# Test delete export
+HREF=$(pulp export pulp list --name $NAME1 |  jq -r '.[0].pulp_href')
+expect_succ pulp export pulp delete --href "$HREF"
```
mdellweg, line 60: can we make delete export to accept a name too? (is that a unique thing?)

pulpcore/cli/core/export.py
ggainey: Yes - FilesystemExporter already exists, and i think there has been at least one other proposed that I can't recall right now.

tests/scripts/test_pulpexporter.sh
ggainey: exports do not have a name, alas, just an href

pulpcore/cli/core/export.py
mdellweg: So it will be like
```
pulp export filesystem <...>
pulp export pulp <...>
pulp export rsync <...>
...
```

tests/scripts/test_pulpexporter.sh
mdellweg: how is line 59 supposed to work then? Whose name is it?

pulpcore/cli/core/export.py
```diff
@@ -0,0 +1,131 @@
...
+    PulpContext,
+    PulpEntityContext,
+    RepositoryVersionDefinition,
+)
+
+from typing import Any, Dict, List
+import click
+
+
+class ExportContext(PulpEntityContext):
```
mdellweg, line 12: I believe this class is not used for anything. We can go without it.

pulpcore/cli/core/export.py
```diff
@@ -0,0 +1,131 @@
...
+            raise click.ClickException(f"PulpExporter '{name}' not found.")
+        exporter = search_result["results"][0]
+        return exporter
+
+
+@click.group()
+@click.pass_context
+def export(ctx: click.Context) -> None:
+    pulp_ctx: PulpContext = ctx.find_object(PulpContext)
+    ctx.obj = ExportContext(pulp_ctx)
```
mdellweg, line 39:
```suggestion
@click.pass_context
def export() -> None:
    pass
```

pulpcore/cli/core/export.py
daviddavis: I was thinking it would be like:

```
pulp exporter pulp export list --exporter MyExporter
pulp exporter pulp export create --exporter MyExporter
...
```

daviddavis: I'm fine with `pulp export <type>` though since it's less typing.

tests/scripts/test_pulpexporter.sh
ggainey: It's the name of the exporter the export belongs to - see daviddavis' comment above, I will use --exporter instead of --name to make it more clear what's going on

pulpcore/cli/core/export.py
ggainey: --exporter is a fine idea, +1

pulpcore/cli/core/export.py
ggainey: Yeah, I'm not completely sure what I was thinking here - I think incomplete-understanding of How It All Worked when I started refactoring :)

mdellweg (approved):

PR #31, opened 2020-11-11
Allow to send task to background with ctrl-c

mdellweg:
I was thinking to move the "--background" option from the `sync` to the `main` command.
@daviddavis  wdyt?


daviddavis:
I tested and this works great.

> I was thinking to move the "--background" option from the sync to the main command.

Sure.

mdellweg:
Funny thing is that it "worked" before, but now we handle ctrl-c gracefully.

daviddavis (approved):

daviddavis (approved):
LGTM

PR #33, opened 2020-11-13
Refactor

mdellweg:
I decided to split this work from #27

### DIFF
--- a/pulpcore/cli/common.py
+++ b/pulpcore/cli/common.py
@@ -233,12 +233,31 @@ class PulpEntityContext:
         return repo_version
 
 
+class PulpRepositoryContext(PulpEntityContext):
+    """
+    Base class for repository specific contexts.
+    This class provides the basic CRUD commands as well as synchronizing and
+    ties its instances to the global PulpContext for api access.
+    """
+
+    ENTITY = "repository"
+    SYNC_ID: str
+
+    def sync(self, href: str, body: Dict[str, Any]) -> Any:
+        return self.pulp_ctx.call(
+            self.SYNC_ID,
+            parameters={self.HREF: href},
+            body=body,
+        )
+
+
 ##############################################################################
 # Decorator to access certain contexts or for common options
 
 
 pass_pulp_context = click.make_pass_decorator(PulpContext)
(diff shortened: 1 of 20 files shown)


pulpcore/cli/common.py
```diff
@@ -219,6 +235,106 @@ def find_repository_version(self, definition: RepositoryVersionDefinition) -> An
...
+##############################################################################
+# Generic reusable commands
+
+
+@click.command(name="list")
+@limit_option
+@offset_option
+@pass_entity_context
+@pass_pulp_context
+def list_entities(
```
daviddavis, line 281: Why is this only used in repository and task?

pulpcore/cli/core/task.py
```diff
@@ -26,21 +27,14 @@ def task(ctx: click.Context) -> None:
...
-    result = task_ctx.list(limit=limit, offset=offset, parameters=params)
-    pulp_ctx.output_result(result)
+# deepcopy to not effect other list subcommands
+list_tasks = deepcopy(list_entities)
+click.option("--name", help="List only tasks with this name.")(list_tasks)
+click.option("--name-contains", "name__contains", help="List only tasks whose name contains this.")(
+    list_tasks
+)
+click.option("--state", help="List only tasks in this state.")(list_tasks)
+task.add_command(list_tasks)
```
daviddavis, line 37: Interesting.

daviddavis:
Overall, looks good.

daviddavis (approved):
Approving. Would like to get this merged as I'm about to write some new commands for import.

pulpcore/cli/common.py
mdellweg: because i am lazy...
(Not changing every file before someone tells me: "Yes, this is what we want.".)

pulpcore/cli/core/task.py
mdellweg: That's a nice way to say it. I'm not really happy about this, but it's the best i could come up with while reusing the `list_entities`.

mdellweg:
@daviddavis Feel free to merge. We can still change things later. ;)

PR #34, opened 2020-11-18
Fix package namespacing

mdellweg:
We want to be able to add modules to pulpcore.cli, so that must be a namespace, and hence must not have `__init__.py`.

However this is the minimal change to a solution, and i think we now have the possibility to rethink our namespacing.
Do we want to keep the namespace `pulpcore.cli` or should we move to something like `pulp-cli`?
~In the current state of this PR, every file in `pulpcore/cli` is a module in that namespace. I'd like to get rid of all those files by e.g. moving `pulpcore/cli/core_cli.py` into `pulpcore/cli/core/__init__.py`.~ (needed to be done regardless)
WDYAT?


mdellweg:
Looks like for proper type annotation support we cannot have the single file modules in the namespace at all.

mdellweg:
Discussion with @ggainey : Let's keep "pulpcore.cli" to avoid confusion with any thing pulp2.

daviddavis (approved):

PR #36, opened 2020-11-20
Improve polymorphy around RepositoryVersions

mdellweg:


daviddavis:
Had a busy week and didn't get around to reviewing this. Apologies. Will review this on Monday.

pulpcore/cli/common/context.py
```diff
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Tuple
+from typing import Any, ClassVar, Dict, List, Tuple, Type
```
daviddavis: Nice. Did not know about these.

pulpcore/cli/common/context.py
```diff
@@ -183,8 +183,8 @@ def find(self, **kwargs: Any) -> Any:
     def show(self, href: str) -> Any:
         return self.pulp_ctx.call(self.READ_ID, parameters={self.HREF: href})
 
-    def create(self, body: EntityData) -> Any:
-        return self.pulp_ctx.call(self.CREATE_ID, body=body)
+    def create(self, body: EntityData, parameters: Optional[Dict[str, Any]] = None) -> Any:
```
daviddavis, line 186: :+1:

pulpcore/cli/core/export.py
```diff
@@ -35,7 +41,7 @@ def read(pulp_ctx: PulpContext, export_ctx: PulpExportContext, href: str) -> Non
     pulp_ctx.output_result(entity)
 
 
-@pulp.command()
+@pulp.command(deprecated=True)
```
daviddavis, line 44: Why deprecate these commands instead of just removing them?

daviddavis:
These changes look good to me especially the refactoring of the common code.

The only thing that gives me pause is the repository version change. I feel like it would be better from a user experience point of view to try to keep options at the end of commands when possible so users don't have to worry about where the options should go in addition to what they are.

pulpcore/cli/core/export.py
mdellweg: I wanted to get input from @ggainey at least.
Also the answer could be: "We want it to be called 'delete' everywhere."

mdellweg:
I'll have a look if it is possible.
(It looked too natural to 'know' that repository object at that level of the command structure.)

daviddavis:
Ok, if it's not possible, then I have no strong objection to this change.

mdellweg:
This looks like the wrong direction. :/
https://click.palletsprojects.com/en/7.x/advanced/#forwarding-unknown-options

pulpcore/cli/core/export.py
ggainey: I am a huge fan of consistency. Also, I am/was not tied to the particular structure of the export-commands - I am perfectly fine with adjusting! And finally, I am **also** a big fan of using OO to alleviate boilerplate in every class. Def digging this whole PR.

pulpcore/cli/core/export.py
```diff
@@ -86,20 +80,21 @@ def list(ctx: click.Context, exporter: str, limit: int, offset: int, **kwargs: A
...
     start_versions: List[RepositoryVersionDefinition],
 ) -> None:
-    pulp_ctx: PulpContext = ctx.find_object(PulpContext)
-    pexport_ctx: PulpExportContext = ctx.find_object(PulpExportContext)
-    the_exporter = pexport_ctx.find_exporter(exporter)
+    exporter_ctx = PulpExporterContext(pulp_ctx)
+    exporter_href = exporter_ctx.find(name=exporter)["pulp_href"]
     params = {
-        pexport_ctx.HREF: the_exporter["pulp_href"],
+        export_ctx.HREF: exporter_href,  # TODO This looks WRONG!!!
```
ggainey, line 97: RE "LOOKS WRONG" - exporter-to-export is always a little odd - will try to dig up some time to experiment w/this today.

pulpcore/cli/core/export.py
mdellweg: Yeah, that is mainly why i asked, because the "consistency" with the rest of the pulp ecosystem might suggest we call it `delete` instead of `destroy` for all resources.
I definitely want to have the same name in all parts of the CLI as a primary requirement.

pulpcore/cli/common/context.py
ggainey: I am learning a lot reading x9c4's PRs as well

pulpcore/cli/core/export.py
ggainey: Right now it's largely just the three of us working on/with the CLI, I am ok with skipping deprecation cycles for now.

pulpcore/cli/core/export.py
mdellweg: Oh, almost forgot about this. I'm glad my comment triggered you.

mdellweg:
> Ok, if it's not possible, then I have no strong objection to this change.

I didn't find a good way without hijacking the Parser which is discouraged on the click side.

daviddavis:
That's fine. Then this PR looks good to me.

daviddavis (approved):

PR #40, opened 2020-12-10
Add user command

mdellweg:


pulpcore/cli/common/__init__.py
```diff
@@ -42,7 +42,7 @@ def _config_callback(ctx: click.Context, param: Any, value: str) -> None:
     expose_value=False,
 )
 @click.option("--base-url", default="https://localhost", help="Api base url")
-@click.option("--user", help="Username on pulp server")
+@click.option("--username", help="Username on pulp server")
```
mdellweg, line 45: Turn's out when you try to add a command with the same name as an option the click parser incinerates.

mdellweg:
Users are not meant to be manageable vie the rest api to this time.

daviddavis (approved):

PR #46, opened 2020-12-13
Group permissions

mdellweg:


pulpcore/cli/core/context.py
```diff
@@ -136,6 +136,54 @@ def find(self, **kwargs: Any) -> Any:
...
+        pulp_ctx.needs_plugin("pulpcore", min_version="3.10.dev0")
+        super().__init__(pulp_ctx)
+        self.group_ctx = group_ctx
+
+    def find(self, **kwargs: Any) -> Any:
+        """Workaroud for the missing ability to filter"""
+        # # TODO file a bug and fix upstream
+        # if self.pulp_ctx.has_plugin("pulpcore", min_version="3.10.dev0"):
+        #     # Workaround not needed anymore
+        #     return super().find(**kwargs)
```
daviddavis, line 154: Can this code be removed?

CHANGES/100.feature
```diff
@@ -0,0 +1 @@
+Added label commands.
```
daviddavis, line 1: :thinking:

daviddavis:
The code here looks good to me but I am not familiar with rbac in Pulp. May want to get a second opinion from someone who is or I am happy to approve it without that.

pulpcore/cli/core/context.py
mdellweg: No, actually we need figure, whether we can skip the workaround in, say, 3.11.

CHANGES/100.feature
mdellweg: :exploding_head:

mdellweg: We don't check that a changelog is only added when there is also a `#fixes` stanza, right?

daviddavis: Yep, the [wrong] assumption is that there wouldn't be a changelog if there is no `fixes #...`

daviddavis (approved):
LGTM. One suggestion I might make is to add some help text to the `pulp group` command.

mdellweg:
> LGTM. One suggestion I might make is to add some help text to the `pulp group` command.

Not sure what kind of information to add there, but i added help texts to the list, add, and remove commands for group permissions, as the generic ones didn't seem to fit too well.

daviddavis:
I'd say anything with the word "permission" in it to disambiguate "group". So maybe like "manage group permissions"?

daviddavis (approved):

PR #47, opened 2020-12-14
Move upload logic to artifact context

mdellweg:


pulpcore/cli/core/artifact.py
```diff
@@ -36,38 +35,13 @@ def artifact(ctx: click.Context, pulp_ctx: PulpContext) -> None:
 @artifact.command()
 @click.option("--file", type=click.File("rb"), required=True)
 @click.option(
-    "--chunk-size", default=1000000, type=int, help="Chunk size in bytes (default is 1 MB)"
+    "--chunk-size", default=1048576, type=int, help="Chunk size in bytes (default is 1 MB)"
```
daviddavis, line 38: The default web request size limit for nginx is 1 MiB so I’d suggest we keep the default as 1 MB.

daviddavis:
What’s the reasoning behind moving this code to a context? We’re anticipating other commands will reuse it?

pulpcore/cli/core/artifact.py
daviddavis: We could change the help text to say 1000 bytes.

mdellweg: Because it's smaller? I'll never understand those data size units...
But i'm ok with either.

mdellweg:
Actually, yes. I think about the `file content` command being able to reuse `artifact_ctx.upload()`.

pulpcore/cli/core/artifact.py
mdellweg: https://xkcd.com/394/

daviddavis (approved):

mdellweg:
> Actually, yes. I think about the `file content` command being able to reuse `artifact_ctx.upload()`.

Apart from that i believe that the context is the proper level of abstraction for this kind of "business logic" much like the "wait_on_task".

PR #48, opened 2020-12-17
File content commands

mdellweg:


daviddavis:
This code looks great. A couple thoughts. 

First, it would be nice to have some help text on the create and upload commands so users can easily figure out what the difference is and what they should use.

Secondly, I tried to create a file content unit with upload but it failed with a messy error message that is perhaps hard to parse (see below). I worry that this is not enough information for users. Optimally, it would say something like `Artifact with sha256 ... already exists`. And it would be nice to get a expect_fail test as well. I think this change is probably nontrivial though so feel free to let me know if you want me to file a separate issue for this.

```
$ pulp file content upload --file test.txt --relative-path testing.txt
Uploading file requirements-amazon-all.json
.Upload complete. Creating artifact.
Started background task /pulp/api/v3/tasks/f1532627-9251-493d-9849-d4eec7db452c/
.Error: Task /pulp/api/v3/tasks/f1532627-9251-493d-9849-d4eec7db452c/ failed: '{'non_field_errors': [ErrorDetail(string='sha512 checksum must be unique.', code='unique')]}'
```

daviddavis:
Here's the fix for the second point I raised: https://github.com/pulp/pulpcore/pull/1062

mdellweg:
With respect to the error message, i hope that pulp will start to accept uploads for existing artifacts/content units as a way to show, that a new user should have read permissions on those items. That being said, i'm not opposing to improve on the error message in the meantime.

daviddavis:
I think that makes sense--that if a user uploads an artifact that they don't have access to, they should get a success with the artifact info back. But I am not sure if that same behavior should apply to the case where the user uploads an artifact for which they already have perms for.

mdellweg:
@daviddavis I added the help texts.

daviddavis (approved):

PR #49, opened 2020-12-23
Add a check whether the artifact exists before uploading

daviddavis:
I timed this check on my box (6GB of RAM with HDDs) and it adds about 0.6 seconds for a 100MB file, 3s for a 500MB, and 6s for a 1GB file. 

One thing I debated about whether the command should fail if the artifact already exists?

### DIFF
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -15,20 +15,30 @@ class PulpArtifactContext(PulpEntityContext):
     LIST_ID = "artifacts_list"
     READ_ID = "artifacts_read"
 
-    def upload(self, file: IO[bytes], chunk_size: int = 1000000) -> Any:
+    def upload(self, file: IO[bytes], chunk_size: int = 1000000, check_exists: bool = True) -> Any:
         upload_ctx = PulpUploadContext(self.pulp_ctx)
         start = 0
         size = os.path.getsize(file.name)
         sha256 = hashlib.sha256()
-        upload_href = upload_ctx.create(body={"size": size})["pulp_href"]
+
+        if check_exists:
+            for chunk in iter(lambda: file.read(chunk_size), b""):
+                sha256.update(chunk)
+            result = self.list(limit=1, offset=0, parameters={"sha256": sha256.hexdigest()})
+            if len(result) > 0:
+                click.echo("Artifact already exists.", err=True)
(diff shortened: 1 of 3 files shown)


pulpcore/cli/core/context.py
```diff
@@ -15,20 +15,33 @@ class PulpArtifactContext(PulpEntityContext):
...
-    def upload(self, file: IO[bytes], chunk_size: int = 1000000) -> Any:
+    def upload(self, file: IO[bytes], chunk_size: int = 1000000, check_exists: bool = True) -> Any:
         upload_ctx = PulpUploadContext(self.pulp_ctx)
         start = 0
         size = os.path.getsize(file.name)
         sha256 = hashlib.sha256()
-        upload_href = upload_ctx.create(body={"size": size})["pulp_href"]
+
+        if check_exists:
+            while True:
```
daviddavis, line 25: If only we had Python 3.8 and I could use `while chunk := file.read(chunk_size):` here.

mdellweg: I know it's not exactly the same:

https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/modules/content_upload.py#L176

pulpcore/cli/core/context.py
```diff
@@ -15,20 +15,30 @@ class PulpArtifactContext(PulpEntityContext):
...
+        if check_exists:
+            for chunk in iter(lambda: file.read(chunk_size), b""):
+                sha256.update(chunk)
+            result = self.list(limit=1, offset=0, parameters={"sha256": sha256.hexdigest()})
+            if len(result) > 0:
+                click.echo("Artifact already exists.", err=True)
+                return result[0]["pulp_href"]
+
         click.echo(f"Uploading file {file.name}", err=True)
+        upload_href = upload_ctx.create(body={"size": size})["pulp_href"]
```
mdellweg, line 33: Ack to first talk about what is going to happen before it can fail.

mdellweg (approved):
The code change looks sane.
I am just not sure, whether the cli should give the user the possibility to switch "check-exists".

daviddavis:
> I am just not sure, whether the cli should give the user the possibility to switch "check-exists".

This makes sense to me. I'll remove it and if we need it later, we can re-add it.

PR #50, opened 2020-12-23
Create nightly matrix of pulpcore versions

daviddavis:
[noissue]

### DIFF
--- a/.ci/run_container.sh
+++ b/.ci/run_container.sh
@@ -12,7 +12,12 @@ then
   fi
 fi
 
-"${CONTAINER_RUNTIME}" run --rm --detach --name "pulp" --volume "${BASEPATH}/settings:/etc/pulp" --publish "8080:80" pulp/pulp-fedora31
+if [ -z "${IMAGE_TAG+x}" ]
+then
+  IMAGE_TAG="latest"
+fi
+
+"${CONTAINER_RUNTIME}" run --rm --detach --name "pulp" --volume "${BASEPATH}/settings:/etc/pulp" --publish "8080:80" "pulp/pulp-fedora31:$IMAGE_TAG"
 
 echo "Wait for pulp to start."
 for _ in $(seq 10)
@@ -24,6 +29,9 @@ do
     break
   fi
 done
+
+# show pulpcore/plugin versions we're using
+curl -s http://localhost:8080/pulp/api/v3/status/ | jq ".versions"
 
 # shellcheck disable=SC2064
 trap "${CONTAINER_RUNTIME} stop pulp" EXIT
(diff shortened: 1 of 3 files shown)


daviddavis:
Here's an example run: https://github.com/daviddavis/pulp-cli/runs/1602388622

It looks like the pulpexporter test is failing against pulpcore 3.6.

daviddavis:
This may be a terrible idea as I am not sure how we'll handle the case where new functionality depends on new pulpcore versions.

daviddavis:
Actually, I think we could use $PULPCORE_VERSION in our tests.

mdellweg:
In the cli you can guard with `pulp_ctx.has_plugin("pulpcore", min_version="...")`

.ci/run_container.sh
```diff
@@ -12,7 +13,13 @@ then
   fi
 fi
 
-"${CONTAINER_RUNTIME}" run --rm --detach --name "pulp" --volume "${BASEPATH}/settings:/etc/pulp" --publish "8080:80" pulp/pulp-fedora31
+if [ -n "$PULPCORE_VERSION" ]
+then
+  "${CONTAINER_RUNTIME}" build -f "${BASEPATH}/Containerfile" -t pulp --build-arg PULPCORE_VERSION="==$PULPCORE_VERSION" .
+  IMAGE="pulp"
+fi
```
mdellweg, line 20: It would be really nice to have those images available from pulp-oci-images

daviddavis: Here we go: https://github.com/pulp/pulp-oci-images/pull/37

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,35 @@
...
+  CONTAINER_RUNTIME: 'docker'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image_tag:
+          - "latest"
```
mdellweg, line 20: Do we want 3.9 here also, or are you assuming that "latest" is the same?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,35 @@
...
+          - "3.8"
+          - "3.7"
+          - "3.6"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
mdellweg, line 29: Not sure, it is a good idea, to use a python version not covered by the PR workflows.

mdellweg (approved):
Left two notes, but nothing i could not be convinced of.
This is a great piece of work, and it will help us with the version specific workarounds.

.github/workflows/nightly.yml
daviddavis: It should be the same. In fact, there's no 3.9 container at the moment. I'm open to the idea of having one though.

PR #51, opened 2021-01-04
Use scope property to address nested resources

mdellweg:
This depends on https://github.com/pulp/pulpcore/pull/1066 to have proper href-keys for create of nested entities.
Until then only repository versions (that do not have "create") can benefit.

### DIFF
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -169,13 +169,19 @@ class PulpEntityContext:
         "rpm": "rpm_rpm_repository_href",
     }
 
+    # Subclasses for nested entities can define the parameters for there parent scope here
+    @property
+    def scope(self) -> Dict[str, Any]:
+        return {}
+
     def __init__(self, pulp_ctx: PulpContext) -> None:
         self.pulp_ctx: PulpContext = pulp_ctx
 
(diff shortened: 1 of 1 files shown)


mdellweg:
Let's figure out how to make this version dependent for exports in another PR.

daviddavis (approved):

pulpcore/cli/common/context.py
```diff
@@ -169,13 +169,19 @@ class PulpEntityContext:
         "rpm": "rpm_rpm_repository_href",
     }
 
+    # Subclasses for nested entities can define the parameters for there parent scope here
+    @property
+    def scope(self) -> Dict[str, Any]:
```
daviddavis, line 174: I just noticed this: would a `Dict[str, str]` make sense or is there a possibility that the value could be something other than a string?

mdellweg: That's a good question. I can try making it `Dict[str, str]` and whenever mypy complains we can investigate why.

mdellweg: Actually i just realized that a path parameter could also be an integer (like for repository versions). We don't nest anything beneath such an url yet, but it is a possibility. WDYT?

daviddavis: Good point. `Dict[str, Any]` works then.

PR #52, opened 2021-01-05
Add a workaround for weird export api in < 3.10

mdellweg:
This will be needed if https://github.com/pulp/pulpcore/pull/1066 goes into 3.10.

### DIFF
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -5,6 +5,7 @@ import sys
 import click
 
 from pulpcore.cli.common.context import (
+    PulpContext,
     PulpEntityContext,
 )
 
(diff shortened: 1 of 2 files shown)


pulpcore/cli/core/context.py
```diff
@@ -73,12 +74,18 @@ class PulpExporterContext(PulpEntityContext):
...
     ENTITY = "PulpExport"
-    HREF = "core_pulp_pulp_export_href"
+    HREF = "pulp_pulp_export_href"
     LIST_ID = "exporters_core_pulp_exports_list"
     READ_ID = "exporters_core_pulp_exports_read"
     CREATE_ID = "exporters_core_pulp_exports_create"
     DELETE_ID = "exporters_core_pulp_exports_delete"
 
+    def __init__(self, pulp_ctx: PulpContext) -> None:
+        if not pulp_ctx.has_plugin("pulpcore", min_version="3.10.dev0"):
```
ggainey, line 84: It's a little ugly - but so was the problem you solved in pulpcore #1066.

Can you add a ref to that PR as a comment, so we can save the sanity of Future Generations of readers trying to figure out why we're doing this?

(And by 'Future', I mostly mean "me in three weeks")

mdellweg: Sure.

mdellweg:
Made it a draft until https://github.com/pulp/pulpcore/pull/1066 is in.

ggainey (approved):

PR #55, opened 2021-01-08
Refactor exports workaround

mdellweg:
This refactors to exporter workaround in a way to patch the context
class instead of forcefully changing the base class.

[noissue]

### DIFF
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -1,10 +1,11 @@
-from typing import Any, ClassVar, IO
+from typing import Any, ClassVar, Dict, IO, Optional
 import hashlib
 import os
 import sys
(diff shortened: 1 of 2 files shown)


mdellweg:
@ggainey I hope this makes the Workaround less ugly. And also more consistent, because we do no longer change the base context class once we hit an older server. We probably won't do it at any time, but this enables to talk apropriately with multiple servers in the same process.

mdellweg:
This should allow us to to use #51 at least in the unpatched exports context.

ggainey (approved):
This looks reasonable - I am a big fan of puttnig the work into core, and making the plugin-code simpler. I haven't tested it myself, but assuming you've done so, I'm happy with it.

PR #58, opened 2021-01-11
Add global show command

daviddavis:
fixes #8063

### DIFF
--- a/CHANGES/8063.feature
+++ b/CHANGES/8063.feature
@@ -1,0 +1,1 @@
+Added a global show command to display any href.
--- a/pulpcore/cli/core/__init__.py
+++ b/pulpcore/cli/core/__init__.py
@@ -9,10 +9,12 @@ from pulpcore.cli.core.orphans import orphans
 from pulpcore.cli.core.status import status
 from pulpcore.cli.core.task import task
 from pulpcore.cli.core.user import user
+from pulpcore.cli.core.show import show
 
 
 # Register commands with cli
 main.add_command(status)
+main.add_command(show)
 main.add_command(user)
 main.add_command(group)
 main.add_command(artifact)
--- a/pulpcore/cli/core/show.py
+++ b/pulpcore/cli/core/show.py
@@ -1,0 +1,13 @@
+import click
+
+from pulpcore.cli.common.context import PulpContext, pass_pulp_context
+
+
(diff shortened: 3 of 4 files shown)


daviddavis:
There is an alternative implementation that I debated that is very simple and requires no change to openapi. It's a hack though. Would be something like:

```python
@click.command(name="show")
@click.option("--href", required=True, help="HREF of the entry")
@pass_pulp_context
def show(pulp_ctx: PulpContext, href: str) -> None:
    """Show a resource given its href."""
    entity = pulp_ctx.api.show("artifacts_read", parameters={"artifact_href": href})
    pulp_ctx.output_result(entity)
```

ipanova:
what is the benefit of using cli to display an href `pulp show --href <href>` over pure http GET <href> ?

daviddavis:
@ipanova I think the long term goal of the CLI is to make httpie irrelevant so that most likely users won’t have it installed or configured.

mdellweg:
> what is the benefit of using cli to display an href `pulp show --href <href>` over pure `http GET <href> ` ?

Well, it stays in the tool; it reuses the config set for the cli; it can benefit from output formatting.

mdellweg:
> There is an alternative implementation that I debated that is very simple and requires no change to openapi. It's a hack though. Would be something like:
> 
> ```python
> @click.command(name="show")
> @click.option("--href", required=True, help="HREF of the entry")
> @pass_pulp_context
> def show(pulp_ctx: PulpContext, href: str) -> None:
>     """Show a resource given its href."""
>     entity = pulp_ctx.api.show("artifacts_read", parameters={"artifact_href": href})
>     pulp_ctx.output_result(entity)
> ```

Yeah, i, too, have been thinking of just using the existing artifact operation. My second implementation idea however was to inject a new fake operation into the json-api before parsing it. This might actually be the method that would not stop working, if we ever decide to api conform response parsing.

daviddavis:
> My second implementation idea however was to inject a new fake operation into the json-api before parsing it. This might actually be the method that would not stop working, if we ever decide to api conform response parsing.

Hmm, that's an interesting idea. I'm happy to switch this to artfiact read and then we can later perhaps switch it to this fake operation?

mdellweg:
That sounds like a plan!

mdellweg (approved):

PR #60, opened 2021-01-12
Test PR's on different pulp versions

mdellweg:
This will basically test the same in nightly as in PR's.

[noissue]

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -16,16 +16,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image_tag:
-          - "latest"
-          - "3.8"
-          - "3.7"
+        include:
+          - python: 3.6
+            image_tag: latest
+          - python: 3.7
+            image_tag: 3.8
+          - python: 3.8
+            image_tag: 3.7
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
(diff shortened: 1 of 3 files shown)


mdellweg:
I like sparse test matrices.

.github/workflows/nightly.yml
```diff
@@ -16,16 +16,19 @@ jobs:
...
-          - "latest"
-          - "3.8"
-          - "3.7"
+        include:
+          - python: 3.6
+            image_tag: latest
+          - python: 3.7
+            image_tag: 3.8
+          - python: 3.8
+            image_tag: 3.7
```
daviddavis, line 25:
```suggestion
        image:
          - "latest"
          - "3.8"
          - "3.7"
        python:
          - "3.6"
          - "3.7"
          - "3.8"
```

daviddavis (approved):
I'd suggest maybe testing against all combinations for the nightly job but have no objection if we want to keep nightly sparse.

.github/workflows/nightly.yml
mdellweg: That would triple the number of tests run. I'd rather not do it.

mdellweg:
Can we mark the issue-checker test required?

daviddavis:
> Can we mark the issue-checker test required?

I can configure the required checks after we merge this.

PR #61, opened 2021-01-13
Fix missing packaging requirement

daviddavis:
fixes #8092


bmbouter (approved):
I experienced this problem and this resolved it for me.

mdellweg (approved):

PR #62, opened 2021-01-14
Add pulp label commands

### LINKED ISSUE #100: As a CLI user, I can manage and filter by labels

daviddavis:
fixes #100

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -107,6 +107,14 @@ version_option = click.option(
     expose_value=False,
 )
 
+label_select_option = click.option(
+    "--label-select",
+    "pulp_label_select",
+    help="Filter {entities} by a label search query.",
+    type=str,
+    cls=PulpOption,
+)
+
 ##############################################################################
 # Generic reusable commands
 
@@ -231,6 +239,50 @@ def version_command(**kwargs: Any) -> click.Command:
         pulp_ctx.output_result(result)
 
     return callback
+
+
+def label_command(**kwargs: Any) -> click.Command:
+    """A factory that creates a label command group."""
+
+    if "name" not in kwargs:
+        kwargs["name"] = "label"
+    decorators = kwargs.pop("decorators", [name_option, href_option])
+
+    @click.group(**kwargs)
+    def label_group() -> None:
+        pass
+
+    @click.command(name="set")
+    @click.option("--key", required=True, help="Key of the label")
+    @click.option("--value", required=True, help="Value of the label")
+    @pass_entity_context
+    def label_set(entity_ctx: PulpEntityContext, key: str, value: str) -> None:
+        """Add or update a label"""
+        href = entity_ctx.entity["pulp_href"]
+        entity_ctx.set_label(href, key, value)
+
+    @click.command(name="unset")
+    @click.option("--key", required=True, help="Key of the label")
+    @pass_entity_context
+    def label_unset(entity_ctx: PulpEntityContext, key: str) -> None:
+        """Remove a label with a given key"""
+        href = entity_ctx.entity["pulp_href"]
+        entity_ctx.unset_label(href, key)
+
+    @click.command(name="show")
+    @click.option("--key", required=True, help="Key of the label")
+    @pass_entity_context
+    def label_show(entity_ctx: PulpEntityContext, key: str) -> None:
+        """Show the value for a particular label key"""
+        href = entity_ctx.entity["pulp_href"]
+        click.echo(entity_ctx.show_label(href, key))
+
+    for subcmd in [label_set, label_unset, label_show]:
+        for decorator in decorators:
+            subcmd = decorator(subcmd)
+        label_group.add_command(subcmd)
+
+    return label_group
 
 
 @click.command(name="list")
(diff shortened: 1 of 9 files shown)


daviddavis:
Tests will fail until https://github.com/pulp/pulpcore/pull/1070 is merged. I plan to add this to more commands but was hoping to get a review first.

pulpcore/cli/common/generic.py
```diff
@@ -151,3 +151,42 @@ def version_group(
...
+@click.group(name="label")
+@pass_entity_context
+@pass_pulp_context
+@click.pass_context
+def label_group(ctx: click.Context, entity_ctx: PulpEntityContext, pulp_ctx: PulpContext) -> None:
+    """Manage labels for this resource"""
+    pass
+
+
+@click.command(name="set")
```
mdellweg, line 166:
```suggestion
@label_group.command(name="set")
```
I think in the same file we can spare us the extra `add_command` (and the trouble of forgetting it).

tests/scripts/config.source
```diff
@@ -3,17 +3,17 @@ settings="$(dirname "$(realpath "$0")")/config/pulp/settings.toml"
 PULP_BASE_URL="$(sed -n -e 's/^base_url\s*=\s*"\(\S*\)"\s*$/\1/p' "$settings")"
 VERIFY_SSL="$(sed -n -e 's/^verify_ssl\s*=\s*\(\S*\)\s*$/\1/p' "$settings")"
 
+# Configure path to config
+export XDG_CONFIG_HOME="$(dirname "$(realpath "$0")")/config"
+
 # Constants used in tests
-PULPCORE_VERSION=$(http "$PULP_BASE_URL/pulp/api/v3/status/" | jq -r '.versions | .[] | select(.component=="pulpcore").version')
+PULPCORE_VERSION=$(pulp status | jq -r '.versions | .[] | select(.component=="pulpcore").version')
```
mdellweg, line 10: :+1:

tests/scripts/test_label.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(realpath "$0")")/config.source"
+
+require_min_pulp "3.10.0" || exit 0
```
mdellweg, line 6: This is a nice primitive!
I would however move the `exit 0` into the function and accompany it with a `echo SKIPPING`.
Maybe with a special error code, we can tell the "test_scripts.py" that it was skipped instead of passed.

tests/test_scripts.py
```diff
@@ -13,4 +13,4 @@
 @pytest.mark.parametrize("test_name", TEST_NAMES)
 def test_script(test_name):
     run = subprocess.run([os.path.join("tests", "scripts", "test_" + test_name + ".sh")])
-    assert run.returncode == 0
+    assert run.returncode in [0, 3]
```
mdellweg, line 16:
```suggestion
    if run.returncode == 3:
        pytest.skip("Skipped as requested by the script.")
    assert run.returncode == 0
```

mdellweg (approved):
I thought we'd bubble the skipping up to the pytest level.
But i am approving anyway.

daviddavis:
:100:

tests/test_scripts.py
daviddavis: Won't this fail though if the returncode is 3? Maybe we need this:

```python
    if run.returncode == 3:
        pytest.skip("Skipped as requested by the script.")
    else:
        assert run.returncode == 0
```

daviddavis: Oh nevermind, I forgot `pytest.skip` halts execution.

mdellweg (approved):

pulpcore/cli/common/generic.py
```diff
@@ -360,3 +368,38 @@ def version_group(
...
+@pass_entity_context
+@pass_pulp_context
+@click.pass_context
+def label_group(ctx: click.Context, entity_ctx: PulpEntityContext, pulp_ctx: PulpContext) -> None:
+    """Manage labels for this resource"""
+    pass
+
+
+@label_group.command(name="set")
+@click.option("--name", required=True, help="Name of the entry")
```
mdellweg, line 384: I think, the new factory pattern applies quite nicely to those commands. You can move the lookup question into the hands of the code using the factory.
I managed to make some file commands able to understand both name and href that way.

pulpcore/cli/common/generic.py
```diff
@@ -233,6 +241,65 @@ def repair(
...
+
+    if "name" not in kwargs:
+        kwargs["name"] = "label"
+    decorators = kwargs.pop("decorators", [])
+    lookup_decorators = kwargs.pop("lookup_decorators", [name_option, href_option])
+
+    @click.group(**kwargs)
+    @pass_entity_context
+    @pass_pulp_context
+    @click.pass_context
```
mdellweg, line 255: No need to pass if we don't use any of them here.

pulpcore/cli/common/generic.py
```diff
@@ -233,6 +241,65 @@ def repair(
...
+    def label_group(
+        ctx: click.Context, entity_ctx: PulpEntityContext, pulp_ctx: PulpContext
+    ) -> None:
+        pass
+
+    for option in decorators:
+        # Decorate callback
+        label_group = option(label_group)
```
mdellweg, line 263: The only decorators i'd expect to see here are the entity lookup ones if we go for interspersed args. And yet they can be attached one level higher.
I'd prefer to pass the lookup options explicitly as the decorator argument.

mdellweg (approved):

PR #63, opened 2021-01-14
Adding pulp_ansible support

gerrod3:
This adds basic CRUD support for ansible repositories, role remotes, collection remotes, distributions, and syncing.

fixes: #8079
fixes: #8080
fixes: #8081
fixes: #8082
https://pulp.plan.io/issues/8079
https://pulp.plan.io/issues/8080
https://pulp.plan.io/issues/8081
https://pulp.plan.io/issues/8082

### DIFF
--- a/CHANGES/8079.feature
+++ b/CHANGES/8079.feature
@@ -1,0 +1,1 @@
+Basic CRUD support for Ansible repositories
--- a/CHANGES/8081.feature
+++ b/CHANGES/8081.feature
@@ -1,0 +1,1 @@
+Basic CRUD for Ansible collection remotes, use '-t collection' after remote to select type
--- a/pulpcore/cli/ansible/context.py
+++ b/pulpcore/cli/ansible/context.py
@@ -1,0 +1,59 @@
+from pulpcore.cli.common.context import (
+    PulpEntityContext,
+    PulpRepositoryContext,
+    PulpRepositoryVersionContext,
+)
+
+
+# TODO Add Role and Collection Content contexts
+
+
+class PulpAnsibleDistributionContext(PulpEntityContext):
+    ENTITY = "distribution"
+    HREF = "ansible_ansible_distribution_href"
+    LIST_ID = "distributions_ansible_ansible_list"
+    READ_ID = "distributions_ansible_ansible_read"
+    CREATE_ID = "distributions_ansible_ansible_create"
+    UPDATE_ID = "distributions_ansible_ansible_partial_update"
+    DELETE_ID = "distributions_ansible_ansible_delete"
+
+
+class PulpAnsibleRoleRemoteContext(PulpEntityContext):
+    ENTITY = "role remote"
+    HREF = "ansible_role_remote_href"
+    LIST_ID = "remotes_ansible_role_list"
+    READ_ID = "remotes_ansible_role_read"
+    CREATE_ID = "remotes_ansible_role_create"
+    UPDATE_ID = "remotes_ansible_role_partial_update"
+    DELETE_ID = "remotes_ansible_role_delete"
+
+
+class PulpAnsibleCollectionRemoteContext(PulpEntityContext):
+    ENTITY = "collection remote"
+    HREF = "ansible_collection_remote_href"
+    LIST_ID = "remotes_ansible_collection_list"
+    READ_ID = "remotes_ansible_collection_read"
+    CREATE_ID = "remotes_ansible_collection_create"
+    UPDATE_ID = "remotes_ansible_collection_partial_update"
+    DELETE_ID = "remotes_ansible_collection_delete"
+
+
+class PulpAnsibleRepositoryVersionContext(PulpRepositoryVersionContext):
+    HREF = "ansible_ansible_repository_version_href"
+    REPOSITORY_HREF = "ansible_ansible_repository_href"
+    LIST_ID = "repositories_ansible_ansible_versions_list"
+    READ_ID = "repositories_ansible_ansible_versions_read"
+    DELETE_ID = "repositories_ansible_ansible_versions_delete"
+    REPAIR_ID = "repositories_ansible_ansible_versions_repair"
+
+
+class PulpAnsibleRepositoryContext(PulpRepositoryContext):
+    HREF = "ansible_ansible_repository_href"
+    LIST_ID = "repositories_ansible_ansible_list"
+    READ_ID = "repositories_ansible_ansible_read"
+    CREATE_ID = "repositories_ansible_ansible_create"
+    UPDATE_ID = "repositories_ansible_ansible_partial_update"
+    DELETE_ID = "repositories_ansible_ansible_delete"
+    SYNC_ID = "repositories_ansible_ansible_sync"
+    MODIFY_ID = "repositories_ansible_ansible_modify"
+    VERSION_CONTEXT = PulpAnsibleRepositoryVersionContext
--- a/pulpcore/cli/ansible/distribution.py
+++ b/pulpcore/cli/ansible/distribution.py
@@ -1,0 +1,136 @@
+from typing import Optional
+
+import click
+
+
+from pulpcore.cli.common.generic import (
+    list_entities,
+    show_by_name,
+    destroy_by_name,
+)
+from pulpcore.cli.common.context import (
+    PulpContext,
+    pass_pulp_context,
+    pass_entity_context,
+    EntityDefinition,
+)
+from pulpcore.cli.ansible.context import (
+    PulpAnsibleDistributionContext,
+    PulpAnsibleRepositoryContext,
+)
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "distribution_type",
+    type=click.Choice(["ansible"], case_sensitive=False),
+    default="ansible",
+)
+@pass_pulp_context
+@click.pass_context
+def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: str) -> None:
+    if distribution_type == "ansible":
+        ctx.obj = PulpAnsibleDistributionContext(pulp_ctx)
+    else:
+        raise NotImplementedError()
+
+
+distribution.add_command(list_entities)
+distribution.add_command(show_by_name)
+distribution.add_command(destroy_by_name)
+
+
+# TODO Add content_guard option
+@distribution.command()
+@click.option("--name", required=True)
+@click.option("--base-path", required=True)
+@click.option("--repository", help="name of repository")
+@click.option("--version", type=int, help="version of repository, leave blank for always latest")
+@pass_entity_context
+@pass_pulp_context
+def create(
+    pulp_ctx: PulpContext,
+    distribution_ctx: PulpAnsibleDistributionContext,
+    name: str,
+    base_path: str,
+    repository: Optional[str],
+    version: Optional[int],
+) -> None:
+    """Creates a distribution at base-path for repository's content to be discovered at"""
+    body: EntityDefinition = {"name": name, "base_path": base_path}
+    repo: EntityDefinition = PulpAnsibleRepositoryContext(pulp_ctx).find(name=repository)
+    if version is not None and not repository:
+        click.ClickException("You must set --repository when using version")
+    elif version is not None:
+        body["repository_version"] = f'{repo["versions_href"]}{version}/'
+    elif repository:
+        body["repository"] = repo["pulp_href"]
+    result = distribution_ctx.create(body=body)
+    distribution = distribution_ctx.show(result["created_resources"][0])
+    pulp_ctx.output_result(distribution)
+
+
+@distribution.command()
+@click.option("--name", required=True)
+@click.option("--base-path", help="new base_path")
+@click.option("--repository", type=str, default=None, help="new repository to be served")
+@click.option(
+    "--version",
+    type=int,
+    default=None,
+    help="version of new repository to be served, leave blank for always latest",
+)
+@pass_entity_context
+@pass_pulp_context
+def update(
+    pulp_ctx: PulpContext,
+    distribution_ctx: PulpAnsibleDistributionContext,
+    name: str,
+    base_path: Optional[str],
+    repository: Optional[str],
+    version: Optional[int],
+) -> None:
+    """
+    To remove repository or repository_version fields set --repository to ""
+    """
+    dist_body: EntityDefinition = distribution_ctx.find(name=name)
+    href: str = dist_body["pulp_href"]
+    body: EntityDefinition = dict()
+
+    if base_path:
+        body["base_path"] = base_path
+    if repository is not None:
+        if repository == "":
+            # unset repository or repository version
+            if dist_body["repository"]:
+                body["repository"] = ""
+            elif dist_body["repository_version"]:
+                body["repository_version"] = ""
+        else:
+            repo = PulpAnsibleRepositoryContext(pulp_ctx).find(name=repository)
+            if version is not None:
+                if dist_body["repository"]:
+                    distribution_ctx.update(href, body={"repository": ""})
+                body["repository_version"] = f'{repo["versions_href"]}{version}/'
+            else:
+                if dist_body["repository_version"]:
+                    distribution_ctx.update(href, body={"repository_version": ""})
+                body["repository"] = repo["pulp_href"]
+    elif version is not None:
+        # keep current repository, change version
+        if dist_body["repository"]:
+            distribution_ctx.update(href, body={"repository": ""})
+            body["repository_version"] = f'{dist_body["repository"]}versions/{version}/'
+        elif dist_body["repository_version"]:
+            repository_href, _, _ = dist_body["repository_version"].partition("versions")
+            body["repository_version"] = f"{repository_href}versions/{version}/"
+        else:
+            raise click.ClickException(
+                f"Distribution {name} doesn't have a repository set, "
+                f"please specify the repository to use  with --repository"
+            )
+    distribution_ctx.update(href, body=body)
+    dist_body = distribution_ctx.show(href)
+    pulp_ctx.output_result(dist_body)
--- a/pulpcore/cli/ansible/remote.py
+++ b/pulpcore/cli/ansible/remote.py
@@ -1,0 +1,192 @@
+import click
+
+import yaml
+
+from typing import Any, Dict, Optional
+
+from pulpcore.cli.common.generic import (
+    list_entities,
+    show_by_name,
+    destroy_by_name,
+)
+from pulpcore.cli.common.context import (
+    PulpContext,
+    pass_pulp_context,
+    pass_entity_context,
+    PulpEntityContext,
+    EntityDefinition,
+)
+from pulpcore.cli.ansible.context import (
+    PulpAnsibleRoleRemoteContext,
+    PulpAnsibleCollectionRemoteContext,
+)
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "remote_type",
+    type=click.Choice(["collection", "role"], case_sensitive=False),
+    default="collection",
+    is_eager=True,
+)
+@pass_pulp_context
+@click.pass_context
+def remote(ctx: click.Context, pulp_ctx: PulpContext, remote_type: str) -> None:
+    if remote_type == "role":
+        ctx.obj = PulpAnsibleRoleRemoteContext(pulp_ctx)
+    elif remote_type == "collection":
+        ctx.obj = PulpAnsibleCollectionRemoteContext(pulp_ctx)
+    else:
+        raise NotImplementedError()
+
+
+remote_options = {
+    "ca-cert": (str, "a PEM encoded CA certificate"),
+    "client-cert": (str, "a PEM encoded client certificate"),
+    "client-key": (str, "a PEM encode private key"),
+    "tls-validation": (str, "if True, TLS peer validation must be performed"),
+    "proxy-url": (str, "format: scheme//user:password@host:port"),
+    "username": (str, "used for authentication when syncing"),
+    "password": (str, ""),
+    "download-concurrency": (int, "total number of simultaneous connections"),
+    "policy": (str, "policy to use when downloading"),
+    "total-timeout": (str, "aiohttp.ClientTimeout.total for download connections"),
+    "connect-timeout": (str, "aiohttp.ClientTimeout.connect for download connections"),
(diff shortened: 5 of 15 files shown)


pulpcore/cli/ansible/context.py
```diff
@@ -0,0 +1,59 @@
...
+    ENTITY = "distribution"
+    HREF = "ansible_ansible_distribution_href"
+    LIST_ID = "distributions_ansible_ansible_list"
+    READ_ID = "distributions_ansible_ansible_read"
+    CREATE_ID = "distributions_ansible_ansible_create"
+    UPDATE_ID = "distributions_ansible_ansible_update"
+    DELETE_ID = "distributions_ansible_ansible_delete"
+
+
+class PulpAnsibleRemoteRoleContext(PulpEntityContext):
```
mdellweg, line 21:
```suggestion
class PulpAnsibleRoleRemoteContext(PulpEntityContext):
```

pulpcore/cli/ansible/context.py
```diff
@@ -0,0 +1,59 @@
...
+    ENTITY = "role remote"
+    HREF = "ansible_role_remote_href"
+    LIST_ID = "remotes_ansible_role_list"
+    READ_ID = "remotes_ansible_role_read"
+    CREATE_ID = "remotes_ansible_role_create"
+    UPDATE_ID = "remotes_ansible_role_partial_update"  # This makes update have less requirements
+    DELETE_ID = "remotes_ansible_role_delete"
+
+
+class PulpAnsibleRemoteCollectionContext(PulpEntityContext):
```
mdellweg, line 31: same.

pulpcore/cli/ansible/distribution.py
```diff
@@ -0,0 +1,133 @@
...
+            if version:
+                if dist_body["repository"]:
+                    unset_repository(pulp_ctx, distribution_ctx, href, "repository")
+                body["repository_version"] = f'{repo["versions_href"]}{version}/'
+            else:
+                if dist_body["repository_version"]:
+                    unset_repository(pulp_ctx, distribution_ctx, href, "repository_version")
+                body["repository"] = repo["pulp_href"]
+    if content_guard is not None:
+        body["content_guard"] = content_guard
```
mdellweg, line 118: This is not referring to the content guard by it's name.
You'd need to add a corresponding context class and do the lookup (with c_g_ctx.find()) here.

pulpcore/cli/ansible/distribution.py
```diff
@@ -0,0 +1,133 @@
...
+
+def unset_repository(
+    pulp_ctx: PulpContext, distribution_ctx: PulpAnsibleDistributionContext, href: str, repo: str
+) -> None:
+    # unset repository or repository_version so call will succeed
+    pulp_ctx.call(
+        "distributions_ansible_ansible_partial_update",
+        parameters={distribution_ctx.HREF: href},
+        body={repo: ""},
+        uploads=None,
```
mdellweg, line 132:
```suggestion
```
I hope, that's a default.

pulpcore/cli/ansible/distribution.py
```diff
@@ -0,0 +1,133 @@
...
+        body["repository"] = repo["pulp_href"]
+    if content_guard:
+        body["content_guard"] = content_guard
+    result = distribution_ctx.create(body=body)
+    distribution = distribution_ctx.show(result["created_resources"][0])
+    pulp_ctx.output_result(distribution)
+
+
+@distribution.command()
+@click.option("--href", required=True)
```
mdellweg, line 75: We always try to handle objects by their name, not their href.
Try to focus on the admin, not the api.

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,100 @@
...
+@pass_pulp_context
+def update(pulp_ctx: PulpContext, remote_ctx: PulpEntityContext, href: str, args: Any) -> None:
+    """ "
+    Performs a partial update of the remote
+    The one annoying part of this is that you want update a collection remote you need to
+    specify with -t parameter after 'remote' or else defaults to role remote
+    """
+    body: Dict[str, Any] = dict()
+    if args:
+        body.update(extra_options(args))
```
mdellweg, line 88: interesting Idea! I need to think more about this, as you already realized, it lacks the click validation and help automation. Maybe we can get it with a bit of click.Context hackery.

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,100 @@
...
+    pulp_ctx.output_result(body)
+
+
+def extra_options(args: Any) -> Dict[str, Any]:
+    # This won't work if they do something like '--option=param'
+    # Needs to be in format '--option param'
+    if len(args) % 2 != 0:
+        raise click.ClickException("An option is missing a parameter")
+    it: Iterator[str] = iter(args)
+    return {option.strip("-"): param for (option, param) in zip(it, it)}
```
mdellweg, line 100: I really hope, we find a way to not do this manually.

pulpcore/cli/ansible/repository.py
```diff
@@ -0,0 +1,146 @@
...
+        except click.ClickException:
+            r_href = PulpAnsibleRemoteCollectionContext(pulp_ctx).find(name=remote)["pulp_href"]
+        repository["remote"] = r_href
+
+    result = repository_ctx.create(body=repository)
+    pulp_ctx.output_result(result)
+
+
+@repository.command()
+@click.option("--href", required=True)
```
mdellweg, line 71: repos have names. Ideally, the admin would never need to worry about an href.

pulpcore/cli/ansible/repository.py
```diff
@@ -0,0 +1,146 @@
...
+        if description == "":
+            # unset the description
+            description = None
+        if description != repository["description"]:
+            repository["description"] = description
+
+    if remote is not None:
+        if remote == "":
+            # unset the remote
+            r_href: str = ""
```
mdellweg, line 101: We are not limited to 640 kb anymore. ;)
```suggestion
            remote_href: str = ""
```

mdellweg (commented):
Great start. I left a bunch of comments. Mostly regarding the philosophy of this CLI project.

pulpcore/cli/ansible/remote.py
mdellweg: I think, I am onto something here:
https://github.com/pulp/pulp-cli/pull/64
It's even less hacky than i thougt.
Can you see how it maps to your problem?

pulpcore/cli/ansible/distribution.py
gerrod3: Hmmm, so for updates should the cli have an option for --new-name. I originally had it using name for the lookup, but then realized that you can change the name of most objects when performing an update as long as the new one doesn't clash.

mdellweg: Noooo, why should you ever rename something. It leads to problems. Yes, i know the api allows it, but we don't do it now.
Again: Don't focus on the api, but the admin usecase.
BTW: The change i am about to make will probably allow you to change the name. But still why...

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,232 @@
...
+)
+from pulpcore.cli.ansible.context import (
+    PulpAnsibleRoleRemoteContext,
+    PulpAnsibleCollectionRemoteContext,
+)
+
+RT = TypeVar("RT")  # return type
+
+
+def _type_callback(ctx: click.Context, param: Any, value: str) -> str:
```
gerrod3, line 24: @mdellweg This works pretty well for the two remote types. The one thing is that the --help option will show the options from both commands despite what type is selected. Not sure if it's a bug or intended behaviour.

CHANGES/8081.feature
```diff
@@ -0,0 +1 @@
+Basic CRUD for Ansible collection remotes, use '-t collection' after remote to select type
```
bmbouter, line 1: I tried `pulp ansible remote create -t collection --name myremote --url https://galaxy.ansible.com` and it says `Error: no such option: -t`

bmbouter: This is on a49b6ba8a8b864e07cfa1b62db1f39cc508c3997

gerrod3: I guess I worded it poorly, it's `pulp ansible remote -t collection create --name myremote --url https://galaxy.ansible.com`
Also, I'm not sure how --requirements-file is suppose to be inputed into collection remote create. I tried using the example on the workflows, but it didnt work.

mdellweg: You can use `pulp -v <...>` in order to see the http requests made by the cli. This might help you debugging.

pulpcore/cli/ansible/remote.py
mdellweg: Please don't invest too much into this right now. Just keep it working for now. I am experimenting with a more general approach here. And it seems to do just fine according to the help screen.

https://github.com/pulp/pulp-cli/pull/64/files#diff-2330f1d2bec4c88428af0e90f4ad6feff6a1663af7be7b9fedb645ed98846f68R18-R42

CHANGES/8079.feature
```diff
@@ -0,0 +1 @@
+Basic CRUD support for Ansible repositories
```
mdellweg, line 1: I'd prefer, if your editor put a single linebreak at the end of files.

pulpcore/cli/ansible/context.py
```diff
@@ -0,0 +1,59 @@
...
+    DELETE_ID = "distributions_ansible_ansible_delete"
+
+
+class PulpAnsibleRoleRemoteContext(PulpEntityContext):
+    ENTITY = "role remote"
+    HREF = "ansible_role_remote_href"
+    LIST_ID = "remotes_ansible_role_list"
+    READ_ID = "remotes_ansible_role_read"
+    CREATE_ID = "remotes_ansible_role_create"
+    UPDATE_ID = "remotes_ansible_role_partial_update"  # This makes update have less requirements
```
mdellweg, line 27: Please let's keep this consistent.
Maybe we should do PATCH instead of PUT in all places. It exhibits less danger for race conditions.
@ggainey @daviddavis @dkliban ?

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,232 @@
...
+    return value
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "remote_type",
+    type=click.Choice(["role", "collection"], case_sensitive=False),
+    default="role",
```
mdellweg, line 41: is "role" a good choice for the default?
I'd expect the whole ansible infrastructure to move towards collections (they can contain roles and are therefore much more generic).
cc @bmbouter @fao89

pulpcore/cli/ansible/context.py
daviddavis: I agree that using PATCH is probably better than PUT but most places like here use async tasks to perform updates so it's not a high priority IMO.

pulpcore/cli/ansible/remote.py
daviddavis: Yea, collection is better. I think we want to encourage users to start using collections.

fao89: +1 to collection

bmbouter: +1 to collection

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,232 @@
...
+    body: Dict[str, str] = {"name": name, "url": url}
+    create_remote(
+        pulp_ctx=pulp_ctx, remote_ctx=remote_ctx, body=body, remote_options=remote_options
+    )
+
+
+@remote.command(name="create")
+@click.option("--name", required=True)
+@click.option("--url", required=True)
+@click.option("--requirements-file", help="Collections only: A Collection requirements yaml")
```
daviddavis, line 110: I would probably have this be a file option. I think in most cases, users will actually have an actual requirements.yaml file they'll want to use. Perhaps have `--requirements` and `--requirements-file` options which take the requirements file contents and an actual requirements.yaml file respectively?

bmbouter: I'm ok with just taking a file path at the start. Having both is also ok w/ me.

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,244 @@
...
+@click.option("--name", required=True)
+@click.option("--url")
+@click.option("--requirements-file", type=click.File())
+@click.option("--requirements")
+@click.option("--auth-url")
+@click.option("--token")
+@remote_options
+@pass_entity_context
+@pass_pulp_context
+def update_collection(
```
mdellweg, line 197: After thinking more about this, you should really collapse both update commands into one, and maintain a list of allowed parameters (probably in the corresponding contexts, because that is where the polymorphy lives) to error out if wrong ones are provided.

pulpcore/cli/ansible/repository.py
```diff
@@ -0,0 +1,146 @@
...
+    name: str,
+    description: Optional[str],
+    remote: Optional[str],
+) -> None:
+    repository = {"name": name, "description": description}
+    if remote is not None:
+        try:
+            remote_href: str = PulpAnsibleRoleRemoteContext(pulp_ctx).find(name=remote)["pulp_href"]
+        except click.ClickException:
+            remote_href = PulpAnsibleCollectionRemoteContext(pulp_ctx).find(name=remote)[
```
mdellweg, line 63: The order here should reflect the default for the collection type.
Or should we also ask for the `--collection-type` here?

setup.py
```diff
@@ -28,6 +28,7 @@
             "file=pulpcore.cli.file",
             "container=pulpcore.cli.container",
             "rpm=pulpcore.cli.rpm",
+            "ansible=pulpcore.cli.ansible",
```
mdellweg, line 31: I guess it's time to get this into lexical order. XD

mdellweg (commented):
More comments...

If you like, we can split this PR and get the less controversial parts merged earier. OTOH, i believe we are almost there.

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,244 @@
...
+    auth_url: Optional[str],
+    token: Optional[str],
+    **remote_options: Dict[str, Any],
+) -> None:
+    body: Dict[str, Any] = {"name": name, "url": url}
+
+    if requirements_file is not None:
+        body["requirements_file"] = f"{yaml.safe_load(requirements_file)}"
+    elif requirements is not None:
+        body["requirements_file"] = yaml.safe_load(f'"{requirements}"')
```
gerrod3, line 136: This is the solution I found after not getting the JSON modified encoding to work.  It's still kind of hacky and I'm trying to figure out why I have to have two different ways of calling safe_load. Ideally it would look like
`
requirements_file = requirements_file or requirements
if requirements_file is not None:
    body["requirements_file"] = f"{yaml.safe_load(requirements_file)}"
`
However, for some reason it doesn't like accepting the string arguement unless I wrap it in more quotes. Might have to do with how click processes strings.

mdellweg (changes requested):
We are getting there, but still some things to be done...

pulpcore/cli/ansible/context.py
```diff
@@ -0,0 +1,59 @@
...
+# TODO Add Role and Collection Content contexts
+
+
+class PulpAnsibleDistributionContext(PulpEntityContext):
+    ENTITY = "distribution"
+    HREF = "ansible_ansible_distribution_href"
+    LIST_ID = "distributions_ansible_ansible_list"
+    READ_ID = "distributions_ansible_ansible_read"
+    CREATE_ID = "distributions_ansible_ansible_create"
+    UPDATE_ID = "distributions_ansible_ansible_partial_update"
```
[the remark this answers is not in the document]
gerrod3: No it works both ways. I changed it due to an earlier comment saying PATCH was better. PATCH eases the requirements for the command allowing for the update syntax to be cleaner. With PUT, required fields would need to be re-entered even if you weren't updating them, or you would have to specified them to be the same value in the code in order to work like a PATCH command. I'll make a PR to change it everywhere since I think it's better.

pulpcore/cli/ansible/distribution.py
```diff
@@ -0,0 +1,113 @@
...
+def update(
+    pulp_ctx: PulpContext,
+    distribution_ctx: PulpAnsibleDistributionContext,
+    name: str,
+    base_path: Optional[str],
+    repository: Optional[str],
+    version: Optional[str],
+) -> None:
+    """
+    To remove repository or repository_version set --repository to ""
```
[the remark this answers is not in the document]
gerrod3: Yes, I added it to show some behaviour of the command that I didn't think was entirely intuitive. I'll add more descriptions to the commands.

pulpcore/cli/ansible/distribution.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    dist_body = distribution_ctx.find(name=name)
+    href = dist_body["pulp_href"]
+    body = dict()
+    if base_path:
+        body["base_path"] = base_path
+    if repository is not None:
```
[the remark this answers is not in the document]
gerrod3: I think this command would benefit from a comment since I think you are misunderstanding how it works. The two options '--repository' (repository's name) and '--version' (a repository's version number) are not mutually exclusive. If you only specify repository then it'll set the repository field. If you specify repository and version, it'll set the repository_version field.  This is also how it works in the create command.

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,199 @@
...
+    PulpAnsibleCollectionRemoteContext,
+)
+
+
+def _type_callback(ctx: click.Context, param: Any, value: str) -> str:
+    group: click.Group = cast(click.Group, ctx.command)
+    if value == "role":
+        group.commands.pop("create_collection", None)
+    else:
+        group.commands.pop("create_role", None)
```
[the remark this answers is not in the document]
gerrod3: Hmmm, Maybe I had a different interpretation of your polymorphic PR.

pulpcore/cli/ansible/distribution.py
```diff
@@ -0,0 +1,113 @@
...
+    base_path: str,
+    repository: Optional[str],
+    version: Optional[str],
+) -> None:
+    body: Dict[str, Any] = {"name": name, "base_path": base_path}
+    repo: Dict[str, Any] = PulpAnsibleRepositoryContext(pulp_ctx).find(name=repository)
+    if version:
+        body["repository_version"] = f'{repo["versions_href"]}{version}/'
+    elif repository:
+        body["repository"] = repo["pulp_href"]
```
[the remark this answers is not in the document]
gerrod3: Do you want me to add this logic to other plugin commands in a separate pr?

bmbouter:
I'm staying out of the code review, but I just tested all the commands and as a user they are all meeting my needs wonderfully.

pulpcore/cli/ansible/context.py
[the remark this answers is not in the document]
mdellweg: > PATCH eases the requirements for the command allowing for the update syntax to be cleaner.

Not sure i understand that sentence. But i agree that PATCH is probably safer and easier. I just want to have it used consistently in the whole project. And that change should not be done in this PR under the hood.

pulpcore/cli/ansible/distribution.py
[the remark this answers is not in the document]
mdellweg: That would be awesome. And yes, please make it a separate PR. It is easier to review, because a smaller PR is less code to look at and usually needs less iterations.

My estimation is: total_review_time is in O(number_of_lines²)

pulpcore/cli/ansible/distribution.py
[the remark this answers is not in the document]
mdellweg: Oh right, version is just the number.
Maybe we should error on a version number without repository then.

pulpcore/cli/ansible/remote.py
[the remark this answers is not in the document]
mdellweg: Oh the polymorphic approach was declined. ...for good reasons.

mdellweg:
> I'm staying out of the code review, but I just tested all the commands and as a user they are all meeting my needs wonderfully.

That is good to hear! I you'd confirm that the tests cover those needs, that would be even better.

pulpcore/cli/ansible/remote.py
```diff
@@ -0,0 +1,197 @@
...
+    remote_ctx: PulpEntityContext,
+    requirements_file: Any,
+    requirements: Optional[str],
+    auth_url: Optional[str],
+    token: Optional[str],
+) -> EntityDefinition:
+
+    body: EntityDefinition = dict()
+    if isinstance(remote_ctx, PulpAnsibleRoleRemoteContext) and any(
+        [requirements_file, requirements, auth_url, token]
```
mdellweg, line 81: can you add a test where you hit one of those and check for the `"${ERROUTPUT}"`?

mdellweg (commented):
Another round... (Sorry, it's hard to see all aspects in one go on such a huge PR)

Please have a look at the help texts. (I know we do not have good help everywhere, but we should not add confusing or unhelpful messages either.)

pulpcore/cli/ansible/repository.py
```diff
@@ -0,0 +1,158 @@
...
+@pass_pulp_context
+def update(
+    pulp_ctx: PulpContext,
+    repository_ctx: PulpRepositoryContext,
+    name: str,
+    description: Optional[str],
+    remote: Optional[str],
+) -> None:
+    """
+    Trigger an asynchronous partial update task
```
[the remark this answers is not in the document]
gerrod3: I kind of disagree. It does show the task href that was generated to perform the task so I think it's helpful for understanding what the command is doing. Also, this is straight from the openapi generated docs.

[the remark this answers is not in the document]
mdellweg: Well that task information might be hidden behind some level of "-v" or "-q" eventually. The CLI is really meant to abstract the rest api wierdness away from the user. And waiting for tasks is part of that. So the users perspective is that he orders the command to update the repository and the command will return to the user once the deed is done. That should be the information carried by the help text. Hitting ctrl-c to interrupt waiting is considered advanced behavior. The normal should never care about tasks.
And that is the reason, why we don't take anything straight from the API without questioning it.

mdellweg (approved):

daviddavis (approved):

PR #64, opened 2021-01-15
WIP: Move entity lookup one level up in command chain

mdellweg:
This is again the start of a bigger refactoring effort.
But i believe it gives the entity lookup a much more logical place in the command chain.
Also it helps to list the proper available commands when you specify a name.

### DIFF
--- a/pulpcore/cli/core/task.py
+++ b/pulpcore/cli/core/task.py
@@ -1,8 +1,7 @@
-from copy import deepcopy
 import click
 
 from pulpcore.cli.common.generic import (
-    list_entities,
+    list_command,
 )
 from pulpcore.cli.common.context import (
     pass_pulp_context,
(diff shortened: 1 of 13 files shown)


mdellweg:
This change is starting to get even bigger than i anticipated.

pulpcore/cli/core/task.py
```diff
@@ -21,14 +20,19 @@ def task(ctx: click.Context, pulp_ctx: PulpContext) -> None:
     ctx.obj = PulpTaskContext(pulp_ctx)
 
 
-# deepcopy to not effect other list subcommands
```
mdellweg, line 24: I never liked the deepcopy here anyway.

daviddavis:
This is an interesting change but to be honest, I dislike it. For one, it greatly reduces the discoverability of commands/options. For example, you can't call `pulp file repository sync --help` or `pulp file repository update --help` to see what the command does and what the possible options are. In addition, interspersing options means that users have to remember where options go in addition to what those options are.

daviddavis:
I reviewed the pulp_ansible PR and I see some of the motivations for this change. I am a little less apprehensive about this design for `--type` than I am for other parameters like `--name`.

mdellweg:
With this change, we get a clear distinction between parameters to lookup the entity and parameters that should be updated on the entity. This will even allow for a simple renaming schema. Also i see that we can provide optional "name" and "href" at the command group context for lookup.
With a little bit more hackery on the Group subclass, i think we can get all the subcommands back to show up in the help screen.

daviddavis:
I definitely get how this change makes the code easier and how potentially users could view these lookup options differently. But I worry that it's less intuitive for users than keeping options in one place. This conversation I think demonstrates my concern: https://github.com/pulp/pulp-cli/pull/63#discussion_r558596566

mdellweg:
So now, I came to the conclusion, that you are probably right about the conditional commands. We want the help to be executable at all times. Let's consider this branch an experiment, and see what I can recover from it.
For example, I kind of like the lazy entity lookup. If we manage to do the same with the api-docs (maybe impossible) then we can even get the full help befor calling out to the server for the first time.
With the interspersed arguments, we disagree, as I think they give more structure to the buildup of the commands. Also the help is quite reasonable at describing where your params are expected. Let's discuss further.

mdellweg:
Everything interesting has been extracted into other PR's now. Closing.

PR #67, opened 2021-01-15
0.1.1.dev bump

daviddavis:
[noissue]


mdellweg (approved):
Let's merge this immediately.
(Should have been part of the release process to close the window for confusion; IMHO develop should never point to a commit that does not have the dev suffix.)

mdellweg:
Thank you for taking care of the release!

PR #68, opened 2021-01-18
Use click.utils.get_app_dir to find config

mdellweg:
[noissue]


mdellweg:
This may make the app more portable to other platforms. ;)

daviddavis (commented):
:+1:

daviddavis (approved):

PR #71, opened 2021-01-19
Lazy evaluation

mdellweg:


pulpcore/cli/core/context.py
```diff
@@ -75,28 +74,45 @@ class PulpExporterContext(PulpEntityContext):
...
+        if not self.pulp_ctx.has_plugin("pulpcore", min_version="3.10.dev0"):
+            # Workaround for improperly rendered nested resource paths and weird HREF names
+            # https://github.com/pulp/pulpcore/pull/1066
+            if parameters is None:
+                parameters = {}
+            parameters[self.HREF] = self.exporter["pulp_href"]
+        return super().create(parameters=parameters, body=body)
+
+    @property
+    def HREF(self) -> str:  # type: ignore
```
mdellweg, line 102: I'm not proud of tricking the typing system here. Maybe we need to wrap the HREF in a more generally used property we can override here.

pulpcore/cli/core/context.py
```diff
@@ -75,28 +74,45 @@ class PulpExporterContext(PulpEntityContext):
...
+            return "core_pulp_pulp_export_href"
+        return "pulp_pulp_export_href"
+
+    @property
+    def scope(self) -> Dict[str, Any]:
+        if not self.pulp_ctx.has_plugin("pulpcore", min_version="3.10.dev0"):
+            # Workaround for improperly rendered nested resource paths and weird HREF names
+            # https://github.com/pulp/pulpcore/pull/1066
+            return {}
+        return {PulpExporterContext.HREF: self.exporter["pulp_href"]}
```
mdellweg, line 115: @ggainey I know, this sounds like going back and forth on how we do the trickery here. But at creation time of this Context we do not want to ask the server yet for its plugins.

mdellweg:
I added a small section to the README to explain the requirement for the help pages.

And there i think, this is ready for review @daviddavis, @ggainey, @dkliban

README.md
```diff
@@ -105,11 +99,33 @@ eval "$(_PULP_COMPLETE=source_bash pulp)"
 ### Code conventions
 `pulp-cli` comes with python type annotations and black code formatting.
 To verify your code please run `black`, `flake8`, `shellcheck`, and `mypy`.
-
-### Compatibility
+Run `make lint` to check for complience.
```
daviddavis, line 102:
```suggestion
Run `make lint` to check for compliance.
```

daviddavis:
The code here looks good to me. I'm not totally sure I totally understand the value of this feature though? Part of the reason I'm asking is that this would seem to preclude the future possiblity of populating our help text screens with information from the API. I see that you call this out in [#7787](https://pulp.plan.io/issues/7787). Do you have any thoughts on how we can accomplish both?

mdellweg:
> The code here looks good to me. I'm not totally sure I totally understand the value of this feature though? Part of the reason I'm asking is that this would seem to preclude the future possiblity of populating our help text screens with information from the API. I see that you call this out in [#8118](https://pulp.plan.io/issues/8118). Do you have any thoughts on how we can accomplish both?

The goal is to be able to access all commands help screens without the need to provide a pulp server. Also I think taking the docstrings from the API might be undesireable, because they describe certain API calls in a level of detail that might be unappropriate for the corresponding cli command. An example can be found in this discussion:
https://github.com/pulp/pulp-cli/pull/63#discussion_r561993298

daviddavis:
Thank you. I agree that the API text is not always what we want to display in the CLI's help screens. I assume the direction you're proposing is that we write and maintain our own set of help text in the CLI? 

We had this same sort of design in Pulp 2 and there were a lot of gaps in the CLI compared to the API. When I look at say [the file remote creation endpoint](https://docs.pulpproject.org/pulpcore/restapi.html#operation/remotes_file_file_partial_update) compared to the CLI command (which has just url and name), there are a lot of options and help text there we ought to display but aren't. I do think this is a problem we ought to solve (how to make it easier to define help text) but I'm willing to accept that it's perhaps a future one.

I think this over more this weekend and perhaps @ggainey and/or @dkliban want to weigh in as well?

mdellweg:
I should add, that the code itself does not rule it out, but the policy that "a help screen must not call out to a server" does.
OTOH i can vividly imagine us hacking into click's help information gathering and extend it to incorporate boilerplate text modules generated by / with the context modules.

Think `"Name of the {entity_type}.".format(entity_ctx.ENTITY)`.

daviddavis (approved):

PR #72, opened 2021-01-19
Updating release instructions

daviddavis:
[noissue]

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,12 +113,12 @@ So it might be needed to guard certain features / workaround by the available se
 
 ## Releasing
 
-1. Run `bumpversion release`
-1. Generate the changelog (eg `towncrier --yes --version 0.1.0`)
-1. Commit your local changes, open a PR, and get it merged.
+1. Generate the changelog (eg `towncrier --yes --version 0.1.0`) and commit
+1. Run `bumpversion release`, commit your local changes, and note the commit sha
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,12 +113,12 @@ So it might be needed to guard certain features / workaround by the available se
 
 ## Releasing
 
-1. Run `bumpversion release`
-1. Generate the changelog (eg `towncrier --yes --version 0.1.0`)
-1. Commit your local changes, open a PR, and get it merged.
+1. Generate the changelog (eg `towncrier --yes --version 0.1.0`) and commit
```
mdellweg, line 116: With this parameter towncrier is obviously not guessing the version, but could it?

mdellweg, line 116: With this parameter towncrier is obviously not guessing the version, but could it?

mdellweg (approved):
Not sure if we can make life easier around towncrier, so i am approving this as is.

README.md
daviddavis: The problem is that the changelog commit needs to be independent of the commit that sets the release version, and the changelog commit should proceed the release commit. It's possible the workflow could be:

1. Run `bumpversion release`
2. Generate the changelog (eg `towncrier --yes --version 0.1.0`) and commit only the changelog
3. Commit the other files

I don't know that this is any simpler.

PR #73, opened 2021-01-19
Adding pulp_python support

gerrod3:
Pulp Python is very similar to Pulp File, so there isn't too much that needed to be added to fulfill all the functionalities of Pulp Python. I'm going to leave this as a draft until I've finished the tests and #63 has been merged.

[noissue]

### DIFF
--- a/pulpcore/cli/python/__init__.py
+++ b/pulpcore/cli/python/__init__.py
@@ -1,0 +1,20 @@
+from pulpcore.cli.common import main
+from pulpcore.cli.common.context import PulpContext, pass_pulp_context
+from pulpcore.cli.python.content import content
+from pulpcore.cli.python.distribution import distribution
+from pulpcore.cli.python.publication import publication
+from pulpcore.cli.python.remote import remote
+from pulpcore.cli.python.repository import repository
+
+
+@main.group(name="python")
+@pass_pulp_context
+def python_group(pulp_ctx: PulpContext) -> None:
+    pulp_ctx.needs_plugin("python", min_version="3.1.0.dev")
+
+
+python_group.add_command(repository)
+python_group.add_command(remote)
+python_group.add_command(publication)
+python_group.add_command(distribution)
+python_group.add_command(content)
--- a/pulpcore/cli/python/content.py
+++ b/pulpcore/cli/python/content.py
@@ -1,0 +1,86 @@
+import gettext
+from typing import IO, Optional, Union
+
+import click
+
+from pulpcore.cli.common.context import (
+    PulpContext,
+    PulpEntityContext,
+    pass_entity_context,
+    pass_pulp_context,
+)
+from pulpcore.cli.common.generic import create_command, href_option, list_command, show_command
+from pulpcore.cli.core.context import PulpArtifactContext
+from pulpcore.cli.python.context import PulpPythonContentContext
+
+_ = gettext.gettext
+
+
+def _sha256_artifact_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[str]
+) -> Optional[Union[str, PulpEntityContext]]:
+    # Pass None and "" verbatim
+    if value:
+        pulp_ctx: PulpContext = ctx.find_object(PulpContext)
+        return PulpArtifactContext(pulp_ctx, entity={"sha256": value})
+    return value
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "content_type",
+    type=click.Choice(["package"], case_sensitive=False),
+    default="package",
+)
+@pass_pulp_context
+@click.pass_context
+def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> None:
+    if content_type == "package":
+        ctx.obj = PulpPythonContentContext(pulp_ctx)
+    else:
+        raise NotImplementedError()
+
+
+create_options = [
+    click.option("--relative-path", required=True, help=_("Exact name of file")),
+    click.option(
+        "--sha256",
+        "artifact",
+        required=True,
+        help=_("Digest of the artifact to use"),
+        callback=_sha256_artifact_callback,
+    ),
+]
+content.add_command(
+    list_command(
+        decorators=[
+            click.option("--filename", type=str),
+        ]
+    )
+)
+content.add_command(show_command(decorators=[href_option]))
+content.add_command(create_command(decorators=create_options))
+
+
+@content.command()
+@click.option("--relative-path", required=True, help=_("Exact name of file"))
+@click.option("--file", type=click.File("rb"), required=True, help=_("Path to file"))
+@click.option(
+    "--chunk-size", default=1000000, type=int, help=_("Chunk size in bytes (default is 1 MB)")
+)
+@pass_entity_context
+@pass_pulp_context
+def upload(
+    pulp_ctx: PulpContext,
+    entity_ctx: PulpPythonContentContext,
+    relative_path: str,
+    file: IO[bytes],
+    chunk_size: int,
+) -> None:
+    """Create a Python package content unit through uploading a file"""
+    artifact_href = PulpArtifactContext(pulp_ctx).upload(file, chunk_size)
+    content = {"relative_path": relative_path, "artifact": artifact_href}
+    result = entity_ctx.create(body=content)
+    pulp_ctx.output_result(result)
--- a/pulpcore/cli/python/context.py
+++ b/pulpcore/cli/python/context.py
@@ -1,0 +1,93 @@
+from typing import ClassVar
+
+from pulpcore.cli.common.context import (
+    EntityDefinition,
+    PulpEntityContext,
+    PulpRepositoryContext,
+    PulpRepositoryVersionContext,
+)
+
+
+class PulpPythonContentContext(PulpEntityContext):
+    ENTITY = "python content"
+    ENTITIES = "python content"
+    HREF = "python_python_package_content_href"
+    LIST_ID = "content_python_packages_list"
+    READ_ID = "content_python_packages_read"
+    CREATE_ID = "content_python_packages_create"
+
+
+class PulpPythonDistributionContext(PulpEntityContext):
+    ENTITY = "python distribution"
+    ENTITIES = "python distributions"
+    HREF = "python_python_distribution_href"
+    LIST_ID = "distributions_python_pypi_list"
+    READ_ID = "distributions_python_pypi_read"
+    CREATE_ID = "distributions_python_pypi_create"
+    UPDATE_ID = "distributions_python_pypi_partial_update"
+    DELETE_ID = "distributions_python_pypi_delete"
+
+    def preprocess_body(self, body: EntityDefinition) -> EntityDefinition:
+        body = super().preprocess_body(body)
+        for nullable in [
+            "publication",
+        ]:
+            if body.get(nullable) == "":
+                body[nullable] = None
+        return body
+
+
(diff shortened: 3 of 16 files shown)


mdellweg:
I'm happy to merge any uncontroversial parts without waiting for #63.
Also there is almost no possibility for conflicts (maybe despite the single line in setup.py) between different content plugins.

pulpcore/cli/python/content.py
```diff
@@ -0,0 +1,150 @@
...
+from pulpcore.cli.python.context import PulpPythonContentContext
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "content_type",
+    type=click.Choice(["python"], case_sensitive=False),
+    default="python",
```
daviddavis, line 25: Isn't the content type a "package"?

pulpcore/cli/python/content.py
```diff
@@ -0,0 +1,159 @@
...
+]
+content.add_command(
+    list_command(
+        decorators=[
+            click.option("--filename", type=str),
+        ]
+    )
+)
+content.add_command(show_command(decorators=[href_option]))
+content.add_command(create_command(decorators=create_options + python_content_options))
```
mdellweg, line 130: I know, this is probably, how the python plugin works. But I would expect that is tell it to "take that artifact and turn it into a python package" and the server would analyse the file to get all the information from `setup.py`.

gerrod3:
@mdellweg I think the reason the final test doesn't pass is because the CI uses an older version of jq. https://github.com/pulp/pulp-cli/pull/73/checks?check_run_id=1913835565. I ran the tests on my machine with jq-1.6 and so did @daviddavis and we both got it to pass.

mdellweg:
> @mdellweg I think the reason the final test doesn't pass is because the CI uses an older version of jq. https://github.com/pulp/pulp-cli/pull/73/checks?check_run_id=1913835565. I ran the tests on my machine with jq-1.6 and so did @daviddavis and we both got it to pass.

It's only failing in one out of 5 tests, and the other one with the same Python version is just fine. It's quite hard to believe that this is a problem of the jq version.
Can you see, who is returning those unescaped characters? If it's the cli, we might have a serious problem.
`parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 376, column 1`

pulpcore/cli/python/content.py
```diff
@@ -0,0 +1,175 @@
...
+    click.option(
+        "--maintainer",
+        help=_(
+            "The maintainer's name at a minimum; additional contact information may be provided."
+        ),
+    ),
+    click.option("--maintainer_email", help=_("The maintainer's e-mail address.")),
+    click.option("--license", help=_("Text indicating the license covering the distribution")),
+    click.option(
+        "--requires_python",
```
mdellweg, line 91:
```suggestion
        "--requires-python",
```
This holds true for the other options as well.

pulpcore/cli/python/content.py
```diff
@@ -0,0 +1,175 @@
...
+)
+@pass_entity_context
+@pass_pulp_context
+def upload(
+    pulp_ctx: PulpContext,
+    entity_ctx: PulpPythonContentContext,
+    relative_path: str,
+    file: IO[bytes],
+    chunk_size: int,
+    **kwargs: Any,
```
mdellweg, line 168:
```suggestion
```

pulpcore/cli/python/context.py
```diff
@@ -0,0 +1,102 @@
...
+    UPDATE_ID = "remotes_python_python_partial_update"
+    DELETE_ID = "remotes_python_python_delete"
+
+    def preprocess_body(self, body: EntityDefinition) -> EntityDefinition:
+        body = super().preprocess_body(body)
+
+        if body.get("includes"):
+            body["includes"] = json.loads(body["includes"])
+        if body.get("excludes"):
+            body["excludes"] = json.loads(body["excludes"])
```
mdellweg, line 74: Good for now...

pulpcore/cli/python/remote.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+lookup_options = [href_option, name_option]
+create_options = [
+    click.option("--name", required=True),
+]
+python_remote_options = [
+    click.option("--url"),
+    click.option("--includes", help=_("Package allowlist")),
+    click.option("--excludes", help=_("Package blocklist")),
```
mdellweg, line 46: Are those also JSON lists?

pulpcore/cli/python/repository.py
```diff
@@ -0,0 +1,166 @@
...
+    repository_ctx.sync(
+        href=repository_href,
+        body=body,
+    )
+
+
+@repository.command()
+@name_option
+@href_option
+@click.option("--relative-path", required=True)
```
mdellweg, line 112: Would it be nicer to call this "filename"? (It seems to be "filename" on the content model.)

tests/scripts/pulp_python/test_content.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
+
+pulp debug has-plugin --name "pulp_python" --min-version "3.1.0.dev" || exit 3
```
mdellweg, line 6: OK, that might be the reason, the other runs don't fail.
Why do we only allow those tests with 3.1?

And still returning insane output from the CLI is a no no. We might not have seen this before, because shelf reader is the first content with special characters in the header, which let's me to believe this is a general problem.

tests/scripts/pulp_python/test_repository.sh
```diff
@@ -0,0 +1,46 @@
...
+  test "$(echo "$OUTPUT" | jq -r '.|length')" != "0"
+  expect_succ pulp repository list --name "cli_test_python_repo"
+  test "$(echo "$OUTPUT" | jq -r '.|length')" = "1"
+  expect_succ pulp repository list --name-contains "cli_test_python"
+  test "$(echo "$OUTPUT" | jq -r '.|length')" = "1"
+  expect_succ pulp repository list --name-icontains "CLI_test_python"
+  test "$(echo "$OUTPUT" | jq -r '.|length')" = "1"
+  expect_succ pulp repository list --name-in "cli_test_python_repo"
+  test "$(echo "$OUTPUT" | jq -r '.|length')" = "1"
+fi
```
mdellweg, line 45: This looks like test for core, i think we should remove it here.

If you took this from the corresponding file test, i admit bad design on my behalf.

tests/scripts/pulp_python/test_sync.sh
```diff
@@ -0,0 +1,43 @@
...
+  pulp python remote destroy --name "cli_test_python_remote" || true
+  pulp python repository destroy --name "cli_test_python_repository" || true
+}
+trap cleanup EXIT
+
+cleanup
+
+# Prepare
+expect_succ pulp python remote create --name "cli_test_python_remote" --url "$PYTHON_REMOTE_URL" \
+  --includes '["shelf-reader", "aiohttp>=3.2.0,<3.3.1", "celery~=4.0", "Django>1.10.0,<=2.0.6"]'
```
mdellweg, line 18: If you tend to judge projects by their dependencies, this one looks super bizarre. :exploding_head:

mdellweg (changes requested):
Very nice over all.
And thanks for the _.

pulpcore/cli/python/remote.py
gerrod3: Yes they are json lists

tests/scripts/pulp_python/test_content.sh
gerrod3: I needed to add some [fixes](https://github.com/pulp/pulp_python/pull/327) to the Python plugin to get the CLI to work.

mdellweg: Ah, that's the why only 3.1. Check!

pulpcore/cli/python/remote.py
mdellweg: So they can use the json option callback once we merged the access policy PR.
#104

pulpcore/cli/python/content.py
gerrod3: You are right, the upload api does get the information from the package. It even overwrites whatever you pass in with the info from the package, so having these options is misleading, I'll remove them.

pulpcore/cli/python/context.py
gerrod3: Removed to now use load_json_callback.

gerrod3:
Finally got this ready for review! Are we planning to release 0.7.0 this week? I would like to release a new Pulp Python version at the same time to correspond with it.

mdellweg:
Having python support is deffinitely a good reason to make a release.
OTOH i do not intend to sync the releases of the cli with anything on the server side. Rather, try to use `has_plugin` to make the cli compatible with as many server versions as feasible, and mark the rest incompatible (feature by feature) with `needs_plugin`.

pulpcore/cli/python/context.py
```diff
@@ -0,0 +1,91 @@
...
+        return body
+
+
+class PulpPythonRemoteContext(PulpEntityContext):
+    ENTITY = "python remote"
+    ENTITIES = "python remotes"
+    HREF = "python_python_remote_href"
+    LIST_ID = "remotes_python_python_list"
+    CREATE_ID = "remotes_python_python_create"
+    BANDERSNATCH_ID = "remotes_python_python_from_bandersnatch"
```
mdellweg, line 62: This could use some type annotation. `ClassVar[str]` i believe.
Also i'd expect a method on this class to use it.
Are we using it yet?

tests/scripts/pulp_python/test_content.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
+
+pulp debug has-plugin --name "python" --min-version "3.1.0.dev" || exit 3
```
mdellweg, line 6: Do we really need to skip all those tests with versions less than 3.1?
If yes, this means the commands only start to work properly with 3.1. And in that case, you should require that version in `__init__.py`.

https://github.com/pulp/pulp-cli/pull/73/files#diff-607abf748c857c19b5f894cd196a77a9429d608fe1b80c2b666e45f909ed029fR13

mdellweg (changes requested):
This all looks really good to me. Just two things are left now, i think.

pulpcore/cli/python/context.py
mdellweg:
```suggestion
    BANDERSNATCH_ID: ClassVar[str] = "remotes_python_python_from_bandersnatch"
```
And i see we aren't using it by now. So don't bother about that method just yet.

pulpcore/cli/python/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+from pulpcore.cli.python.distribution import distribution
+from pulpcore.cli.python.publication import publication
+from pulpcore.cli.python.remote import remote
+from pulpcore.cli.python.repository import repository
+
+
+@main.group(name="python")
+@pass_pulp_context
+def python_group(pulp_ctx: PulpContext) -> None:
+    pulp_ctx.needs_plugin("python", min_version="3.1.0.dev")
```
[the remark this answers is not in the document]
gerrod3: Yes, it currently is.

PR #74, opened 2021-01-20
Make container push repositories read-only

mdellweg:
ContainerPushRepository is going to be changed. So this removes all
writing action from the commands and to unblock our ci removes all tests
for this repository type for now. We can add them back when the
pulp_container plugin has settled.

[noissue]


ipanova (approved):
I just want to highlight that we might end up adding back post to push repos due to this request https://pulp.plan.io/issues/8104
AH confirmed that this is something they will need.

mdellweg:
@ipanova thanks fro bringing it up, but as i read it, that refers to features the cli din't know of by this day.

Maybe i didn't make it clear that this only removes create/update/delete for push repos. You can still show/list them.

PR #77, opened 2021-01-21
Switched update-ids to use partial updates (PATCH)

gerrod3:
[noissue]


mdellweg:
What about `UPDATE_ID = "uploads_update"` ?

gerrod3:
> What about `UPDATE_ID = "uploads_update"` ?

There is no uploads_partial_update. This entity's update command is more of an upload command disguised as an update.

mdellweg (approved):
Agreed.

PR #83, opened 2021-01-25
Refactor help pages test

mdellweg:
Traverse all the commands recursively to find all help pages.

This test ensures, that all help pages can be rendered without throwing an exception and without internet connection.


daviddavis (approved):
:+1:

Issue #85, opened 2021-01-25
Using -b with async ansible commands errors or gives incorrect results


gerrod3:
The '-b' or '--background' option runs commands without waiting for any tasks generated by the commands to complete. Some of the async ansible commands show the results of the tasks expecting them to be finished.  Using the '-b' option skips the waiting causing them to error or show an incorrect result. The commands currently affected are:

- distribution create & update
- remote update
- repository update

Solution:
Use the passed pulp_context to check if the '-b' option is set and then appropriately show the results of the task or not.

mdellweg:
I believe we found a deeper problem in the design how we handle tasks here. Maybe, with `background=True`, the task loop should exit in the same way as hitting ctrl-c, with an exception.

PR #86, opened 2021-01-25
Fixes -b usuage on async ansible commands

gerrod3:
fixes:  #85

### DIFF
--- a/pulpcore/cli/ansible/distribution.py
+++ b/pulpcore/cli/ansible/distribution.py
@@ -62,9 +62,11 @@ def create(
         body["repository_version"] = f'{repo["versions_href"]}{version}/'
     elif repository:
         body["repository"] = repo["pulp_href"]
+
     result = distribution_ctx.create(body=body)
-    distribution = distribution_ctx.show(result["created_resources"][0])
-    pulp_ctx.output_result(distribution)
+    if not pulp_ctx.background_tasks:
+        distribution = distribution_ctx.show(result["created_resources"][0])
+        pulp_ctx.output_result(distribution)
 
 
 @distribution.command()
@@ -127,5 +129,6 @@ def update(
                 f"please specify the repository to use  with --repository"
             )
     distribution_ctx.update(href, body=body)
-    dist_body = distribution_ctx.show(href)
-    pulp_ctx.output_result(dist_body)
+    if not pulp_ctx.background_tasks:
+        dist_body = distribution_ctx.show(href)
+        pulp_ctx.output_result(dist_body)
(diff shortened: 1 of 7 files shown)


pulpcore/cli/ansible/distribution.py
```diff
@@ -67,9 +67,11 @@ def create(
         body["repository_version"] = f'{repo["versions_href"]}{version}/'
     elif repository:
         body["repository"] = repo["pulp_href"]
+
     result = distribution_ctx.create(body=body)
-    distribution = distribution_ctx.show(result["created_resources"][0])
-    pulp_ctx.output_result(distribution)
+    if not pulp_ctx.background_tasks:
```
mdellweg, line 72: I've seen, that we don't output anything from `update` calls elsewhere.
Which is in line with other unix shell commands like `mv`, `cp` and even `rm`.

pulpcore/cli/ansible/distribution.py
```diff
@@ -112,16 +114,18 @@ def update(
             repo = PulpAnsibleRepositoryContext(pulp_ctx).find(name=repository)
             if version is not None:
                 if dist_body["repository"]:
-                    distribution_ctx.update(href, body={"repository": ""})
+                    pulp_ctx.wait_for_task(distribution_ctx.update(href, body={"repository": ""}))
```
mdellweg, line 117: This is because distributions cannot be updated "in one go" (and both file and rpm probably don't even have this problemas they facilitate publications)? I consider this a bug in pulpcore. And workarounds for bugs in pulpcore belong into the corresponding context classes.
Using `pulp_ctx.wait_for_task` is mixing several levels of abstraction together.

And the much bigger problem is, this is lying to the user, as it is not performing stuff in the background as requested.
i.e. if the resource is blocked by another long running task, the command will wait for that task, and then issue the second, which will very likely be executed immediately.
Does the pulp tasking system not allow to queue up both tasks, and they will be executed in that order yielding the expected result without awaiting any of them?
If all of this does not work as expected, i'd rather issue an error, that this command is not capable of "running in the background".

mdellweg (changes requested):

pulpcore/cli/ansible/distribution.py
gerrod3: According to dalley and david the tasks will be queued up and executed in order if they operate on the same resource. So I don't need the wait_for_task call after all, I'll get rid of them.

pulpcore/cli/ansible/distribution.py
gerrod3: You commented on the creation method which does show the output for other commands, but this one is the only one that is an async task.  As for the other update commands, if wait is on you might as well show the result of the task.

daviddavis:
@gerrod3 can you rebase?

gerrod3:
Closing since fixed in #130

PR #90, opened 2021-01-26
Move changelog and CI to Github Actions

### LINKED ISSUE #89: Move changelog template and issue checker workflow to Github Actions

daviddavis:
fixes #89

### DIFF
--- a/.ci/scripts/validate_commit_message.py
+++ b/.ci/scripts/validate_commit_message.py
@@ -1,36 +1,28 @@
 import re
-import requests
 import subprocess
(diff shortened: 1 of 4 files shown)


mdellweg (approved):
So this will check if there is a changelog for an issue or "[noissue]" as before?

daviddavis:
> So this will check if there is a changelog for an issue or "[noissue]" as before?

That's correct. I don't know if that makes sense anymore but I guess we can re-evaluate later..?

PR #82, opened 2021-01-22
Generic factory commands

mdellweg:
~This is based on the lazy entity from #71 .~

@daviddavis Have a look at the `pulp file repository show` command at this point of the experiment.

PR #97, opened 2021-01-27
Add command to show all repositories

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/cli/common/__init__.py
+++ b/pulpcore/cli/common/__init__.py
@@ -6,6 +6,7 @@ import pkg_resources
 import toml
 
 from pulpcore.cli.common.context import PulpContext
+from pulpcore.cli.common.debug import debug
 
 __version__ = "0.2.1.dev"
 
@@ -90,6 +91,9 @@ def main(
         debug_callback=_debug_callback,
     )
     ctx.obj = PulpContext(api_kwargs=api_kwargs, format=format, background_tasks=background)
+
+
+main.add_command(debug)
 
 
 ##############################################################################
--- a/pulpcore/cli/core/repository.py
+++ b/pulpcore/cli/core/repository.py
@@ -1,0 +1,37 @@
+import click
+
+from pulpcore.cli.common.context import PulpContext, PulpRepositoryContext, pass_pulp_context
+from pulpcore.cli.common.generic import list_command
+
+
+@click.group()
+@pass_pulp_context
+@click.pass_context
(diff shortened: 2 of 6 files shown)


mdellweg:
Tests pending. Also I'd like to add filtering for names once we have #82 .

pulpcore/cli/common/__init__.py
```diff
@@ -92,6 +93,17 @@ def _debug_callback(level: int, x: str) -> None:
     ctx.obj = PulpContext(api_kwargs=api_kwargs, format=format, background_tasks=background)
 
 
+@main.command()
+@click.option("--name", required=True)
+@click.option("--min-version")
+@click.option("--max-version")
+@pass_pulp_context
+def has_plugin(pulp_ctx: PulpContext, name: str, min_version: str, max_version: str) -> None:
```
mdellweg, line 101: @daviddavis this is an alternative approach to query for versions in the tests.

daviddavis: I use the api json schema a lot (since our help texts are rather incomplete) and was thinking of adding a command to return the api schema json. I wonder if maybe we should have a `debug` group for these commands? wdyt?

pulpcore/cli/core/repository.py
```diff
@@ -0,0 +1,37 @@
...
+from pulpcore.cli.common.context import PulpContext, PulpRepositoryContext, pass_pulp_context
+from pulpcore.cli.common.generic import list_command
+
+
+@click.group()
+@pass_pulp_context
+@click.pass_context
+def repository(ctx: click.Context, pulp_ctx: PulpContext) -> None:
+    """
+    Perform actions on all repositories.
```
daviddavis, line 12:
```suggestion
    List all repositories.
```
?

daviddavis: Oh duh, I see now there is a list command.

pulpcore/cli/common/__init__.py
mdellweg: love the debug idea.
We can include jq like postprocessing into the schema json. :thinking: or into all the outputs...

daviddavis: Cool, I'll approve this PR if you just want to merge as is.

daviddavis (approved):

pulpcore/cli/common/__init__.py
mdellweg: added "debug", and thrilled to see more helper there.

PR #98, opened 2021-01-27
Add isort

daviddavis:
[noissue]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,3 +8,7 @@ issue_format = "`#{issue} <https://github.com/pulp/pulp-cli/issues/{issue}>`_"
 
 [tool.black]
 line-length = 100
+
+[tool.isort]
+profile = "black"
+line_length = 100
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ black:
 lint:
 	find . -name '*.sh' -print0 | xargs -0 shellcheck -x
 	black --diff --check .
+	isort -c --diff .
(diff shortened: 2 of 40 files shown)


pyproject.toml
```diff
@@ -8,3 +8,8 @@ issue_format = "`#{issue} <https://github.com/pulp/pulp-cli/issues/{issue}>`_"
 
 [tool.black]
 line-length = 100
+
+[tool.isort]
+multi_line_output = 3
+include_trailing_comma = true
```
mdellweg, line 14:
```suggestion
profile = "black"
multi_line_output = 3
```

The docs say, you should use something like this.

mdellweg (approved):
I was asking myself before, whether we should move the linting from the matrixed test workflows to the single check workflow. WDYT?

daviddavis:
> I was asking myself before, whether we should move the linting from the matrixed test workflows to the single check workflow. WDYT?

I like the idea of having a separate workflow for linting. What do you mean by "single check"? I kind of like how we use different steps for different linters though as it makes it easy to scan a job and see what's failing.

mdellweg:
> > I was asking myself before, whether we should move the linting from the matrixed test workflows to the single check workflow. WDYT?
> 
> I like the idea of having a separate workflow for linting. What do you mean by "single check"? I kind of like how we use different steps for different linters though as it makes it easy to scan a job and see what's failing.

I meant the "issue checker" workflow.

PR #99, opened 2021-01-28
Move linting and issue checking to a lint workflow

daviddavis:
[noissue]


.github/workflows/lint.yaml
```diff
@@ -1,8 +1,8 @@
 ---
-name: Issue Checker
+name: Lint Checker
```
mdellweg, line 2:
```suggestion
name: Linting
```

mdellweg (approved):
One suggestion on the naming.

PR #102, opened 2021-01-28
Output raw data when piped

daviddavis:
[noissue]

### DIFF
--- a/tests/scripts/test_output.sh
+++ b/tests/scripts/test_output.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(realpath "$0")")/config.source"
+
+# skip this test if pygments is not installed
+pip show pygments > /dev/null || exit 3
+
+# the foreground color escape sequence is "ESC[38;5"
+output=$(script -qefc "pulp status")
+if [[ "$output" != *"[38;5"* ]]
+then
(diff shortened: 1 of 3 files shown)


mdellweg:
Why don't we just switch of the coloring on a tty?
I might want to pipe yaml to a tool that understands it.

daviddavis:
The only caveat is that we are also formatting the json but I removed that too by setting indent to `None` if `isatty`. Also, I added an `isatty` propery.

tests/scripts/test_output.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(realpath "$0")")/config.source"
+
+cleanup() {
+  pulp file repository destroy --name "cli_test_file_repo" || true
+}
+trap cleanup EXIT
+
+expect_succ pulp file repository create --name "cli_test_file_repo" --description "Test repository for CLI tests"
```
mdellweg, line 11: Is this handled in a way that $OUTPUT looks like it was on a TTY?

daviddavis: Nope, let me try something different.

mdellweg: So you found something?
I'd be happy to take it without a test, because faking a user session is hard...

daviddavis: I did. Check it out.

mdellweg (approved):

PR #105, opened 2021-01-31
Group tests in folders

daviddavis:
[noissue]

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,9 +91,19 @@ As a rule of thumb, all necessary workarounds should be implemented in the corre
 Tests are shell scripts in `tests/scripts` with names like `test_*.sh`.
 They should should focus on the cli operation and are not a replacement for pulp integration tests;
 i.e. make sure the cli translates to the right api calls, but do not care about pulp internals.
-In order to perform, a running instance of pulp with all necessary plugins installed must be configured in `tests/scripts/config/pulp/settings.toml`.
-All tests are excercised using `make test`.
-Individual tests can be selected by pattern via `pytest -k file`.
+
+#### Running Tests
+
+In order to run tests, a running instance of pulp with all necessary plugins installed must be
+configured in `tests/settings.toml`.
+
+To run tests:
+
+```
+make test                           # all tests
+pytest -m pulp_file                 # tests for pulp_file
+pytest -m pulp_file -k test_remote  # run tests/scripts/pulp_file/test_remote.sh
+```
 
 ## Releasing
 
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -1,21 +1,22 @@
-import os
 import subprocess
+from pathlib import Path
 
 import pytest
 
-TEST_NAMES = [
-    name[5:-3]
-    for name in os.listdir("tests/scripts")
-    if name.startswith("test_") and name.endswith(".sh")
+TEST_SCRIPTS = [
+    pytest.param(
+        path.resolve(),
+        id=path.stem,
+        marks=[] if path.parent.name == "scripts" else getattr(pytest.mark, path.parent.name),
+    )
+    for path in Path("tests/scripts").glob("**/test_*.sh")
 ]
 
 
-@pytest.mark.parametrize("test_name", TEST_NAMES)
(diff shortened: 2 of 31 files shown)


README.md
```diff
@@ -126,7 +126,7 @@ They should should focus on the cli operation and are not a replacement for pulp
 i.e. make sure the cli translates to the right api calls, but do not care about pulp internals.
 In order to perform, a running instance of pulp with all necessary plugins installed must be configured in `tests/scripts/config/pulp/settings.toml`.
 All tests are excercised using `make test`.
-Individual tests can be selected by pattern via `pytest -k file`.
+Individual tests can be selected by pattern via `pytest -k pulp_file`.
```
daviddavis, line 129: One thing I am trying to correct for is the possibility that tests outside of pulp_file might contain "file" which would match this expression.

mdellweg: What if we look for "test_file" then you could still have "test_rpm_do_sth_with_file".

tests/test_scripts.py
```diff
@@ -1,16 +1,14 @@
...
 import subprocess
+from pathlib import Path
 
 import pytest
 
 TEST_NAMES = [
-    name[5:-3]
-    for name in os.listdir("tests/scripts")
-    if name.startswith("test_") and name.endswith(".sh")
+    str(name)[14:-3].replace("/", ".") for name in Path("tests/scripts").glob("**/test_*.sh")
```
mdellweg, line 7: This will find any test script file in any nested subdirectory, right? This does not conform to how you do the relative lookup of the config in `config.source`.
What if we fold the tests inside out: Have a setup script to provide the needed functions and env-vars and then source a specific test. It would also remove some repeated weirdness from the tests. It may make it impossible to run the tests by simply calling that script however, but with `pytest -k` we should be ok there.

Is it problematic to have "/" in the test name? I'd rather not replace all "/" with dots an than all dots with "/", as it's not a noop.

daviddavis: > This will find any test script file in any nested subdirectory, right? This does not conform to how you do the relative lookup of the config in config.source.

How so? 

> Is it problematic to have "/" in the test name? I'd rather not replace all "/" with dots an than all dots with "/", as it's not a noop.

It is indeed a problem:

```
$ pytest -k "test/test"
ERROR: Wrong expression passed to '-k': test/test_remote: at column 5: unexpected character "/"
```

README.md
daviddavis: It's still a problem though if there's core test that starts with "file_..." since the core tests aren't namespaced.

tests/test_scripts.py
daviddavis: I see what you mean about config.source. I can look at the pattern you suggest of inverting config.source. Perhaps I could rename it as test_runner.sh and have it accept a list of `TESTS`?

mdellweg:
I think, i solved the marking and also the "/" problem in https://github.com/mdellweg/pulp-cli/tree/mark_tests.
Feel free to take it from there. I'm excited to see this one coming forward.

daviddavis:
Ha, I also worked on marking: https://github.com/daviddavis/pulp-cli/commit/65dfc7eb58484cf9c807124234d296c3beb4ebd9

mdellweg:
At least it looks similar. ;)

daviddavis:
Ok, I added in the mark changes.

README.md
```diff
@@ -91,9 +91,19 @@ As a rule of thumb, all necessary workarounds should be implemented in the corre
...
 They should should focus on the cli operation and are not a replacement for pulp integration tests;
 i.e. make sure the cli translates to the right api calls, but do not care about pulp internals.
-In order to perform, a running instance of pulp with all necessary plugins installed must be configured in `tests/scripts/config/pulp/settings.toml`.
-All tests are excercised using `make test`.
-Individual tests can be selected by pattern via `pytest -k file`.
+
+#### Running Tests
+
+In order to run tests, a running instance of pulp with all necessary plugins installed must be
+configured in `tests/settings.toml`.
```
mdellweg, line 98: good catch!

mdellweg (approved):

PR #106, opened 2021-02-01
Add some docs about architectural design

mdellweg:
I started to draft this and then i realized, there might be some overlap with what is in README.md.

Should we put all that stuff in README, or is it time to think about generating proper docs with navigating and searching?
(I sure hope that the _use_ of the cli is mainly selfexplanatory.)

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,96 @@
+# Pulp CLI Architecture
+
+## Plugin system
+
+The Pulp CLI is designed with a plugin structure. Plugins can either live in the pulp-cli package or be shipped independently.
+By convention, all parts of the CLI are packages in the open namespace `pulpcore.cli`.
+A plugin can register itself with the main app by specifying by specifying its main module as a `pulp_cli.plugins` entrypoint in `setup.py`.
+
+```python
+entry_points={
+    "pulp_cli.plugins": [
+        "myplugin=pulpcore.cli.myplugin",
+    ],
+}
+```
+
+The plugin can then attach subcommands to the `pulpcore.cli.common.main` command.
+
+```python
+from pulpcore.cli.common import main
+
+@main.command()
+def my_command():
+    pass
+```
+
+## Contexts
+
+In `click`, every subcommand is accompanied by a `click.Context`, and objects can be attached to them.
+In this CLI we attach a `PulpContext` to the main command.
+It acts as an abstraction layer that handles the communication to the pulp server through its `api` property.
+Further we encourage the handling of communication with certain endpoints by subclassing the `PulpEntityContext`.
+By attaching them to the contexts of certain command groups, they are accessible to commands via the `pass_entity_context` decorator.
+Those entity contexts should provide a common interface to the layer of `click` commands that define the user interaction.
+
+```python
+@main.command()
+@pass_pulp_context
+@click.pass_context
+def my_command(ctx, pulp_ctx):
+    ctx.obj = MyEntityContext(pulp_ctx)
+
+
+@my_command.command()
+@pass_entity_context
+def my_sub_command(entity_ctx):
+    ... href = ...
+    entity_ctx.destroy(href)
+```
+
+### Version dependent codepaths
+
+Each pulp CLI release is designed to support multiple pulp server versions.
+The entity contexts should be used to implement necessary server version-dependent workarounds.
+To facilitate this, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods.
+
+```python
+class MyEntityContext(PulpEntityContext):
+    def show(self, href):
+        if not self.pulp_ctx.has_plugin("pulp_my_content", min_version="1.2.3"):
+            return lookup_my_content_legacy(href)
+        else:
+            return super().show(href)
+
+
+@main.command()
+@pass_pulp_context
+@click.pass_context
+def my_command(ctx, pulp_ctx):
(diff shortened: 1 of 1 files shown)


daviddavis:
Is this ready for review?

daviddavis:
> I started to draft this and then i realized, there might be some overlap with what is in README.md.

> Should we put all that stuff in README, or is it time to think about generating proper docs with navigating and searching?

I think overlap is fine. I like having stuff separate from the README because I think it'll grow. We should think about publishing the docs I think at some point.

mdellweg:
Bring on the review!

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
+# Pulp CLI Architecture
+
+## Plugin system
+
+The Pulp CLI is designed with a plugin structure. Plugins can either live in the main package or be shipped independently.
```
daviddavis, line 5:
```suggestion
The Pulp CLI is designed with a plugin structure. Plugins can either live in the pulp-cli package or be shipped independently.
```

daviddavis: I'm assuming that `main` is the module name and pulp-cli is the name of the package. Or maybe I misunderstand?

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+from pulpcore.cli.common import main
+
+@main.command()
+def my_command():
+    pass
+```
+
+## Contexts
+
+In `click`, every subcommand is accompanied by a `click.Context`, and objects can be attached to them.
```
daviddavis, line 29: Is this always true? If not, I'd suggest:

```suggestion
In `click`, every subcommand is usually accompanied by a `click.Context`, and objects can be attached to them.
```

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+In this CLI we attach a `PulpContext` to the main command.
+It abstracts the communication to the pulp server through its `api` property.
```
daviddavis, line 31:
```suggestion
It handles the communication to the pulp server through its `api` property.
```

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+Further we encourage to abstract the communication with certain endpoints by subclassing the `PulpEntityContext`.
```
daviddavis, line 32:
```suggestion
Further we encourage the handling of communication with certain endpoints by subclassing the `PulpEntityContext`.
```

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+@pass_entity_context
+def my_sub_command(entity_ctx):
+    ... href = ...
+    entity_ctx.destroy(href)
+```
+
+### Version dependent codepaths
+
+The pulp CLI is designed to support a number of pulp server versions at the same time.
+The entity contexts should be used to implement necessary server version depedent workarounds.
```
daviddavis, line 54:
```suggestion
The entity contexts should be used to implement necessary server version-dependent workarounds.
```

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+@my_command.command()
+@pass_entity_context
+def my_sub_command(entity_ctx):
+    ... href = ...
+    entity_ctx.destroy(href)
+```
+
+### Version dependent codepaths
+
+The pulp CLI is designed to support a number of pulp server versions at the same time.
```
daviddavis, line 53:
```suggestion
Each pulp CLI release is designed to support multiple pulp server versions.
```

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+@pass_pulp_context
+@click.pass_context
+def my_command(ctx, pulp_ctx):
+    pulp_ctx.needs_plugin("pulp_my_content", min_version="1.0.0")
+    ctx.obj = MyEntityContext(pulp_ctx)
+```
+
+## Generics
+
+For certain regularly repeating patterns, like listing all entities of a kind, we provide generic commands that will use the underlying context objects.
```
daviddavis, line 76:
```suggestion
For certain often repeated patterns like listing all entities of a  particular kind, we provide generic commands that use the underlying context objects.
```

docs/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+
+```python
+from pulpcore.cli.common.generic import name_option, show_command,
+
+lookup_params = [name_option]
+my_command.add_command(show_command(decorators=lookup_params))
+```
+
+Some of these commands understand extra arguments.
+They can be attached by passing a list of `click.Options` via the decorators argument.
```
daviddavis, line 86:
```suggestion
They can be attached by passing a list of `click.Options` via the `decorators` argument.
```

daviddavis (approved):
Added some suggestions but nothing blocking.

docs/architecture.md
mdellweg: `main` is only the command entrypoint to be precise. And precise is good!

docs/architecture.md
mdellweg: It is. The command you are writing and decorating is only called via `ctx.invoke()`.
Optional is the use of `ctx.obj`.

docs/architecture.md
mdellweg: What about?
```suggestion
It acts as an abstraction layer that handles the communication to the pulp server through its `api` property.
```
Would this imply that we do not accept contributions that disregard the abstraction and call e.g. requests directly?

docs/architecture.md
mdellweg: trust a native speaker...

PR #110, opened 2021-02-03
Add nested task subgroup to file repository

mdellweg:


daviddavis:
Code lgtm. Is this waiting on a test?

mdellweg:
> Code lgtm. Is this waiting on a test?

I thought both a test as well as some more thought into design of the user facing command structure.

mdellweg:
OK,i just realized, that it suffers from the interspersed parameter syndrom.
Need to change a few things to make this work. :smiling_imp:

daviddavis:
I wonder if this pattern might be applicable to https://github.com/pulp/pulp-cli/issues/132

mdellweg:
> I wonder if this pattern might be applicable to #132

Maybe. We should follow a common design pattern here i think to keep the command structure consistent.

pulpcore/cli/file/repository.py
```diff
@@ -60,6 +70,9 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
 
 
 lookup_options = [href_option, name_option]
+nested_lookup_options = [
+    click.option("--repository", callback=_repository_callback, expose_value=False)
+]
```
mdellweg, line 75: I am wondering, if this kind of lookup should be used for the version and label command groups too.

pulpcore/cli/file/repository.py
```diff
@@ -32,6 +33,15 @@
 _ = gettext.gettext
 
 
+def _repository_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[str]
+) -> Optional[Union[str, PulpEntityContext]]:
```
daviddavis, line 38: Seems like it's simpler to use `Union[str, PulpEntityContext, None]` here. Also, I'm not sure how this would return a `PulpEntityContext`?

pulpcore/cli/file/repository.py
daviddavis: It might help to see some code. At least for labels, I can't visualize a benefit.

pulpcore/cli/file/repository.py
mdellweg: right!

pulpcore/cli/file/repository.py
mdellweg: The problem is, that we cannot use `--name` and `--href` here because the `TaskContext` would ingest them.
So we are left with `--repository` and maybe `--repository-href`. For `label` it would be for the sake of consistency.

daviddavis (approved):

daviddavis:
Does this keep the lookup options on other label subcommands (e.g. `pulp file remote label`) the same? Is the goal to switch those up too?

mdellweg:
> Does this keep the lookup options on other label subcommands (e.g. `pulp file remote label`) the same? Is the goal to switch those up too eventually?

Yes, in the end its meant to use the same structure everywhere.

daviddavis (approved):

PR #113, opened 2021-02-03
Move --repository in version command to the end

mdellweg:
Do i need an issue to add a changelog entry? Is the mechanism there already in place?

### DIFF
--- a/CHANGES/123.removal
+++ b/CHANGES/123.removal
@@ -1,0 +1,1 @@
+Moved the location of `--repository` option after the last command for version commands.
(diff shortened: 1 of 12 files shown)


daviddavis:
Here's your issue: https://github.com/pulp/pulp-cli/issues/123 :)

CHANGES/123.removal
```diff
@@ -0,0 +1 @@
+Moved the location of the `--version` command to the scope of the last command.
```
daviddavis, line 1: Doesn't this change move `--repository` to the end?

mdellweg: After the last command at least. There is no specific order of options.
How can i word it better?

daviddavis: What about "Moved the location of \`--repository\` option after the last command"?

daviddavis (approved):

PR #114, opened 2021-02-04
Add some config management commands

### LINKED ISSUE #111: Make initial config file creation easier
Ok so the config file is already pretty easy to make:  https://github.com/pulp/pulp-cli#configuration It would be easier still though if the user didn't have to make the dir and then then the config file. If they could just have to edit it to customize it instead of the default that would be easier.

Does it make any sense to do any of the following?

a) have the config file created by default on installation.

b) have a CLI command that would populate the config file (kind of like pulp smash has).

What do you think?

daviddavis:
fixes #111

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,0 +1,89 @@
+# Installing and configuration
+
+## Installation
+
+The pulp-cli package can be installed from a variety of sources. After installing, see the next
+section on how to configure pulp-cli.
+
+### From PyPI
+
+```
+pip install pulp-cli[pygments]  # colorized output
+pip install pulp-cli  # no color output
+```
+
+### From a source checkout
+
+```
+git clone <your_fork_url>
+cd pulp-cli
+pip install -e .
+```
+
+## Configuration
+
+The CLI can be configured by using a toml file.
+By default the location of this file is `~/.config/pulp/settings.toml`.
+However, this can be customized by using the `--config` option.
+Any settings supplied as options to a command will override these settings.
+
+Example file:
+
+```
+[cli]
+base_url = "https://pulp.dev"
+verify_ssl = false
+format = "json"
+```
+
+### Generate
+
+To generate a new configuration:
+
+```
+pulp config create --base_url "http://localhost"
+pulp config create -e  # with an editor (e.g. vim)
+pulp config create -i  # interactive prompts
+```
+
+### Validate
+
+The easiest way to validate your config is to call a command that interacts with your Pulp server:
+
+```
+pulp status
+```
+
+To statically validate your config file (check that it exists, that it has not erroneous values,
+etc) use the validate command:
+
+```
+pulp config validate
+pulp config validate --strict  # validates that all settings are present
+```
+
+### netrc
+
+If no user/pass is supplied either in the config file or as an option,
+then the CLI will attempt to use `~/.netrc`.
+Here is a `.netrc` example for localhost:
+
+```
+machine localhost
+login admin
+password password
+```
+
+### Katello
+
+If you have a Katello environment and wish to use pulp-cli to connect to Pulp, you'll need to
+configure client certificate authentication:
+
+```toml
+[cli]
+base_url = "https://<your FQDN>"
+cert = "/etc/pki/katello/certs/pulp-client.crt"
+key = "/etc/pki/katello/private/pulp-client.key"
+verify_ssl = false
+```
+
--- a/pulpcore/cli/core/config.py
+++ b/pulpcore/cli/core/config.py
@@ -1,0 +1,122 @@
+from pathlib import Path
+from typing import Any, Optional
+
+import click
+import toml
+
+from pulpcore.cli.common import config_options
+
+LOCATION = str(Path(click.utils.get_app_dir("pulp"), "settings.toml"))
+SETTINGS = ["base_url", "username", "password", "cert", "key", "verify_ssl", "format"]
+
+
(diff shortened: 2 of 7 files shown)


pulpcore/cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+        verify_ssl = (
+            click.prompt(
+                "Verify SSL Connection",
+                type=click.Choice(["yes", "no"], case_sensitive=False),
+                default="yes",
+            )
+            == "yes"
+        )
+
+    if not overwrite and Path(location).exists():
```
mdellweg, line 58: we should check before asking questions.

pulpcore/cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+    username: Optional[str],
+    password: Optional[str],
+    verify_ssl: bool,
+    format: str,
+) -> None:
+    if interactive:
+        location = click.prompt("Config location", default=LOCATION)
+        base_url = click.prompt("API base url", default="https://localhost")
+        username = click.prompt("Username", default="")
+        password = click.prompt("Password", default="")
```
mdellweg, line 43: We could fetch the defaults from the context below, i think. It would guarantee some consistency.

mdellweg:
Would you mind adding a `pulp config edit` command that just opens the existing config file in the editor, and maybe preseeding it in case it was nonexisting?

pulpcore/cli/core/config.py
daviddavis: Which context?

daviddavis:
> Would you mind adding a pulp config edit command that just opens the existing config file in the editor, and maybe preseeding it in case it was nonexisting?

Sure, yes. I think though it should raise an error if the file doesn't exist. I think the job of creating and seeding the file belongs to `create`.

docs/install.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+### Validate
+
+To validate an existing config:
+
+```
+pulp config validate
+pulp config validate --strict  # validates that all settings are present
+```
```
mdellweg, line 56: We could add a note that connectivity can be checked with `pulp status`, and may be that validation here is _only_ static analysis (is there a user friendly term for this?).

docs/install.md
```diff
@@ -0,0 +1,82 @@
...
+machine localhost
+login admin
+password password
+```
+
+### Katello
+
+If you have a Katello environment and wish to use pulp-cli to connect to Pulp, you'll need to
+configure client certificate authentication:
```
mdellweg, line 73: Not for this PR, but are we sure, that providing the cert will not collide with an existing netrc config?

pulpcore/cli/core/config.py
```diff
@@ -0,0 +1,117 @@
...
+        location = click.prompt("Config file location", default=LOCATION)
+        _check_location(location)
+        for setting in SETTINGS:
+            settings[setting] = prompt_config_option(setting)
+    else:
+        _check_location(location)
+        for setting in SETTINGS:
+            settings[setting] = locals()[setting] or ""
+
+    output = f"[cli]\n{toml.dumps(settings)}"
```
mdellweg, line 64: This looks hacky to me, not being a toml expert.
Would this also work?
```suggestion
    output = toml.dumps({"cli": settings})
```

pulpcore/cli/core/config.py
```diff
@@ -0,0 +1,117 @@
...
+            raise click.ClickException("No output from editor. Aborting.")
+
+    Path(location).parent.mkdir(parents=True, exist_ok=True)
+    with Path(location).open("w") as sfile:
+        sfile.write(output)
+
+    click.echo(f"Created config file at '{location}'.")
+
+
+@config.command(help="Open the settings config file in an editor")
```
mdellweg, line 78: Oh, you have peeked into my gettext experiments? :partying_face:

tests/scripts/test_config.sh
```diff
@@ -22,3 +31,16 @@ expect_fail pulp --username test --password test --client "/some/path" status
...
+
+# CONFIG COMMAND
+
+expect_fail pulp config edit --location $test_settings
+
+expect_succ pulp config create --location $test_settings --base-url "http://testing"
+cat $test_settings | grep 'base_url = "http://testing"'
+cat $test_settings | grep 'verify_ssl = true'
+
+expect_fail pulp config create --location $test_settings
```
mdellweg, line 44: `make me a sandwich`
`access denied`
`sudo make me a sandwich`

```suggestion
expect_fail pulp config create --location $test_settings
expect_succ pulp config create --location $test_settings --overwrite
```

pulpcore/cli/core/config.py
daviddavis: I must have looked at a bad example as this does work.

docs/install.md
daviddavis: Yes, I will investigate.

pulpcore/cli/core/config.py
mdellweg: I was thinking about this `ctx.default_map`.

https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/common/__init__.py#L23

docs/install.md
daviddavis: I filed an issue so I wouldn't forget https://github.com/pulp/pulp-cli/issues/135

mdellweg (approved):

PR #117, opened 2021-02-04
Fix release instructions

daviddavis:
Given that we don't have release branches and aren't doing patch releases, we should be bumping develop to the next y-release version.


mdellweg (approved):
I agree, it won't give much value to do point releases until 1.0.0.

PR #122, opened 2021-02-05
Add support for client cert auth

daviddavis:
[noissue]

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,19 @@ login admin
 password password
 ```
 
+### Katello
+
+If you have a Katello environment and wish to use pulp-cli to connect to Pulp, you'll need to
+configure client certificate authentication:
+
+```toml
+[cli]
+base_url = "https://<your FQDN>"
+cert = "/etc/pki/katello/certs/pulp-client.crt"
+key = "/etc/pki/katello/private/pulp-client.key"
+verify_ssl = false
+```
+
 ## Known issues
 
   * Redirecting from `http` to `https`, as done by a typical Pulp installation,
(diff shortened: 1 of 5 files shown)


pulpcore/cli/common/__init__.py
```diff
@@ -42,6 +42,10 @@ def _config_callback(ctx: click.Context, param: Any, value: Optional[str]) -> No
 @click.option("--base-url", default="https://localhost", help="Api base url")
 @click.option("--username", help="Username on pulp server")
 @click.option("--password", help="Password on pulp server")
+@click.option("--client-cert", help="Path to client certificate")
+@click.option(
+    "--client-key", help="Path to client private key. Not required if client-cert contains this."
+)
```
daviddavis, line 48: I'm tempted to use `--cert` and `--key` here since that's the options curl uses.

ianballou:
Working for me with Katello!

README.md
```diff
@@ -65,6 +65,19 @@ login admin
 password password
 ```
 
+### Katello
```
ianballou, line 68: Works for me! Tested with a Katello development box.

pulpcore/cli/common/__init__.py
mdellweg: Would work for me.

mdellweg: Now you missed those two appearances.

mdellweg (approved):
:cake:

mdellweg (approved):

PR #131, opened 2021-02-08
Skip tests, if plugin is not available

### LINKED ISSUE #91: Skip tests if a particular plugin is not installed

mdellweg:
fixes #91

### DIFF
--- a/CHANGES/91.misc
+++ b/CHANGES/91.misc
@@ -1,0 +1,1 @@
+Marked tests to be skipped if a needed plugin is unavaliable.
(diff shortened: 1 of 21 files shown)


daviddavis (approved):
LGTM. I guess once we have the test folder PR merged we can come up with something more elegant than adding `pulp debug has-plugin ...` to each test? Is that your thought as well?

mdellweg:
> LGTM. I guess once we have the test folder PR merged we can come up with something more elegant than adding `pulp debug has-plugin ...` to each test? Is that your thought as well?

I haven't thought that far yet. I thought, this is to enable developers with a subset of plugins to run `make test`. But you have a point there.

PR #136, opened 2021-02-10
Move test environment setup into a fixture

mdellweg:
Additionally, this will run the test scripts in a temporary directory.
In tmpdir handling in the config.script snippet was removed. It never
worked properly anyway. The pytest tmp_path fixture however provides
automatic cleanup for files older than three runs.
Another result is, that config.script no longer depends on any specific
location of files.

[noissue]


Makefile
```diff
@@ -9,10 +9,10 @@ lint:
 	mypy
 	@echo "🙊 Code 🙈 LGTM 🙉 !"
 
-tests/scripts/config/pulp/settings.toml:
+tests/settings.toml:
 	cp $@.example $@
 
-test: | tests/scripts/config/pulp/settings.toml
+test: | tests/settings.toml
```
mdellweg, line 15: This move is a minor inconvenience to current developers. But it makes this file so much more accessible to new contributors, that i think it is worth the trouble.

tests/scripts/config.source
```diff
@@ -1,26 +1,5 @@
-# Provide values from config
-settings="$(dirname "$(realpath "$0")")/config/pulp/settings.toml"
```
mdellweg, line 2: @daviddavis the riddance of this line should help you to unblock your test subdirectory story.

tests/scripts/config.source
```diff
@@ -1,26 +1,5 @@
...
-CONTAINER_REMOTE_URL="https://registry-1.docker.io"
-CONTAINER_IMAGE="hello-world"
-RPM_REMOTE_URL="${PULP_FIXTURES_URL}/rpm-unsigned"
-ANSIBLE_COLLECTION_REMOTE_URL="https://galaxy.ansible.com/"
-ANSIBLE_ROLE_REMOTE_URL="https://galaxy.ansible.com/api/v1/roles/?namespace__name=elastic"
-
 # Library for test helper functions
 
-TMP="$(mktemp -d)"
-trap "rm -rf $TMP" EXIT
```
mdellweg, line 22: Somehow this trap was never executed. Maybe because we set a new trap in every other test script.

tests/scripts/test_config.sh
```diff
@@ -4,19 +4,15 @@
...
 
-cleanup() {
-  rm $test_settings
-}
-trap cleanup EXIT
-
-
 # SETTINGS OPTION
 
+sed -e '/base_url/c\base_url = "http://badurl"' "$good_settings" > "$bad_settings"
```
mdellweg, line 14: I am pretty sure, this will match the expectations of our tests better than the static one.

tests/test_scripts.py
```diff
@@ -11,8 +11,11 @@
 
 
 @pytest.mark.parametrize("test_name", TEST_NAMES)
-def test_script(test_name):
-    run = subprocess.run([os.path.join("tests", "scripts", "test_" + test_name + ".sh")])
+def test_script(test_name, cli_settings, tmp_path):
+    run = subprocess.run(
+        [os.path.realpath(os.path.join("tests", "scripts", "test_" + test_name + ".sh"))],
+        cwd=tmp_path,
```
mdellweg, line 17: And this helps us to keep the pollution from some tests out of our git tree.

tests/scripts/test_file_content_bulk.sh
```diff
@@ -6,8 +6,6 @@
 pulp debug has-plugin --name "pulp_file" || exit 3
 
 cleanup() {
-  rm test_1.txt test_2.txt test_3.txt
-  rm add_content.json remove_content.json
```
mdellweg, line 10: Let's keep those files around a bit for post mortem analysis.
pytest will clean them up eventually.

tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    "CONTAINER_REMOTE_URL": "https://registry-1.docker.io",
+    "CONTAINER_IMAGE": "hello-world",
+    "RPM_REMOTE_URL": PULP_FIXTURES_URL + "/rpm-unsigned",
+    "ANSIBLE_COLLECTION_REMOTE_URL": "https://galaxy.ansible.com/",
+    "ANSIBLE_ROLE_REMOTE_URL": "https://galaxy.ansible.com/api/v1/roles/?namespace__name=elastic",
+}
+
+
+@pytest.fixture
+def cli_settings(tmp_path, monkeypatch):
```
daviddavis, line 22: Some docstring explaining what this function is doing might be nice. It took me a while to figure it out (had to read through the pytest docs) but then again, I'm no pytest expert.

tests/test_scripts.py
daviddavis: Does pytest clean up the tmp_path if the test fails?

tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    settings = toml.load("tests/settings.toml")
+    settings_path = tmp_path / "config" / "pulp"
+    settings_path.mkdir(parents=True)
+    with open(settings_path / "settings.toml", "w") as settings_file:
+        toml.dump(settings, settings_file)
+    monkeypatch.setenv("XDG_CONFIG_HOME", str(tmp_path / "config"))
+    monkeypatch.setenv("PULP_BASE_URL", settings["cli"]["base_url"])
+    monkeypatch.setenv("VERIFY_SSL", json.dumps(settings["cli"].get("verify_ssl", True)))
```
daviddavis, line 30: Why are you calling `json.dumps` here?

mdellweg: Because that yields lowercase true/false. `True` would break the scripts.
Maybe tolower would be better than trying to mimic toml with json. :nerd_face:

tests/test_scripts.py
mdellweg: pytest will keep the tmp_path for (defualt) 3 consecutive runs. Independent of failure.

tests/conftest.py
mdellweg: I'll write something.

Braindump: "Create a tmp_dir; place config there, point XDG_CONFIG_HOME, supply other usefull environment vars"

Maybe, i should place the other vars into their own fixture?

daviddavis: > Braindump: "Create a tmp_dir; place config there, point XDG_CONFIG_HOME, supply other usefull environment vars"

Sounds good.

> Maybe, i should place the other vars into their own fixture?

Up to you. I don't have a strong preference.

mdellweg: Maybe later. I just thought the name does not really match anymore.

daviddavis (approved):

PR #140, opened 2021-02-12
Change container commands to use generics

mdellweg:
This allows --href lookups.

[noissue]

### DIFF
--- a/pulpcore/cli/container/context.py
+++ b/pulpcore/cli/container/context.py
@@ -1,6 +1,7 @@
 import gettext
 
 from pulpcore.cli.common.context import (
+    EntityDefinition,
     PulpEntityContext,
     PulpRepositoryContext,
     PulpRepositoryVersionContext,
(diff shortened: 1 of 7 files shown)


pulpcore/cli/container/context.py
```diff
@@ -16,13 +18,28 @@ class PulpContainerDistributionContext(PulpEntityContext):
...
+    def preprocess_body(self, body: EntityDefinition) -> EntityDefinition:
+        body = super().preprocess_body(body)
+        for nullable in [
+            "ca_cert",
+            "client_cert",
+            "client_key",
+            "password",
+            "proxy_url",
+            "username",
+        ]:
```
mdellweg, line 38: There might be a remote chance, we get the names of those beasts from the schema.

pulpcore/cli/container/distribution.py
```diff
@@ -1,21 +1,53 @@
-from typing import Optional
+from typing import Literal, Optional, Union, cast
```
mdellweg, line 1: `Literal` seems to be too new. :(

daviddavis (approved):

PR #141, opened 2021-02-12
Use session fixture to setup settings

mdellweg:
everytime i look, i learn something new with pytest...

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,8 +17,23 @@ ENV_CONSTANTS = {
 }
 
 
+@pytest.fixture(scope="session")
+def pulp_cli_settings(tmp_path_factory):
+    """
+    This fixture will setup the config file once per session only.
+    It is most likely not useful to be included standalone.
+    The `pulp_cli_env` fixture, however depends on it and sets $XDG_CONFIG_HOME up accordingly.
+    """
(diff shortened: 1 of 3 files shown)


tests/scripts/config/pulp/settings.toml
```diff
@@ -1,7 +0,0 @@
-[cli]
-# base_url = "http://localhost:8080"
-base_url = "https://pulp3-sandbox-debian-testing"
-# base_url = "http://pulp3-source-debian-testing"
-username = "admin"
-password = "password"
-verify_ssl = false
```
daviddavis, line 7: :+1:

tests/conftest.py
```diff
@@ -17,8 +17,22 @@
 }
 
 
+@pytest.fixture(scope="session")
+def pulp_cli_settings(tmp_path_factory):
+    """
+    This session fixture will setup the config file once.
+    Use trough `pulp_cli_env` in order to setup XDG_CONFIG_HOME
```
daviddavis, line 24: Is this suppose to say "Use through"?

daviddavis (approved):

tests/conftest.py
mdellweg: Yes, but as it does not seem too obvious to you, let me rephrase it.

PR #144, opened 2021-02-13
Added worker command.

daviddavis:
fixes #144

### DIFF
--- a/CHANGES/144.feature
+++ b/CHANGES/144.feature
@@ -1,0 +1,1 @@
+Added worker command.
--- a/pulpcore/cli/core/worker.py
+++ b/pulpcore/cli/core/worker.py
@@ -1,0 +1,24 @@
+import click
+
+from pulpcore.cli.common.context import PulpContext, pass_pulp_context
+from pulpcore.cli.common.generic import href_option, list_command, name_option, show_command
+from pulpcore.cli.core.context import PulpWorkerContext
+
+
+@click.group()
+@pass_pulp_context
+@click.pass_context
+def worker(ctx: click.Context, pulp_ctx: PulpContext) -> None:
+    ctx.obj = PulpWorkerContext(pulp_ctx)
+
+
+filter_options = [
+    click.option("--name"),
+    click.option("--name-contains", "name__contains"),
+    click.option("--name-icontains", "name__icontains"),
+    click.option("--missing/--not-missing", default=None),
+    click.option("--online/--not-online", default=None),
+]
+
+worker.add_command(list_command(decorators=filter_options))
+worker.add_command(show_command(decorators=[href_option, name_option]))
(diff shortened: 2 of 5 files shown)


CHANGES/144.feature
```diff
@@ -0,0 +1 @@
+Added worker command.
```
mdellweg, line 1: Meh, self referencing does not seem to work. But i like the idea of having a changelog without the need to file an issue just to say: This is what i do in the PR anyway.

OTOH, i thought gh handles PRs as a special sort of issues.

It should work however if you do not put the `fixes` in the commit message. Which means we should add more checking for spurious changelog snippets. What if "(no [noissue]) and (no fixes #??)" implies we need one for the PR number?

log.out
```diff
@@ -0,0 +1 @@
+[{"pulp_created": "2021-02-13T13:33:04.220502Z", "pulp_href": "/pulp/api/v3/workers/3754d4ea-b554-432d-b20e-45026ce46c12/", "name": "resource-manager", "last_heartbeat": "2021-02-13T22:16:02.245766Z"}]
```
mdellweg, line 1: Those files i would not expect here. Is the cwd mechanism not working for you?

pulpcore/cli/core/worker.py
```diff
@@ -0,0 +1,27 @@
...
+filter_options = [
+    click.option("--name"),
+    click.option("--name-contains", "name__contains"),
+    click.option("--name-icontains", "name__icontains"),
+    click.option("--missing/--not-missing", default=None),
+    click.option("--online/--not-online", default=None),
+]
+
+worker.add_command(list_command(decorators=filter_options))
+worker.add_command(show_command(decorators=[href_option, name_option]))
```
mdellweg, line 27: Whenever i think, this is a lot of boilerplate, i need to remind myself, that it would be ca. 8-16 times the code if spelled out here in more detail...

pulpcore/cli/core/worker.py
```diff
@@ -0,0 +1,27 @@
...
+def worker(ctx: click.Context, pulp_ctx: PulpContext) -> None:
+    ctx.obj = PulpWorkerContext(pulp_ctx)
+
+
+filter_options = [
+    click.option("--name"),
+    click.option("--name-contains", "name__contains"),
+    click.option("--name-icontains", "name__icontains"),
+    click.option("--missing/--not-missing", default=None),
+    click.option("--online/--not-online", default=None),
```
mdellweg, line 23: So this is how you make a tristate option. :+1:

mdellweg (approved):
This all looks very good. We just need to find a convenient way to add changelogs without issues.

log.out
daviddavis: I don't know. I re-ran the tests and it didn't appear. Will keep an eye out.

mdellweg:
The approval stands. ;)

PR #145, opened 2021-02-14
Add config profile option

mdellweg:
This is a request for comment.
Do we want something like this?

### DIFF
--- a/tests/scripts/test_config.sh
+++ b/tests/scripts/test_config.sh
@@ -1,10 +1,11 @@
 #!/bin/sh
 
 # shellcheck source=tests/scripts/config.source
-. "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
+. "$(dirname "$(realpath "$0")")"/config.source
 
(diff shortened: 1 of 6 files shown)


daviddavis:
This looks good to me. :+1:

mdellweg:
I'm going to add some tests...

mdellweg:
I moved some tests i think are not testing pulpcore in any reasonable way out of the pulpcore test directory.
@daviddavis is this OK for you?

tests/scripts/test_config.sh
```diff
@@ -29,6 +30,14 @@ expect_fail pulp --username test --password test --client "/some/path" status
 expect_fail pulp --key "/some/path" file remote list
 
 
+# CONFFIG PROFILE
```
daviddavis, line 33:
```suggestion
# CONFIG PROFILE
```

daviddavis (approved):

tests/scripts/test_config.sh
mdellweg: Hmm this keyboard of mine is really oversensitive...

PR #147, opened 2021-02-15
Use gettext with help strings

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/cli/ansible/__init__.py
+++ b/pulpcore/cli/ansible/__init__.py
@@ -1,8 +1,12 @@
+import gettext
+
 from pulpcore.cli.ansible.distribution import distribution
 from pulpcore.cli.ansible.remote import remote
 from pulpcore.cli.ansible.repository import repository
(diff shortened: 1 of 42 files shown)


daviddavis (approved):
:man_dancing:

daviddavis:
Actually, why not do like we do in pulpcore?

```
from gettext import gettext as _
```

mdellweg:
I was reading more and more, and i think we want to use the class based approach.
So ATM even `_ = lambda (s): s` would to the trick.
I think, i want to setup `_` in every plugins `__init__` and import it from there.

mdellweg:
> I think, i want to setup `_` in every plugins `__init__` and import it from there.

Nope, circular dependency. We could use the `translation.install()` method. But we would need to find a way to merge the plugins translations into the global table. :exploding_head:

mdellweg:
I'm going to merge it as is, to get us in `_`-land soon. We will need to find a proper way to install the translation catalogs later.

PR #154, opened 2021-02-18
Fix ansible sync check

daviddavis:
[noissue]

### DIFF
--- a/tests/scripts/pulp_ansible/test_sync.sh
+++ b/tests/scripts/pulp_ansible/test_sync.sh
@@ -37,7 +37,7 @@ expect_succ pulp ansible repository sync --name "cli_test_ansible_repository" --
 expect_succ pulp ansible repository version list --repository "cli_test_ansible_repository"
 test "$(echo "$OUTPUT" | jq -r length)" -eq 2
 expect_succ pulp ansible repository version show --repository "cli_test_ansible_repository" --version 1
-test "$(echo "$OUTPUT" | jq -r '.content_summary.present."ansible.role".count')" -eq 56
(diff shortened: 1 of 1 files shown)


tests/scripts/pulp_ansible/test_sync.sh
```diff
@@ -37,7 +37,7 @@ expect_succ pulp ansible repository sync --name "cli_test_ansible_repository" --
 expect_succ pulp ansible repository version list --repository "cli_test_ansible_repository"
 test "$(echo "$OUTPUT" | jq -r length)" -eq 2
 expect_succ pulp ansible repository version show --repository "cli_test_ansible_repository" --version 1
-test "$(echo "$OUTPUT" | jq -r '.content_summary.present."ansible.role".count')" -eq 56
+test "$(echo "$OUTPUT" | jq -r '.content_summary.present."ansible.role".count')" -ge 57
```
mdellweg, line 40: So there was a change in the fixtures repo?

mdellweg (approved):
Thank you for analysing this.
I was suspecting the OCI-container to have pickled up a bug...

tests/scripts/pulp_ansible/test_sync.sh
daviddavis: We test against galaxy since syncing requires a live api. The role was updated (another role version was uploaded). We could look at using a different role that hasn’t changed or possibly a static fixture somehow. I can file an issue for this but for now I think we should just merge this change to fix the CI..?

mdellweg: Sure. We should merge this. But do we really need this check? Does it tell us anything about the tool under test (the cli) here? Then why that big a number? I'd go with "more than none" as a successful sync result.

PR #157, opened 2021-02-19
Modify ansible sync tests to check greater than 0

daviddavis:
[noissue]


mdellweg (approved):
Looks good. I don't see any more obvious such cases.

PR #159, opened 2021-02-20
Move permissions tests to pulpcore dir

daviddavis:
[noissue]


daviddavis:
I'm assuming that these belong in the pulpcore directory since they test pulpcore? It looks like they were maybe created before we created the subdirectories.

mdellweg:
Oh right. They probably need one more dirname though

mdellweg (approved):

PR #161, opened 2021-02-22
Provide a factory for resource options

### LINKED ISSUE #158: Allow to specify plugin and type along with the name for a dependent resource
The referencing schema for resources (like remote on repositories) should follow the pattern: `[[app_label:]type:]name`

Given that "file" is the default type, all those should be equivalent:
`pulp file repository update --name "test" --remote "test_remote"`
`pulp file repository update --name "test" --remote "file:test_remote"`
`pulp file repository update --name "test" --remote "file:file:test_remote"`

If the name has colons: `::name:with:colons`

mdellweg:
The options accept resources by the "[[<plugin>:]<type>:]name" pattern.

fixes #158

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -1,12 +1,13 @@
 import gettext
 import json
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union
 
 import click
 
 from pulpcore.cli.common.context import (
     DEFAULT_LIMIT,
     EntityDefinition,
+    EntityFieldDefinition,
     PluginRequirement,
     PulpContentContext,
     PulpContext,
@@ -21,6 +22,7 @@ from pulpcore.cli.common.context import (
 )
 
 _ = gettext.gettext
+
 _F = TypeVar("_F")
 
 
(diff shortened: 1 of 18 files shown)


mdellweg:
WIP: should be used in more places if the design is agreed upon.

gerrod3:
Should we limit the naming convention to only be used for options where a known resource, i.e. remotes, repositories, content, ..., is being asked for? Basically, should we have a global way to refer to a resource besides that resource's pulp_href? Something like "pulp show --name repository:file:file:foo". Maybe it doesn't make sense since there isn't many other operations we could do across all resources that aren't covered by their specific plugin commands already, but maybe you have some other ideas.

Other than that I think the design looks good!

pulpcore/cli/common/generic.py
```diff
@@ -170,6 +172,53 @@ def load_json_callback(
     cls=PulpOption,
 )
 
+
+def resource_option(*args: Any, **kwargs: Any) -> Callable[[F], F]:
+    default_plugin: str = kwargs.pop("default_plugin", None)
+    default_type: str = kwargs.pop("default_type", None)
```
daviddavis, line 178: :+1:

daviddavis:
This looks good to me so far.

daviddavis:
> The options accept resources by the "[[:]:]name" pattern.

Can you add test(s) for this?

daviddavis:
Also this needs a rebase. :)

mdellweg:
> > The options accept resources by the "[[:]:]name" pattern.
> 
> Can you add test(s) for this?

Tests are added.
You can now specify a remote of the name `test_1` as:
* `test_1`
* `:test_1`
* `::test_1`
* `file::test_1`
* `file:file:test_1`
* `:file:test_1`
and `test:2` as:
* `::test:2`
* `file::test:2`
* `file:file:test:2`
* `:file:test:2`

Obviously we discourage the use of `:` in names, but we can still handle it.

daviddavis (approved):
LGTM :+1:

Issue #134, opened 2021-02-08
Show distribution information by base_path


jlsherrill:
I think it would be useful to be able to lookup a yum distribution by its base path, something like:

pulp rpm distribution list --base-path=FOO
or
pulp rpm distribution show --base-path=FOO

PR #164, opened 2021-02-25
Added distribution lookup option --base-path

gerrod3:
fixes: #134

### DIFF
--- a/CHANGES/134.feature
+++ b/CHANGES/134.feature
@@ -1,0 +1,1 @@
+Distributions can now be listed with options --base-path and --base-path-contains
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -170,6 +170,21 @@ label_select_option = click.option(
     cls=PulpOption,
 )
 
+base_path_option = click.option(
+    "--base-path",
+    help=_("Base-path of the {entity}"),
+    type=str,
+    cls=PulpOption,
+)
+
+base_path_contains_option = click.option(
+    "--base-path-contains",
+    "base_path__icontains",
+    help=_("{entity} base-path contains search"),
+    type=str,
+    cls=PulpOption,
+)
+
 ##############################################################################
 # Generic reusable commands
 
--- a/tests/scripts/pulp_file/test_distribution.sh
+++ b/tests/scripts/pulp_file/test_distribution.sh
@@ -37,6 +37,12 @@ expect_succ pulp file distribution update \
   --base-path "cli_test_file_distro" \
   --publication "$PUBLICATION_HREF"
 
+
+expect_succ pulp file distribution list --base-path "cli_test_file_distro"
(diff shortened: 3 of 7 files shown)


CHANGES/134.feature
```diff
@@ -0,0 +1,2 @@
+Distributions can now be found with option --base-path for list and show commands
+
```
daviddavis, line 2:
```suggestion
```

tests/scripts/pulp_file/test_distribution.sh
```diff
@@ -37,6 +38,12 @@ expect_succ pulp file distribution update \
   --base-path "cli_test_file_distro" \
   --publication "$PUBLICATION_HREF"
 
+
+expect_succ pulp file distribution show --base-path "cli_test_file_distro"
+test "$(echo "$OUTPUT" | jq -r .pulp_href)" = "$DISTRIBUTION_HREF"
+expect_succ pulp file distribution list --base-path "cli"
+test "$(echo "$OUTPUT" | jq -r .[0].pulp_href)" = "$DISTRIBUTION_HREF"
```
daviddavis, line 45: This seems very brittle--I often run this test against my dev server which may have other file distributions. Maybe we could just check that the number of results is > 0?

daviddavis:
I had imagined that the `--base-path` would just be a filter on list. But now that I think about it, I can't see any downside I suppose to having it on show.

tests/scripts/pulp_file/test_distribution.sh
mdellweg: If the filter `--base-path "cli"` is working correctly, there must only be a single item returned. So we could even check for cout==1, no?

pulpcore/cli/common/generic.py
```diff
@@ -112,6 +112,19 @@ def load_json_callback(
...
+def _base_path_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[str]
+) -> Optional[str]:
+    if value is not None:
+        if ctx.info_name == "show":
+            entity_ctx: PulpEntityContext = ctx.find_object(PulpEntityContext)
+            entity_ctx.entity = {"base_path": value}
+        elif ctx.info_name == "list":
+            param.name = "base_path__icontains"
+            param.expose_value = True
```
mdellweg, line 124: I'd say, if you want a base_path__icontains filter, that make it a --base-path-contains option.

tests/scripts/pulp_file/test_distribution.sh
daviddavis: That's true altough the filter is `base_path__icontains` though so there's still a chance.

daviddavis:
We talked about this PR at our pulp-cli meeting and we thought maybe the `--base-path` option on show didn't make sense since we're not going to offer it on other endpoints such as update. It seems more of a filter than a lookup option?

daviddavis (approved):

mdellweg (approved):

PR #169, opened 2021-03-04
Add proxy username and password to remote

mdellweg:
This is combined with a workaround for pulpcore < 3.11 to insert the
credentials into the proxy_url.

[noissue]

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,7 +3,7 @@ strict = True
 show_error_codes = True
 files = pulpcore/**/*.py
 
-[mypy-urllib3]
+[mypy-urllib3.*]
 ignore_missing_imports = True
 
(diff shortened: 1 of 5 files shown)


mypy.ini
```diff
@@ -3,7 +3,7 @@ strict = True
 show_error_codes = True
 files = pulpcore/**/*.py
 
-[mypy-urllib3]
+[mypy-urllib3.*]
```
daviddavis, line 6: What does this change do?

daviddavis (approved):

mypy.ini
mdellweg: It made the linting pass. I think it's needed for the import from urllib3.utils

PR #173, opened 2021-03-10
Teach pulp-rpm-repository about retain-package-versions.

### LINKED ISSUE #172: Feature: expose retain-package-versions to pulp/rpm/repository

ggainey:
closes #172

### DIFF
--- a/tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
+++ b/tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
@@ -6,9 +6,9 @@
 pulp debug has-plugin --name "rpm" || exit 3
 
 cleanup() {
-  pulp rpm remote destroy --name "cli_test_rpm_remote" || true
-  pulp rpm repository destroy --name "cli_test_rpm_repository" || true
   pulp rpm distribution destroy --name "cli_test_rpm_distro" || true
+  pulp rpm repository destroy --name "cli_test_rpm_repository" || true
+  pulp rpm remote destroy --name "cli_test_rpm_remote" || true
 }
 trap cleanup EXIT
 
@@ -39,6 +39,10 @@ expect_succ curl "$curl_opt" --head --fail "$PULP_BASE_URL/pulp/content/cli_test
 
 expect_succ pulp rpm repository version list --repository "cli_test_rpm_repository"
 expect_succ pulp rpm repository version repair --repository "cli_test_rpm_repository" --version 1
+
+expect_succ pulp rpm repository update --name "cli_test_rpm_repository" --retain-package-versions 2
+expect_succ pulp rpm repository show --name "cli_test_rpm_repository"
+test "$(echo "$OUTPUT" | jq -r '.retain_package_versions')" = "2"
 
 expect_succ pulp rpm distribution destroy --name "cli_test_rpm_distro"
 expect_succ pulp rpm publication destroy --href "$PUBLICATION_HREF"
--- a/CHANGES/172.feature
+++ b/CHANGES/172.feature
@@ -1,0 +1,1 @@
+Taught rpm/repository about the retain-package-versions attribute.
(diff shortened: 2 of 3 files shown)


tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -6,9 +6,9 @@
 pulp debug has-plugin --name "rpm" || exit 3
 
 cleanup() {
-  pulp rpm remote destroy --name "cli_test_rpm_remote" || true
-  pulp rpm repository destroy --name "cli_test_rpm_repository" || true
   pulp rpm distribution destroy --name "cli_test_rpm_distro" || true
+  pulp rpm repository destroy --name "cli_test_rpm_repository" || true
+  pulp rpm remote destroy --name "cli_test_rpm_remote" || true
```
mdellweg, line 11: Cleanup in reverse order?
Fine for me! :+1:

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -40,6 +40,10 @@ expect_succ curl "$curl_opt" --head --fail "$PULP_BASE_URL/pulp/content/cli_test
 expect_succ pulp rpm repository version list --repository "cli_test_rpm_repository"
 expect_succ pulp rpm repository version repair --repository "cli_test_rpm_repository" --version 1
 
+expect_succ pulp rpm repository update --name "cli_test_rpm_repository" --retain-package-versions 2
+expect_succ pulp rpm repository show --name "cli_test_rpm_repository"
+test "$(echo "$OUTPUT" | jq -r '.retain_package_versions')" = "2"
```
mdellweg, line 45: To disable that function, do you nullify it, or will it be set to 0, -1, ...?
Can you add a test for that?

mdellweg (commented):
Is this a rather recently added feature in pulp_rpm?
If yes, do you think we can guard it's use by `has_plugin`?

edit: Probybly `needs_plugin` being lazy is the better choice.

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
ggainey: Set to 0 turns it off - will add a test, good call.

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
ggainey: Just making the cleanup() match the order at test-end :)

ggainey: > Is this a rather recently added feature in pulp_rpm?
> If yes, do you think we can guard it's use by `has_plugin`?
> 
> edit: Probybly `needs_plugin` being lazy is the better choice.

The feature has been in pulp_rpm since June , released in pulp_rpm 3.5 - does it really need a guard?

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
mdellweg: In that case (also being an integer) we probably don't need a test. For nullifying there can be interesting side effects.

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
mdellweg: June is older than any CLI-Release. I believe we agreed to support about 5 releases back. So assuming your test would fail on earlier versions of pulp_rpm (which it doesn't in the cli) i think we are safe.

mdellweg (approved):

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
ggainey: repushed a change constraining retain-package-versions to integer

PR #174, opened 2021-03-11
Remove deprecated commands

mdellweg:
Remove _by_name, _by_href and list_entities commands and replace their
use by the newer generics.

[noissue]


mdellweg:
Some long overdue cleanup.
But we have a problem. Our heuristic to recognize a triggered task is not sufficient, as exports have a task field.

gerrod3 (approved):
Looks good!

daviddavis (approved):

PR #175, opened 2021-03-12
Fixed python commands not working with pulpcore 3.10

gerrod3:
[noissue]

Forgot to add this.


pulpcore/cli/common/context.py
```diff
@@ -39,6 +39,7 @@
     core="pulpcore",
     deb="pulp_deb",
     file="pulp_file",
+    python="pulp_python",
```
mdellweg, line 42: Without that, the commands never worked with pulpcore <= 3.10, right?

mdellweg (approved):

pulpcore/cli/common/context.py
gerrod3: Yeah.

PR #180, opened 2021-03-16
Move common remote options to generic

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -3,7 +3,7 @@ import gettext
 import json
 import sys
 import time
-from typing import IO, Any, ClassVar, Dict, List, Optional, Tuple, Type, Union
+from typing import IO, Any, ClassVar, Dict, List, Optional, Set, Tuple, Type, Union
 
 import click
 import yaml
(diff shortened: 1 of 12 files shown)


pulpcore/cli/common/context.py
```diff
@@ -319,7 +321,11 @@ def __init__(
...
         return {
-            key: value.pulp_href if isinstance(value, PulpEntityContext) else value
+            key: None
+            if key in self.NULLABLES and value == ""
+            else value.pulp_href
+            if isinstance(value, PulpEntityContext)
+            else value
             for key, value in body.items()
             if value is not None
         }
```
daviddavis, line 331: I'm finding this code hard to parse but maybe it's because I don't often see ternaries in Python?

mdellweg: Is this a request to rewrite it?
I could turn it into a loop (which i don't want) or rewrite the quaternary in to a helper function (not really keen, but better).

daviddavis: A nested ternary in a dict comprehension seems very hard to read. What's the downside to using a loop?

mdellweg: Would look like this:
```
    def _preprocess_value(self, key: str, value: Any) -> Any
            if key in self.NULLABLES and value == "":
                return None
            if isinstance(value, PulpEntityContext):
                return value.pulp_href
            return value

    def preprocess_body(self, body: EntityDefinition) -> EntityDefinition:
        return {
            key: self._preprocess_value(key, value)
            for key, value in body.items()
            if value is not None
        }
```

daviddavis (approved):

PR #183, opened 2021-03-18
Add new pulp releasee to test matrix

mdellweg:
[noissue]

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,16 +17,18 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python: 3.9
-            image_tag: nightly
-          - python: 3.6
-            image_tag: latest
-          - python: 3.9
-            image_tag: 3.9
-          - python: 3.7
-            image_tag: 3.8
-          - python: 3.8
-            image_tag: 3.7
+          - python: "3.9"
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -20,7 +20,9 @@ jobs:
           - python: 3.9
             image_tag: nightly
           - python: 3.6
-            image_tag: latest
+            image_tag: 3.11
```
mdellweg, line 23: I know, this is a change in behavior, but I think, it is on the one hand more explicit, and on the other hand might give us a sign that we need to rebuild one of those container images for plugin bugfix releases at the right time.

cc @daviddavis

.github/workflows/main.yml
```diff
@@ -17,16 +17,18 @@ jobs:
...
-          - python: 3.7
-            image_tag: 3.8
-          - python: 3.8
-            image_tag: 3.7
+          - python: "3.9"
+            image_tag: "nightly"
+          - python: "3.6"
+            image_tag: "3.11"
+          - python: "3.6"
+            image_tag: "3.10"
```
mdellweg, line 25: I hate YAML!
Not in general, but this automated type guessing is driving me mad.

daviddavis (approved):

Issue #185, opened 2021-03-23
Add label command to ansible distributions


gerrod3:
Related to this pulp_ansible issue: https://pulp.plan.io/issues/8441

PR #186, opened 2021-03-23
Added label command to ansible distribution

gerrod3:
fixes: #185

### DIFF
--- a/pulpcore/cli/ansible/distribution.py
+++ b/pulpcore/cli/ansible/distribution.py
@@ -14,12 +14,14 @@ from pulpcore.cli.common.context import (
     pass_entity_context,
     pass_pulp_context,
 )
+from pulpcore.cli.common.generic import PluginRequiredVersion as PRV
 from pulpcore.cli.common.generic import (
     base_path_contains_option,
     base_path_option,
     create_command,
     destroy_command,
     href_option,
+    label_command,
     label_select_option,
     list_command,
     name_option,
@@ -79,6 +81,9 @@ distribution.add_command(list_command(decorators=filter_options))
 distribution.add_command(show_command(decorators=lookup_options))
 distribution.add_command(destroy_command(decorators=lookup_options))
 distribution.add_command(create_command(decorators=create_options))
+distribution.add_command(
+    label_command(need_plugins=[PRV("core", "3.10.0"), PRV("ansible", "0.8.0.dev")])
+)
 
 
 # TODO Add content_guard option
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -1,6 +1,6 @@
 import gettext
 import json
-from typing import Any, Optional, Tuple, Union
+from typing import Any, NamedTuple, Optional, Tuple, Union
 
 import click
 
@@ -18,6 +18,12 @@ from pulpcore.cli.common.context import (
 )
 
 _ = gettext.gettext
+
+
+class PluginRequiredVersion(NamedTuple):
+    name: str
+    min: Optional[str] = None
+    max: Optional[str] = None
 
 
 class PulpCommand(click.Command):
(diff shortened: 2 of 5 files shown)


pulpcore/cli/ansible/distribution.py
```diff
@@ -79,6 +80,7 @@ def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: s
 distribution.add_command(show_command(decorators=lookup_options))
 distribution.add_command(destroy_command(decorators=lookup_options))
 distribution.add_command(create_command(decorators=create_options))
+distribution.add_command(label_command())
```
mdellweg, line 83: This is added in an upcoming pulp_ansible version, right?
Can we pass a list of needed plugins with versions to the `label_command` generator?
And than we should have a test that can excercise this at least on the nightly container.

pulpcore/cli/common/generic.py
```diff
@@ -390,10 +390,15 @@ def label_command(**kwargs: Any) -> click.Command:
...
         kwargs["name"] = "label"
     decorators = kwargs.pop("decorators", [name_option, href_option])
+    needs = kwargs.pop("needs", [("core", "3.10.0")])
 
     @click.group(**kwargs)
-    def label_group() -> None:
-        pass
+    @pass_pulp_context
+    def label_group(pulp_ctx: PulpContext) -> None:
+        for plugin in needs:
```
mdellweg, line 398: And this loop variable i'd call `item` or `needed_plugin`.

pulpcore/cli/common/generic.py
```diff
@@ -390,10 +390,15 @@ def label_command(**kwargs: Any) -> click.Command:
...
+            min_ver = plugin[1] if len(plugin) >= 2 else None
+            max_ver = plugin[2] if len(plugin) == 3 else None
+            pulp_ctx.needs_plugin(plugin[0], min_ver, max_ver)
```
mdellweg, line 401: We can do something fancy with a namedtuple here. But maybe not today...

mdellweg (commented):
This exactly what i dreamed of. :rocket: 
I just want some variables to be renamed.

pulpcore/cli/common/generic.py
```diff
@@ -390,10 +390,15 @@ def label_command(**kwargs: Any) -> click.Command:
     if "name" not in kwargs:
         kwargs["name"] = "label"
     decorators = kwargs.pop("decorators", [name_option, href_option])
+    needs = kwargs.pop("needs", [("core", "3.10.0")])
```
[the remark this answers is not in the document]
gerrod3: For both the kwargs and the variable name?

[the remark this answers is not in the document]
mdellweg: Yes. Sure they should be the same.
Which one do you prefer?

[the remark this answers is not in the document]
gerrod3: `need_plugins`

pulpcore/cli/common/generic.py
gerrod3: Ask and you shall receive.

pulpcore/cli/common/generic.py
```diff
@@ -390,10 +396,13 @@ def label_command(**kwargs: Any) -> click.Command:
...
     decorators = kwargs.pop("decorators", [name_option, href_option])
+    need_plugins = kwargs.pop("need_plugins", [PluginRequiredVersion("core", "3.10.0")])
 
     @click.group(**kwargs)
-    def label_group() -> None:
-        pass
+    @pass_pulp_context
+    def label_group(pulp_ctx: PulpContext) -> None:
+        for item in need_plugins:
+            pulp_ctx.needs_plugin(item.name, item.min, item.max)
```
mdellweg, line 405: Would this work?
```suggestion
        for item in need_plugins:
            pulp_ctx.needs_plugin(*item)
```

or this?
```suggestion
        [pulp_ctx.needs_plugin(*item) for item in need_plugins]
```

gerrod3: 1. Yes. 2. Yes, but mypy does not like it since `needs_plugin` doesn't return anything.

mdellweg: yep, i agree the latter is rather unintuive. Usually you use list comprehensions that do _not_ have side effects. This would be only using the side effect.

mdellweg (approved):
Is ansible 0.8.0 containing that feature in the nightly container already?

gerrod3:
> Is ansible 0.8.0 containing that feature in the nightly container already?

I merged the ansible labels feature Wednesday, so it should I think.

PR #187, opened 2021-03-26
Added interactive-shell option to pulp-cli.

### LINKED ISSUE #181: Feature: add a "pulp shell" command to start an interactive command-interpreter

gerrod3:
Here it is @ggainey.  Still need to add ability to read the outputs and use external shell commands. 

closes #181

### DIFF
--- a/CHANGES/181.feature
+++ b/CHANGES/181.feature
@@ -1,0 +1,1 @@
+Added an interactive-shell mode to pulp-cli
--- a/pulpcore/cli/common/__init__.py
+++ b/pulpcore/cli/common/__init__.py
@@ -5,6 +5,7 @@ from typing import Any, Callable, Optional
 import click
 import pkg_resources
 import toml
+from click_shell import make_click_shell
 
 from pulpcore.cli.common.context import PulpContext
 from pulpcore.cli.common.debug import debug
@@ -147,6 +148,18 @@ def main(
 
 
 main.add_command(debug)
+
+
+@main.command("shell")
+@click.pass_context
+def pulp_shell(ctx: click.Context) -> None:
+    """Activate an interactive shell-mode"""
+    make_click_shell(
+        ctx.parent,
+        prompt="pulp> ",
+        intro="Starting Pulp3 interactive shell...",
+        hist_file=os.path.join(click.utils.get_app_dir("pulp"), "cli-history"),
+    ).cmdloop()
 
 
 ##############################################################################
--- a/docs/shell.md
+++ b/docs/shell.md
@@ -1,0 +1,33 @@
+Starting the CLI with "pulp shell" drops you into the shell:
+```
+(pulp) [vagrant@pulp3 ~]$ pulp shell
+Starting Pulp3 interactive shell...
+pulp> help
+
+Documented commands (type help <topic>):
+========================================
+access-policy  config     export    group      orphans     rpm     task
+ansible        container  exporter  importer   python      show    user
+artifact       debug      file      migration  repository  status  worker
+
+Undocumented commands:
+======================
+exit  help  quit
+
+pulp> status
+{
+    ...
+}
+
+pulp> exit
+(pulp) [vagrant@pulp3 ~]$
+```
+
+Issuing the command with arguments works as it does currently:
+```
+(pulp) [vagrant@pulp3 ~]$ pulp status
+{
+    ...
+}
+(pulp) [vagrant@pulp3 ~]$
+```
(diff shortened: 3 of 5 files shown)


pulpcore/cli/common/__init__.py
```diff
@@ -144,11 +144,18 @@ def _debug_callback(level: int, x: str) -> None:
         debug_callback=_debug_callback,
     )
     ctx.obj = PulpContext(api_kwargs=api_kwargs, format=format, background_tasks=background)
+    if ctx.invoked_subcommand == "shell":
+        make_click_shell(ctx, prompt="pulp-cli>", intro="Starting Pulp3 CLI...").cmdloop()
```
mdellweg, line 148: Can't this command be executed instead of the `pass` in line 156?
Maybe with `ctx.parent` to run in the correct context...

CHANGES/181.feature
```diff
@@ -0,0 +1,35 @@
...
+```
+
+Issuing the command with arguments works as it does currently:
+```
+(pulp) [vagrant@pulp3 ~]$ pulp status
+{
+    ...
+}
+(pulp) [vagrant@pulp3 ~]$
+```
```
daviddavis, line 35: This is probably too much detail for the changelog. I'd recommend creating a shell doc in the docs folder.

mdellweg: I agree. However, it's unclear to me which one is the pr we want to use here, and this file is originally from the other one.

ggainey: We def want to use @gerrod3 's PR, it answers the "run as pulp shell" request from my PR. I'm going to close mine and we'll continue on this one.

I think moving the example to its own doc page is a fine idea.

pulpcore/cli/common/__init__.py
gerrod3: You are correct!

CHANGES/181.feature
gerrod3: +1 Moved!

pulpcore/cli/common/__init__.py
```diff
@@ -149,6 +150,13 @@ def _debug_callback(level: int, x: str) -> None:
 main.add_command(debug)
 
 
+@main.command("shell")
+@click.pass_context
+def pulp_shell(ctx: click.Context) -> None:
+    """Activate an interactive shell-mode"""
+    make_click_shell(ctx.parent, prompt="pulp-cli>", intro="Starting Pulp3 CLI...").cmdloop()
```
mdellweg, line 157:
```suggestion
    make_click_shell(ctx.parent, prompt="pulp>", intro="Starting Pulp3 interactive shell...").cmdloop()
```
IMHO at this point it stops being a cli. ;)

daviddavis:
I tested this out. It works great. One small weird thing I see though is that the docs have:

```
pulp-cli> status
```

But I see a different prompt:

```
pulp>status
```

I don't have a huge pref between the two but it seems like they should be consistent and also it would be nice to have a space after the `>`.

daviddavis (approved):

ggainey:
Add change to rename history-file, and make sure it ends up in cfg-directory please!

mdellweg:
> Add change to rename history-file, and make sure it ends up in cfg-directory please!

I think it should be a cache-directory like `.cache/pulp/cli-history` maybe.

pulpcore/cli/common/__init__.py
```diff
@@ -149,6 +150,18 @@ def _debug_callback(level: int, x: str) -> None:
...
 
+@main.command("shell")
+@click.pass_context
+def pulp_shell(ctx: click.Context) -> None:
+    """Activate an interactive shell-mode"""
+    make_click_shell(
+        ctx.parent,
+        prompt="pulp> ",
+        intro="Starting Pulp3 interactive shell...",
+        hist_file=os.path.join(os.path.expanduser("~"), ".cache/pulp/cli-history"),
```
mdellweg, line 161: I had hoped, click would provide something like this for cache dirs too:
https://click.palletsprojects.com/en/7.x/api/#click.get_app_dir
But it seems not to be the case. So please disregard my former comment. Using that function should allow this project to stay platform independent.

pulpcore/cli/common/__init__.py
```diff
@@ -149,6 +150,18 @@ def _debug_callback(level: int, x: str) -> None:
...
 
+@main.command("shell")
+@click.pass_context
+def pulp_shell(ctx: click.Context) -> None:
+    """Activate an interactive shell-mode"""
+    make_click_shell(
+        ctx.parent,
+        prompt="pulp> ",
+        intro="Starting Pulp3 interactive shell...",
+        hist_file=os.path.join(click.utils.get_app_dir("pulp"), "cli-history"),
```
mdellweg, line 161: :+1:

mdellweg (approved):

Issue #101, opened 2021-01-28
Make export/import process simpler for CLI


gerrod3:
Users should be able to export the repositories they want in one command and shouldn't need to know the difference between 'export' and 'exporter'. I propose we remove the 'export' commands and re-brand them under 'exporter'. Exporter should get two new subcommands: 'export' and 'history'.

The 'history' command would get three subcommands from the original 'export' command: 'list', 'show', and 'delete'. Export objects are basically history records of previous exports and thus would fit nicely under this context. If a user doesn't specify a subcommand when using 'history' the default behaviour should be 'list' so they can see all the export objects aka the history.

The new 'export' sub command will take over the original export's 'run' command. This command will have extra functionality to allow for the creation of an exporter so the whole process can be done in one command.

Examples:

with Exporter named foo:

```
$pulp exporter pulp export --name foo [export options]
Started export task. Task href: ...
```

with no Exporter:
```
$pulp exporter pulp export --base-path tmp/exports/ --repository file_repo file --repository rpm_repo rpm [export options]
Would you like to save these configurations? [y/n]
$y
Please give a name to the exporter config options:
$bar
Config options saved in exporter object bar
Started export task. Task href: ...
```

If the user would specify 'no' on saving the config options, the command would give the created exporter a default name. The other commands under 'exporter' should have their help descriptions updated to reflect the exporter's role as a config object.

A similar process should be applied to the 'importer' commands so that the experience is consistent.

mdellweg:
I like the fact that the "exporter" can "export" and the result is an "export". This is like the "repository" can "sync" and the result is a "repository version".

gerrod3:
So is that a +1 or -1 for the design? Using your example the repository version commands are under the repository command and I'm proposing we do a similar thing and place the export commands under the exporter ones.  

> I like the fact that the "exporter" can "export" and the result is an "export". 

I don't like this. I think the wording is confusing when you use "export" to mean a verb and noun which is why I want to change the wording within the commands.

daviddavis:
FWIW, I've always used the noun when working on the API due to the RESTful architecture (ie users create/read/update/delete exports). I have no objection to using the verb in the CLI but I think hiding exports will be difficult (and perhaps even unwise).

mdellweg:
OK what about the "exporter" has a command "run" and the result is an "export"?
Whether we want to list and view and delete exports as a subcommand group of exporter is probably another question.

And yes, that was a +1...

dkliban:
[...]

PR #195, opened 2021-04-06
Added "import pulp show/list/run".

### LINKED ISSUE #93: As a user, I can import pulp exports

ggainey:
closes #93

### DIFF
--- a/pulpcore/cli/core/pulp_import.py
+++ b/pulpcore/cli/core/pulp_import.py
@@ -1,0 +1,83 @@
+import gettext
+from typing import Any, Dict
+
+import click
+
+from pulpcore.cli.common.context import (
+    DEFAULT_LIMIT,
+    PulpContext,
+    pass_entity_context,
+    pass_pulp_context,
+)
+from pulpcore.cli.common.generic import Mutex, destroy_command, href_option, show_command
+from pulpcore.cli.core.context import PulpImportContext, PulpImporterContext
+
+_ = gettext.gettext
+
+
+@click.group(name="import")
+def pulp_import() -> None:
+    pass
+
+
+@pulp_import.group()
+@pass_pulp_context
+@click.pass_context
+def pulp(ctx: click.Context, pulp_ctx: PulpContext) -> None:
+    ctx.obj = PulpImportContext(pulp_ctx)
+
+
+lookup_options = [href_option]
+
+pulp.add_command(show_command(decorators=lookup_options))
+pulp.add_command(destroy_command(decorators=lookup_options))
+
+
+@pulp.command()
+@click.option("--importer", type=str, required=True, help=_("Name of owning PulpImporter"))
+@click.option(
+    "--limit", default=DEFAULT_LIMIT, type=int, help=_("Limit the number of imports to show.")
+)
+@click.option("--offset", default=0, type=int, help=_("Skip a number of imports to show."))
+@pass_entity_context
+@pass_pulp_context
+def list(
+    pulp_ctx: PulpContext,
+    import_ctx: PulpImportContext,
+    importer: str,
+    limit: int,
+    offset: int,
+    **kwargs: Any
+) -> None:
+    params = {k: v for k, v in kwargs.items() if v is not None}
+    importer_ctx = PulpImporterContext(pulp_ctx)
+    import_ctx.importer = importer_ctx.find(name=importer)
+    result = import_ctx.list(limit=limit, offset=offset, parameters=params)
+    pulp_ctx.output_result(result)
+
+
+@pulp.command()
+@click.option("--importer", required=True)
+@click.option("--toc", type=str, cls=Mutex, not_required_if=["path"])
+@click.option("--path", type=str, cls=Mutex, not_required_if=["toc"])
+@pass_entity_context
+@pass_pulp_context
+def run(
+    pulp_ctx: PulpContext,
+    import_ctx: PulpImportContext,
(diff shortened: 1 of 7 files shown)


tests/scripts/pulpcore/test_pulpimporter.sh
```diff
@@ -32,3 +35,41 @@ expect_succ pulp importer pulp update --name "cli_test_importer" --repo-map src1
...
+# Test toc/path uniqueness
+expect_fail pulp import pulp run --importer $IMPORTER  --toc "$EXPORT_TOC" --path "$EXPORT_FILE"
+expect_fail pulp import pulp run --importer $IMPORTER
+
+expect_succ pulp import pulp destroy --href "$HREF"
+
+#
+# Clean up export files
+rm -f "$EXPORT_TOC"
+rm -f "$EXPORT_FILE"
```
daviddavis, line 75: I think you could put these in cleanup() and guard against the possibility of them being undefined with `-z ${EXPORT_TOC+x}` maybe?

ggainey: Fine idea - fixed

pulpcore/cli/core/pulp_import.py
```diff
@@ -0,0 +1,83 @@
...
+    importer: str,
+    limit: int,
+    offset: int,
+    **kwargs: Any
+) -> None:
+    params = {k: v for k, v in kwargs.items() if v is not None}
+    importer_ctx = PulpImporterContext(pulp_ctx)
+    import_ctx.importer = importer_ctx.find(name=importer)
+    result = import_ctx.list(limit=limit, offset=offset, parameters=params)
+    pulp_ctx.output_result(result)
```
gerrod3, line 56: This can be replaced with the list_command factory if you create a callback for the `importer` option and pass that option in as a decorator.  Check out the [repository callback](https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/common/generic.py#L81)

pulpcore/cli/core/pulp_import.py
```diff
@@ -0,0 +1,83 @@
...
+@click.option("--toc", type=str, cls=Mutex, not_required_if=["path"])
+@click.option("--path", type=str, cls=Mutex, not_required_if=["toc"])
+@pass_entity_context
+@pass_pulp_context
+def run(
+    pulp_ctx: PulpContext,
+    import_ctx: PulpImportContext,
+    importer: str,
+    path: str,
+    toc: str,
```
gerrod3, line 70: Same thing here, this can be replaced with the create_command factory if you create that callback for the importer object. Also if you pass in 'name' for kwargs you can have the command be named 'run' https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/common/generic.py#L274

gerrod3 (commented):
Looking at [export.py](https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/core/export.py) I think that file could also be optimized, but that should be a separate PR.  Is there a particular reason that the syntax for the export/import commands is `pulp export pulp _action_`? Why is the second "pulp" necessary, seems to be just an extra word not needed/used for the command.

daviddavis:
> Is there a particular reason that the syntax for the export/import commands is `pulp export pulp _action_`? Why is the second "pulp" necessary, seems to be just an extra word not needed/used for the command.

There are different types of exporters/exports/importers/imports. The pulp refers to the fact that it's a pulp export and not say a [filesystem export](https://github.com/pulp/pulpcore/blob/2d5abaa7987a653f0d88cb10010da14acaa225c0/pulpcore/app/models/exporter.py#L66).

We could do `pulp export -t pulp` and default type the type to "pulp" so that users can just call `pulp export`.

pulpcore/cli/core/pulp_import.py
ggainey: #101 asks to refactor the import/export commands. This PR is to add import-functionality that works the way export currently does. We can address all the comments here - but as part of #101.

Right now, I a) need access to import, b) quickly, and c) in a way that's consistent with export. All of the comments here, apply to the export-side code that this mirrors, we should think about a better approach for and then fix both sides at the same time.

ggainey: (my comment above was in response to https://github.com/pulp/pulp-cli/pull/195#pullrequestreview-630491374 , apologies.) Working on changing to use the factory as described here.

gerrod3: @ggainey I started some work on refactoring export.py and in it I figured out the callback needed for using the list factory, should be applicable for import as well: https://github.com/gerrod3/pulp-cli/commit/73dc0998e52941c8a1fb5363dc5c620713468597

ggainey:
So I have now spent literally three times longer trying to refactor this to take advantage of the "easier" New Way, than I did to implement this in the first place, without success. The factory methods and decorators look like a great idea - but without some kind of overview doc/tutorial, they're a little too magic for me, it appears.

I'm going to just close this PR; this functionality can wait until we refactor/replace export/import as a whole.

daviddavis:
I feel like this code is basically done. Would be nice to have these commands and tests even if they don't use the factories. What if we just merge this as-is?

mdellweg:
> I feel like this code is basically done. Would be nice to have these commands and tests even if they don't use the factories. What if we just merge this as-is?

Fine for me. Although, with the refactoring, the command structure will likely change.

PR #196, opened 2021-04-07
Add confirmation on invalid config

### LINKED ISSUE #156: pulp config refuses to work in case the config file is not valid
When the existing config file is not valid, not even editing the config file works:
```
$ pulp config edit              
Usage: pulp [OPTIONS] COMMAND [ARGS]...
Try 'pulp --help' for help.

Error: Invalid value for '--format': invalid choice: jsonn. (choose from json, yaml, none)
```

In this special case, `pulp --format=json config edit` helps.

In general though, if the config file fails to parse, it would be nice to get a warning and still be able to run the `pulp config ... `commands.

mdellweg:
fixes #156

### DIFF
--- a/pulpcore/cli/core/config.py
+++ b/pulpcore/cli/core/config.py
@@ -24,9 +24,6 @@ def config() -> None:
 @click.option("--editor", "-e", is_flag=True, help=_("Edit the config file in an editor"))
 @click.option("--overwrite", "-o", is_flag=True, help=_("Overwite any existing config file"))
 @click.option("--location", default=LOCATION, type=click.Path(resolve_path=True))
-@click.option(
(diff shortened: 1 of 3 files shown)


pulpcore/cli/core/config.py
```diff
@@ -68,7 +65,6 @@ def prompt_config_option(name: str) -> Any:
         for setting in SETTINGS:
             settings[setting] = locals()[setting] or ""
 
-    output = f"[cli]\n{toml.dumps(settings)}"
```
daviddavis, line 71: :+1:

daviddavis (approved):
Code LGTM. A test would be nice but not necessary.

mdellweg:
I'm not sure how to properly test interactive stuff.

PR #198, opened 2021-04-08
Guardian's AnonymousUser is no longer automatically created.

ipanova:
[noissue]

Required PR: https://github.com/pulp/pulpcore/pull/1239


mdellweg (approved):
The required PR should ne even be needed.
So this could be merged right away.

PR #199, opened 2021-04-08
Add support for autopublish and autodistribute

### LINKED ISSUE #155: Add support for autopublish and autodistribute
https://pulp.plan.io/issues/7469

daviddavis:
fixes #155

### DIFF
--- a/pulpcore/cli/file/distribution.py
+++ b/pulpcore/cli/file/distribution.py
@@ -1,8 +1,9 @@
 import gettext
+from typing import Optional, Union
 
 import click
 
-from pulpcore.cli.common.context import PulpContext, pass_pulp_context
+from pulpcore.cli.common.context import PulpContext, PulpEntityContext, pass_pulp_context
 from pulpcore.cli.common.generic import (
     base_path_contains_option,
     base_path_option,
@@ -16,9 +17,19 @@ from pulpcore.cli.common.generic import (
     show_command,
     update_command,
 )
-from pulpcore.cli.file.context import PulpFileDistributionContext
+from pulpcore.cli.file.context import PulpFileDistributionContext, PulpFileRepositoryContext
 
 _ = gettext.gettext
+
+
+def _repository_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[str]
+) -> Union[str, PulpEntityContext, None]:
+    # Pass None and "" verbatim
+    if value:
+        pulp_ctx: PulpContext = ctx.find_object(PulpContext)
+        return PulpFileRepositoryContext(pulp_ctx, entity={"name": value})
+    return value
 
 
 @click.group()
@@ -40,14 +51,13 @@ def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: s
 
 filter_options = [label_select_option, base_path_option, base_path_contains_option]
 lookup_options = [href_option, name_option]
-create_options = [
-    click.option("--name", required=True),
-    click.option("--base-path", required=True),
-    click.option("--publication"),
-]
 update_options = [
     click.option("--base-path"),
     click.option("--publication"),
+    click.option("--repository", callback=_repository_callback),
+]
+create_options = update_options + [
+    click.option("--name", required=True),
 ]
 
 distribution.add_command(list_command(decorators=filter_options))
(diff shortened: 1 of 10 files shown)


daviddavis:
If this looks good, I'll add in support for pulp_rpm.

pulpcore/cli/file/distribution.py
```diff
@@ -40,14 +51,13 @@ def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: s
...
 lookup_options = [href_option, name_option]
-create_options = [
-    click.option("--name", required=True),
-    click.option("--base-path", required=True),
-    click.option("--publication"),
-]
 update_options = [
     click.option("--base-path"),
     click.option("--publication"),
+    click.option("--repository", callback=_repository_callback),
```
mdellweg, line 57: :+1:

mdellweg:
This all looks good to me. And i believe the nightly is failing, because those changes will land there tomorrow!?

pulpcore/cli/rpm/repository.py
```diff
@@ -55,17 +56,22 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
...
+    click.option(
+        "--metadata-checksum-type", type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False)
+    ),
+    click.option(
+        "--package-checksum-type", type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False)
+    ),
+    click.option("--gpgcheck", type=click.Choice(("0", "1"))),
+    click.option("--repo-gpgcheck", type=click.Choice(("0", "1"))),
+    click.option("--sqlite-metadata/--no-sqlite-metadata", default=None),
+    click.option("--autopublish/--no-autopublish", default=None),
```
mdellweg, line 72: This kind of completes the repository options, not necessarily related to auto publish.
Do we want to reflect that in the changelog?

mdellweg (approved):
One uncertainty wrt the changelog.

pulpcore/cli/file/repository.py
```diff
@@ -64,14 +64,14 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
...
-create_options = [
-    click.option("--name", required=True),
-    click.option("--description"),
-    click.option("--remote", callback=_remote_callback),
-]
 update_options = [
     click.option("--description"),
     click.option("--remote", callback=_remote_callback),
+    click.option("--manifest"),
+    click.option("--autopublish/--no-autopublish", default=None),
```
dralley, line 71: IMO the default should simply be --no-autopublish.

pulpcore/cli/rpm/repository.py
```diff
@@ -55,17 +56,22 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
...
     click.option("--remote", callback=_remote_callback),
+    click.option(
+        "--metadata-checksum-type", type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False)
+    ),
+    click.option(
+        "--package-checksum-type", type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False)
+    ),
+    click.option("--gpgcheck", type=click.Choice(("0", "1"))),
+    click.option("--repo-gpgcheck", type=click.Choice(("0", "1"))),
+    click.option("--sqlite-metadata/--no-sqlite-metadata", default=None),
```
dralley, line 71: Same here

pulpcore/cli/file/repository.py
mdellweg: This is an update_option. So if not specified, we should def _not_ set (and then update) this option to `False`.
I think having `None` here, keeps it from being touched on update calls, and on create calls i'd rather let the server decide on the default.

dralley: Ah, I believe I misunderstood what it does then

daviddavis: I agree it should default to `--no-autopublish`. The dilemma is that this will cause `autopublish` to always be set in the requests which won't work for older versions of pulp_file/pulp_rpm and we support 5 versions back.

I could either:

1. Check the version of pulp_file and if it's less than 1.7.0, silently strip out the autopublish key/value from the request.
2. Keep the default as None which won't set autopublish in requests unless the user explicitly sets `--autopublish` or `--no-autopublish`. I could add a comment for us to change the default this once we drop support for 1.7.0.

Thoughts?

daviddavis: Welp, ignore my comment. I am too slow.

PR #200, opened 2021-04-09
Add 3.12 to the test matrix

mdellweg:
[noissue]

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,8 @@ jobs:
         include:
           - python: "3.9"
             image_tag: "nightly"
+          - python: "3.8"
+            image_tag: "3.12"
           - python: "3.6"
             image_tag: "3.11"
           - python: "3.6"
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -27,8 +29,6 @@ jobs:
             image_tag: "3.9"
           - python: "3.7"
             image_tag: "3.8"
-          - python: "3.8"
-            image_tag: "3.7"
```
mdellweg, line 31: I think, we decided to keep compatibility with 5 versions of pulpcore.
OTOH 3.7 seems to get bugfix update still. Should we keep it around?

mdellweg:
Also this is depending on the success of:

https://github.com/pulp/pulp-oci-images/pull/70

.github/workflows/main.yml
daviddavis: Good question. I believe Katello and galaxy are both on 3.7 so I suppose we should.

daviddavis (approved):

PR #203, opened 2021-04-12
Fix mypy invocation for namespace packages

mdellweg:
https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules


daviddavis:
I think unpinned is fine since we don't have a way to update dependencies (eg dependabot). However, it looks like the options won't work in mypy <0.800 so maybe worth setting `mypy>=0.800`?

daviddavis (approved):

PR #204, opened 2021-04-12
Update mypy

daviddavis:
[noissue]


mdellweg:
This time, i was a few minutes faster. ;)
#203 
The question is, do we still want to pin the version, or do we believe that we fixed _that_ problem for good?

PR #205, opened 2021-04-14
Add some handy pulp rpm commands

vonsch:
* ```pulp rpm content-package list``` for listing rpm content (packages)
 * ```pulp rpm content-package upload``` for uploading rpm content (packages)
 * ```pulp rpm repository modify``` for adding/removing rpm content from repositories

### DIFF
--- a/pulpcore/cli/rpm/context.py
+++ b/pulpcore/cli/rpm/context.py
@@ -10,6 +10,15 @@ from pulpcore.cli.common.context import (
 _ = gettext.gettext
 
 
+class PulpRpmPackageContentContext(PulpEntityContext):
+    ENTITY = "rpm_package content"
+    ENTITIES = "rpm_package content"
+    HREF = "rpm_package_href"
+    LIST_ID = "content_rpm_packages_list"
+    READ_ID = "content_rpm_packages_read"
+    CREATE_ID = "content_rpm_packages_create"
+
+
 class PulpRpmDistributionContext(PulpEntityContext):
     ENTITY = "distribution"
     HREF = "rpm_rpm_distribution_href"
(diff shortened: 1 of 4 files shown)


vonsch:
Please note I'm quite poor python developer, so you might consider this only as an example and baseline for your PR

pulpcore/cli/rpm/context.py
```diff
@@ -10,6 +10,15 @@
 _ = gettext.gettext
 
 
+class PulpRpmPackageContentContext(PulpEntityContext):
```
mdellweg, line 13: I think, It's safe to call this `PulpRpmPackageContext`.

pulpcore/cli/rpm/repository.py
```diff
@@ -110,3 +111,41 @@ def sync(
         href=repository_href,
         body=body,
     )
+
+
+@repository.command()
+@name_option
+@href_option
+@click.option("--add-href", type=str, help="Comma-separated list of rpm content HREFs")
+@click.option("--remove-href", type=str, help="Comma-separated list of rpm content HREFs")
```
mdellweg, line 120: We use json input to specify the content in the file case. And i'd really like to keep the interface consistent, so the user does not need to learn a new way with every plugin.

mdellweg (commented):
Thank you for this contribution. However i think there is overlapping work being done in #194 .
Maybe you should coordinate with @gerrod3

vonsch:
> Thank you for this contribution. However i think there is overlapping work being done in #194 .
> Maybe you should coordinate with @gerrod3

Thank you for pointing to this, I checked #194 and it seems that half of my PR is not necessary at all. Since #194 seems to be more proper approach than mine (and it includes tests etc), I will wait till it gets merged and then I will rebase and improve my PR.

gerrod3:
@vonsch The new content factory subgroup feature has been merged.  Check out [file plugin's](https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/file/repository.py#L93-L140 ) implementation of their content commands using this new factory.

mdellweg:
@vonsch may i kindly ask if you are still working on this?

vonsch:
@mdellweg I didn't have enough time to continue on this task, so let's close the PR. I will create new PR when I get again onto this (or some of my colleagues)

PR #206, opened 2021-04-15
Add debug wait-for-pulp command

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/cli/common/debug.py
+++ b/pulpcore/cli/common/debug.py
@@ -17,10 +17,13 @@ def debug() -> None:
 
 @debug.command()
 @click.option("--name", required=True)
-@click.option("--min-version")
-@click.option("--max-version")
+@click.option("--min-version", help=_("Succeed only if the installed version is not smaller."))
+@click.option("--max-version", help=_("Succeed only if the installed version is smaller."))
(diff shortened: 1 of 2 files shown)


mdellweg:
Yet another "wait-for-pulp" implementation.
But maybe in this context it deserves a different name. Please add suggestions.

pulpcore/cli/common/debug.py
```diff
@@ -17,10 +18,34 @@ def debug() -> None:
...
     sys.exit(0 if available else 1)
+
+
+@debug.command()
+@click.option("--seconds", type=int, default=10, help=_("Number of seconds to wait at most."))
+@pass_pulp_context
+def wait_for_pulp(pulp_ctx: PulpContext, seconds: int) -> None:
+    """
+    Repeatedly check for the status of the server to be available.
+    """
```
mdellweg, line 39: Maybe it's worth to include this functionality in the regular `status` command.

daviddavis: I could see that as long as it's not the default behavior.

daviddavis (approved):

daviddavis (approved):

PR #210, opened 2021-04-23
Add retained_versions option to repository commands.

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/cli/ansible/distribution.py
+++ b/pulpcore/cli/ansible/distribution.py
@@ -10,11 +10,11 @@ from pulpcore.cli.ansible.context import (
 from pulpcore.cli.common.context import (
     EntityDefinition,
     EntityFieldDefinition,
+    PluginRequirement,
     PulpContext,
     pass_entity_context,
     pass_pulp_context,
 )
-from pulpcore.cli.common.generic import PluginRequiredVersion as PRV
 from pulpcore.cli.common.generic import (
     base_path_contains_option,
     base_path_option,
@@ -82,7 +82,12 @@ distribution.add_command(show_command(decorators=lookup_options))
 distribution.add_command(destroy_command(decorators=lookup_options))
 distribution.add_command(create_command(decorators=create_options))
 distribution.add_command(
-    label_command(need_plugins=[PRV("core", "3.10.0"), PRV("ansible", "0.8.0.dev")])
+    label_command(
+        need_plugins=[
+            PluginRequirement("core", "3.10.0"),
+            PluginRequirement("ansible", "0.8.0.dev"),
+        ]
+    )
 )
 
 
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -3,7 +3,7 @@ import gettext
 import json
 import sys
 import time
-from typing import IO, Any, ClassVar, Dict, List, Optional, Set, Tuple, Type, Union
+from typing import IO, Any, ClassVar, Dict, List, NamedTuple, Optional, Set, Tuple, Type, Union
 
 import click
 import yaml
@@ -32,6 +32,14 @@ BATCH_SIZE = 25
 EntityDefinition = Dict[str, Any]
 RepositoryDefinition = Tuple[str, str]  # name, pulp_type
 RepositoryVersionDefinition = Tuple[str, str, int]  # name, pulp_type, version
+
+
+class PluginRequirement(NamedTuple):
+    name: str
+    min: Optional[str] = None
+    max: Optional[str] = None
+    feature: Optional[str] = _("this command")
+
 
 new_component_names_to_pre_3_11_names: Dict[str, str] = dict(
     ansible="pulp_ansible",
@@ -74,7 +82,7 @@ class PulpContext:
     def __init__(self, api_kwargs: Dict[str, Any], format: str, background_tasks: bool) -> None:
         self._api: Optional[OpenAPI] = None
         self._api_kwargs = api_kwargs
-        self._needed_plugins: List[Tuple[str, Optional[str], Optional[str]]] = []
+        self._needed_plugins: List[PluginRequirement] = []
         self.isatty: bool = sys.stdout.isatty()
 
         self.format: str = format
@@ -90,8 +98,8 @@ class PulpContext:
             except OpenAPIError as e:
                 raise click.ClickException(str(e))
             # Rerun scheduled version checks
-            for name, min_version, max_version in self._needed_plugins:
-                self.needs_plugin(name, min_version, max_version)
+            for plugin in self._needed_plugins:
+                self.needs_plugin(plugin.name, plugin.min, plugin.max, plugin.feature)
         return self._api
 
     @property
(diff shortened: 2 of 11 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -132,36 +160,36 @@ def load_json_callback(
 ##############################################################################
 # Decorator common options
 
+def pulp_option(*args: Any, **kwargs: Any) -> Callable[..., Any]:
```
daviddavis, line 163: I tried for 2 hours to make mypy happy here but no luck. The type from click is a bit complex:

https://github.com/pallets/click/blob/ffeddbaf0c8226201136593d29324c951f2b2fa3/src/click/decorators.py#L229

mdellweg: I think TypeVar can help to describea callable that returnes the same type as it received.

pulpcore/cli/common/generic.py
```diff
@@ -40,6 +40,34 @@ def format_help_text(
...
+    def __init__(
+        self,
+        *args: Any,
+        needs_plugin: Optional[str] = None,
+        min_version: Optional[str] = None,
+        max_version: Optional[str] = None,
+        **kwargs: Any,
+    ):
+        if (min_version or max_version) and not needs_plugin:
+            raise click.ClickException(
```
mdellweg, line 52: I'd reserve ClickExceptions to user errors. This one is an implementation error. Should be ok to use assert here.

pulpcore/cli/common/generic.py
```diff
@@ -40,6 +40,34 @@ def format_help_text(
 
 
 class PulpOption(click.Option):
+    def __init__(
+        self,
+        *args: Any,
+        needs_plugin: Optional[str] = None,
```
mdellweg, line 46: Should we allow a single version to require multiple plugins?
May be overkill...

pulpcore/cli/common/generic.py
```diff
@@ -40,6 +40,34 @@ def format_help_text(
...
+
+    def process_value(self, ctx: click.Context, value: Any) -> Any:
+        if value is not None and self.needs_plugin:
+            pulp_ctx = ctx.find_object(PulpContext)
+            pulp_ctx.needs_plugin(
+                self.needs_plugin,
+                self.min_version,
+                self.max_version,
+                _("the {name} option").format(name=self.name),
+            )
```
mdellweg, line 68: :1st_place_medal:

pulpcore/cli/common/context.py
```diff
@@ -43,6 +44,8 @@
     rpm="pulp_rpm",
 )
 
+NeededPlugin = namedtuple("NeededPlugin", ["name", "min_version", "max_version", "feature"])
```
mdellweg, line 47: There is a new way to define named tuples (with typing hints):

https://docs.python.org/3/library/typing.html#typing.NamedTuple

mdellweg: Also we have it as `pulpcore.cli.common.generic.PluginRequiredVersion` already.

pulpcore/cli/common/generic.py
mdellweg:
```
_F = TypeVar("_F")
def pulp_option(*args: Any, **kwargs: Any) -> Callable[[_F], _F]:
```
works for me.

daviddavis: Huh, could have sworn I tried that at one point. It works. Thank you.

pulpcore/cli/common/generic.py
daviddavis: I considered this and thought it was unlikely that an option would require more than one plugin. It's possible that an option would require a specific plugin version and a core version but then I'd expect the plugin to pin that version of core.

pulpcore/cli/common/context.py
```diff
@@ -33,6 +33,14 @@
 RepositoryDefinition = Tuple[str, str]  # name, pulp_type
 RepositoryVersionDefinition = Tuple[str, str, int]  # name, pulp_type, version
 
+
+class PluginRequiredVersion(NamedTuple):
```
daviddavis, line 37: Had to move this here to avoid circular dependency.

pulpcore/cli/common/generic.py
mdellweg: I accept that argument. We can still consider to attach this variable directly to the `PluginRequiredVersion` tuple.
It would even spare us some validation.

daviddavis:
This is going to fail until https://github.com/pulp/pulpcore/pull/1204 is merged.

pulpcore/cli/common/generic.py
daviddavis: So devs would supply a `PluginRequiredVersion` when setting needs_plugin? The only validation that this would save us from is having to check that needs_plugin is set if min/max version is supplied. The trade off is that we have to import and instantiate `PluginRequiredVersion` every time we want to set the option requirement. I'm not opposed to it but it doesn't seem like a decisive advantage.

daviddavis: Here's a PoC for having a needs_plugins parameter that takes `PluginRequiredVersion` (which I've renamed to `PluginRequirement` due to fewer characters). Let me know what you think. If you think it's better, I can merge it into this PR.

https://github.com/daviddavis/pulp-cli/commit/f56ba15f9f88105349bdadfe1fafa5616bd65fb1

pulpcore/cli/ansible/distribution.py
```diff
@@ -7,14 +7,9 @@
...
-    EntityDefinition,
-    EntityFieldDefinition,
-    PulpContext,
-    pass_entity_context,
-    pass_pulp_context,
-)
-from pulpcore.cli.common.generic import PluginRequiredVersion as PRV
+from pulpcore.cli.common.context import EntityDefinition, EntityFieldDefinition
+from pulpcore.cli.common.context import PluginRequiredVersion as PRV
+from pulpcore.cli.common.context import PulpContext, pass_entity_context, pass_pulp_context
```
mdellweg, line 12: What was bad about the old version of bunched import?

pulpcore/cli/common/context.py
```diff
@@ -74,7 +82,7 @@ class PulpContext:
     def __init__(self, api_kwargs: Dict[str, Any], format: str, background_tasks: bool) -> None:
         self._api: Optional[OpenAPI] = None
         self._api_kwargs = api_kwargs
-        self._needed_plugins: List[Tuple[str, Optional[str], Optional[str]]] = []
+        self._needed_plugins: List[PluginRequiredVersion] = []
```
mdellweg, line 85: :+1:

pulpcore/cli/ansible/distribution.py
daviddavis: Isort automatically did this. It breaks alias imports (`PluginRequiredVersion as PRV`) into a separate line. Since PluginRequiredVersion moved from `pulpcore.cli.common.generic` to `pulpcore.cli.common.context`, it broke up the `pulpcore.cli.common.context` into multiple lines you see here to keep the alphabetic order.

pulpcore/cli/common/generic.py
mdellweg: I like the name `PluginRequirement`.
Need to think about the whole change. I thought, it would also change the signature of `needs_plugin` and `has_plugin`, but that might be a bad idea at this point in time.
Maybe we should think about this separately from this pr.

daviddavis: Ok, sounds good. I'll rename PluginRequiredVersion to PluginRequirement here and then open a new PR for us to consider.

pulpcore/cli/ansible/distribution.py
mdellweg: I see. Thanks.

PR #212, opened 2021-04-26
Support for mirror flag when syncing rpm repositories

vonsch:
[noissue]

### DIFF
--- a/CHANGES/212.feature
+++ b/CHANGES/212.feature
@@ -1,0 +1,1 @@
+Added mirror flag support for pulp rpm repository sync.
--- a/pulpcore/cli/rpm/repository.py
+++ b/pulpcore/cli/rpm/repository.py
@@ -1,5 +1,5 @@
 import gettext
-from typing import Optional, Union
+from typing import Any, Dict, Optional, Union
 
 import click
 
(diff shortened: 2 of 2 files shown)


vonsch:
Btw I also considered if it would be better to use ```--mirror/--no-mirror``` python click options instead of ```--mirror True```/```--mirror False```, but not sure what's better.

mdellweg:
We use `--bool/--no-bool` everywhere else.

vonsch:
@mdellweg thanks for suggestion, I modified PR

mdellweg (approved):
@daviddavis do you think, we need a changelog here?

daviddavis:
I don't see the harm in adding a changelog. It's good to always call out new additions.

That said, I have no objection to not having a changelog either.

vonsch:
@mdellweg If prefer to have changelog entry, I can easily add one. However if I understand the ```CHANGES/``` directory layout and file naming well, I must first create some github issue and then reference it in the ```CHANGES/<issue_number>.feature, right?

daviddavis:
You could just use this PR number (ie `CHANGES/212.feature`).

vonsch:
@daviddavis ok, added

mdellweg (approved):
Thank you very much!

PR #216, opened 2021-04-28
Refactor option version requirement

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -38,7 +38,7 @@ class PluginRequirement(NamedTuple):
     name: str
     min: Optional[str] = None
     max: Optional[str] = None
-    feature: Optional[str] = _("this command")
+    feature: Optional[str] = None
 
 
 new_component_names_to_pre_3_11_names: Dict[str, str] = dict(
@@ -206,7 +206,7 @@ class PulpContext:
         name: str,
         min_version: Optional[str] = None,
         max_version: Optional[str] = None,
-        feature: Optional[str] = _("this command"),
+        feature: Optional[str] = None,
     ) -> None:
         if self._api is not None:
             if not self.has_plugin(name, min_version, max_version):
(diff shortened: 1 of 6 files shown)


gerrod3:
@daviddavis Is this PR ready to go? I would like to use this option version requirement feature soon.

daviddavis:
@gerrod3 you should be able to use it today. See https://github.com/pulp/pulp-cli/blob/master/pulpcore/cli/file/repository.py#L85 as an example. This just changes the option to use the `PluginRequirement` class.

gerrod3:
> @gerrod3 you should be able to use it today. See https://github.com/pulp/pulp-cli/blob/master/pulpcore/cli/file/repository.py#L85 as an example. This just changes the option to use the `PluginRequirement` class.

It also changes the `needs_plugin` option to a list. I think we should merge this refactor before I build anymore features on top of it.

daviddavis:
I'm happy to update any new code once this gets merged. But I will also set this to ready to review.

pulpcore/cli/common/context.py
```diff
@@ -217,6 +217,8 @@ def needs_plugin(
                     separator = ","
                 if max_version is not None:
                     specifier += f"{separator}<{max_version}"
+                if feature is None:
+                    feature = _("this command")
```
mdellweg, line 221: :+1:

mdellweg (approved):
Looks like it needs a rebase.

pulpcore/cli/common/generic.py
```diff
@@ -39,27 +39,27 @@ class PulpOption(click.Option):
...
-                self.needs_plugin,
-                self.min_version,
-                self.max_version,
-                _("the {name} option").format(name=self.name),
-            )
+            for plugin_requirement in self.needs_plugins:
+                if plugin_requirement.feature:
+                    feature = plugin_requirement.feature
+                else:
+                    feature = _("the {name} option").format(name=self.name)
```
gerrod3, line 55: Should the `.format` be on the inside string? I don't know if it matters, but I'm pretty sure it works either way.

daviddavis: No, we always want to call `format()` outside of `_()` otherwise the string will have the name substituted for `{name}` and thus won't match the translated string.

daviddavis: This is also one reason why we can't use f-strings with gettext.

PR #218, opened 2021-05-01
Update Python commands with new options

gerrod3:
[noissue]


mdellweg (approved):
A changelog for the added options is optional.

PR #219, opened 2021-05-01
Use PluginRequirement for has/needs_plugin

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -3,7 +3,20 @@ import gettext
 import json
 import sys
 import time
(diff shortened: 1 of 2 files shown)


daviddavis:
This code looks good to me. I'm interested though in why this keeps support for plugin/min_version/max_version instead of just standardizing on using PluginRequirement?

mdellweg:
Because there is at least one plugin out of this tree using it.

daviddavis:
Thank you, makes sense. Is the plan to keep plugin/min_version/max_version around permanently or temporarily?

daviddavis (approved):
This lgtm. 👍

mdellweg:
> Thank you, makes sense. Is the plan to keep plugin/min_version/max_version around permanently or temporarily?

I hope temporarily.

PR #229, opened 2021-05-06
Fix requirements file option for ansible collection remote create.

### LINKED ISSUE #226: As a user I want to be able to pass a path to a YAML file with a list of Ansible Collections I wish to be represented by a Ansible Remote in the Ansible plugin
When creating or updating a remote Collection for the Ansible plugin the REST API documents a parameter, "requirements_file." This parameter is currently exposed via the pulp CLI tool. However, it only accepts a string directly. When you instead pass a file path to a YAML file it fails to serialize the YAML file correctly to JSON and throws an exception. My YAML file looked similar to the following:
---
collections:
  - testing.ansible_testing_content
  - pulp.squeezer

daviddavis:
fixes #226


daviddavis:
We should probably follow the pattern we use elsewhere and have `--requirements` accept a string or a file with `@` instead of having `--requirements` and `--requirements-file`.

mdellweg:
> We should probably follow the pattern we use elsewhere and have `--requirements` accept a string or a file with `@` instead of having `--requirements` and `--requirements-file`.
> 
> I can file a feature issue if others agree.

I just thought it was that way...

mdellweg (approved):

PR #236, opened 2021-05-12
Make dry_run a setting and --force its inverse

mdellweg:
fixes TBA

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -87,3 +87,6 @@ key = "/etc/pki/katello/private/pulp-client.key"
 verify_ssl = false
 ```
 
+As Katello uses Pulp as a backend all modifying actions in Pulp should be performed via Katello.
+Therefore you are also strongly encourged to set `dry_run = false`, to prevent accidentally calling into dangerous commands.
+This setting can in turn be overwritten on the command-line with the `--force` flag.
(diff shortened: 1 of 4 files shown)


daviddavis (approved):
LGTM. Maybe update https://github.com/pulp/pulp-cli/blob/develop/docs/install.md#katello ?

docs/install.md
```diff
@@ -87,3 +87,6 @@ key = "/etc/pki/katello/private/pulp-client.key"
 verify_ssl = false
 ```
 
+As Katello uses Pulp as a backend all modifying actions in Pulp should be performed via Katello.
+Therefore you are also strongly encourged to set `dry_run = false`, to prevent accidentally calling into dangerous commands.
+This setting can in turn be overwritten on the command-line with the `--force` flag.
```
mdellweg, line 92: @daviddavis 
Is this, what you had in mind?

daviddavis (approved):

docs/install.md
```diff
@@ -87,3 +87,6 @@ key = "/etc/pki/katello/private/pulp-client.key"
 verify_ssl = false
 ```
 
+As Katello uses Pulp as a backend all modifying actions in Pulp should be performed via Katello.
+Therefore you are also strongly encourged to set `dry_run = false`, to prevent accidentally calling into dangerous commands.
```
evgeni, line 91: Should that have been `dry_run = true`? ;-)

mdellweg: absolutely!

PR #237, opened 2021-05-13
Pin dependencies and enable dependabot

### LINKED ISSUE #235: Pin dependencies and use dependabot to update them
After click 8 broke the CLI, we decided to pin all dependencies to major versions at least.

daviddavis:
fixes #235

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,11 +22,11 @@ setup(
     python_requires=">=3.6",
     install_requires=[
         "click<8.0.0",
-        "click-shell",
+        "click-shell~=2.0",
         "packaging",
-        "PyYAML",
-        "requests",
-        "toml",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -22,11 +22,11 @@
     python_requires=">=3.6",
     install_requires=[
         "click<8.0.0",
-        "click-shell",
+        "click-shell~=2.0",
```
daviddavis, line 25: It appears they don't have patch releases.

setup.py
```diff
@@ -22,11 +22,11 @@
...
         "packaging",
```
daviddavis, line 26: It seems like we are always are bumping the major and our use of packaging is very limited so decided not to bother pinning.

mdellweg (approved):

Issue #240, opened 2021-05-14
Allow users to specify --includes/excludes for Python remotes with requirements.txt files


gerrod3:
I think the CLI is the best place to put this feature since it can read the files on the system.

Related pulp_python issue: https://github.com/pulp/pulp_python/issues/345

PR #241, opened 2021-05-14
Python remote packages can be specified with requirements.txt files

gerrod3:
fixes: #240

### DIFF
--- a/CHANGES/240.feature
+++ b/CHANGES/240.feature
@@ -1,0 +1,1 @@
+Python remote fields --includes/--excludes can now be specified with requirements.txt files
--- a/pulpcore/cli/python/remote.py
+++ b/pulpcore/cli/python/remote.py
@@ -1,4 +1,6 @@
 import gettext
+import json
+from typing import Any, List, Optional
 
 import click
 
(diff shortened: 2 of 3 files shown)


mdellweg:
I'm not in the position to decide, whether it would be a good idea, to let the API itself to understand such requirements files, but have you considered using an existing parsing function like here: https://stackoverflow.com/a/59971236 ?

gerrod3:
> I'm not in the position to decide, whether it would be a good idea, to let the API itself to understand such requirements files, but have you considered using an existing parsing function like here: https://stackoverflow.com/a/59971236 ?

I thought about it, but using the parsing function would prevent using requirements files that are conglomerates of other requirements files.  The only parsing that is being done here is removing spaces and comments, the rest of the validation is still done by the API.

dralley:
@gerrod3  This looks good in terms of implementation, my only concern is whether we want to support not just requirements files but also .lock files for package managers like Poetry and Pipenv.  And maybe a user a user would want to add more than one requirements.txt.

dralley (approved):

mdellweg (approved):
@dralley thanks for looking!
Making your review official.

PR #242, opened 2021-05-14
Fix the default value for bool settings

daviddavis:
[noissue]


mdellweg (approved):
That should work.
Let's unblock people...
I have a different solution upcoming in #239

mdellweg:
BTW: Thank you for debugging this!

PR #246, opened 2021-05-19
Make click-shell an optional dependency

mdellweg:
There is no way i found, to make the cli depend on click <9 normally, but <8 if the shell extension is installed.

Maybe we can do smething like this only after https://github.com/clarkperkins/click-shell/pull/29 has been resolved.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -8,8 +8,9 @@ section on how to configure pulp-cli.
 ### From PyPI
 
 ```
+pip install pulp-cli  # minimal dependencies
 pip install pulp-cli[pygments]  # colorized output
-pip install pulp-cli  # no color output
+pip install pulp-cli[shell]  # with interactive shell mode
 ```
 
 ### From a source checkout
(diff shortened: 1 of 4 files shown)


daviddavis (approved):

mdellweg:
And maybe we are also blocked on this: https://github.com/pallets/click/issues/1885

mdellweg:
But actually this does not block us if we keep the <8.0.0 for now.

mdellweg:
The question remains, whether we want to deliberately break our optional dependency on click-shell by not pinning click < 8.

daviddavis:
Oh, I see. I don't have a strong opinion. Let's break it and see who complains?

daviddavis:
Also, I'm a bit worried that click-shell hasn't responded to your PR yet. It makes me worried that the repo is dead.

mdellweg:
It turns out, that click 8.0.1 is also changing some typing concepts, that i fail to adjust without breaking the mypy linting on click~=7.1.2, and the bash completion stopped working out of the box. So i'd suggest we merge this as is and look at click 8 support in a separate effort.

PR #247, opened 2021-05-19
Adding mkdocs

daviddavis:
Here's an option to build and publish our docs. Feedback welcome. If this looks good, I can work on a github action workflow to publish our docs to docs.pulpproject.org. 

Also, one thing I need to figure out is how to get the links working.


fao89:
we use mkdocs on pulp-installer and pulp-operator, which links are not working?

daviddavis:
I'm currently symlinking the README.md at `docs/index.md` and all the relative links in README.md to places like `docs/install.md` don't work.

I think I should maybe break up README.md into files like `contributing.md` and create an actual `docs/index.md` instead of symlinking the README at `docs/index.md`.

fao89:
> I'm currently symlinking the README.md at `docs/index.md` and all the relative links in README.md to places like `docs/install.md` don't work.
> 
> I think I should maybe break up README.md into files like `contributing.md` and create an actual `docs/index.md` instead of symlinking the README at `docs/index.md`.

we did something similar on the installer on the installer, all roles are symlinked, here is an example of pulp_all_services Readme.md referring to pulp_services:
```
- [pulp_services](../pulp_services/): A role to install & configure Pulp 3's
```

daviddavis:
@pulp/cli I fixed the links. Please test this out and let me know what you think.

To look at the docs, check out this PR and run `make servedocs`.

README.md
```diff
@@ -4,117 +4,6 @@ This is a command line interface for Pulp 3.
...
-```
-
-Read [the installation and configuration doc](docs/install.md) for more information.
-
-## Known issues
-
-  * Redirecting from `http` to `https`, as done by a typical Pulp installation,
-    does not work properly with `POST` and `PUT` requests.
-    Please use `https://` in the base url.
-    Note that the attempt to use `http` leaks sensitive data over an unencrypted connection.
```
daviddavis, line 36: FYI, I dropped this. I don't think it's an issue anymore since we don't follow redirects.

mdellweg: So what happens now, if you try to use the wrong schema?

daviddavis: You get a connection error.

daviddavis: ^ That's if there's no redirect.

If there's a redirect, then users receive [this error](https://github.com/pulp/pulp-cli/blob/42f62321e61465b28c9dcbc0c9822538631dcb16/pulpcore/cli/common/openapi.py#L290-L294)

daviddavis:
@pulp/cli it looks like both pulp_installer and pulp-operator are using mkdocs as well but they just publish to https://readthedocs.org/. I could set that up and it would be a lot less work than using docs.pulpproject.org. However, I feel like the latter is a better location. Thoughts?

ggainey:
> @pulp/cli it looks like both pulp_installer and pulp-operator are using mkdocs as well but they just publish to https://readthedocs.org/. I could set that up and it would be a lot less work than using docs.pulpproject.org. However, I feel like the latter is a better location. Thoughts?

We're trying to deprecate/remove all readthedocs and consolidate on docs.pulpproject, yeah? I'd like everything to be in one place.

docs/index.md
```diff
@@ -0,0 +1,5 @@
+# Pulp 3 CLI Docs
+
+This is the documentation for Pulp 3's command line interfact.
```
mdellweg, line 3:
```suggestion
This is the documentation for Pulp 3's command line interface.
```

docs/releasing.md
```diff
@@ -0,0 +1,10 @@
+# Releasing (for internal use)
+
+1. Generate the changelog (eg `towncrier --yes --version 0.1.0`) and commit
+1. Run `bumpversion release`, commit your local changes, and note the commit sha
+1. Run `bumpversion minor` to update the version to the next dev release version
+1. Push your commits, open a PR, and get it merged
+1. After your PR is merged, pull the latest changes from develop
+1. Now tag your release commit (e.g. `git tag 0.1.0`) and push to pulp/pulp-cli
+1. Monitor the build job and then check PyPI to make sure the package has been uploaded
```
mdellweg, line 9: Those jobs will publish the docs, right?
```suggestion
1. Monitor the build job and then check PyPI to make sure the package has been uploaded and the docs updated
```

mdellweg (commented):
A typo and a suggestion.
Great piece of work!

mdellweg (approved):
I think, we should merge this soon, to make the "edit on GH" banner work.
Hopefully it will lead to further improvements of the docs.

PR #249, opened 2021-05-20
console_scripts needs to be a list

evgeni:
See https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
         "pygments": ["pygments"],
     },
     entry_points={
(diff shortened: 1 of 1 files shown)


mdellweg (approved):
I believe, both work, but maybe this is a thing with older version of setup-tools.

evgeni:
> I believe, both work, but maybe this is a thing with older version of setup-tools.

It confused the hell out of my `pyp2rpm` at least.

Issue #251, opened 2021-05-20
Update RPM tests for autopublish


gerrod3:
Autopublish is changing in pulpcore 3.13 and RPM 3.13. It will no longer modify the distribution's `publication` field. RPM released 3.12 with the old version of autopublish so the current test needs to remain there, but an additional check needs to be added for RPM<=3.13.0.dev.

PR #252, opened 2021-05-20
Update RPM commands for new 3.13 auto-publish

gerrod3:
fixes: #251

### DIFF
--- a/pulpcore/cli/rpm/distribution.py
+++ b/pulpcore/cli/rpm/distribution.py
@@ -60,6 +60,7 @@ create_options = [
 update_options = [
     click.option("--base-path"),
     click.option("--publication"),
+    click.option("--repository", callback=_repository_callback),
 ]
 
 distribution.add_command(list_command(decorators=filter_options))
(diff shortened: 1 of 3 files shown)


pulpcore/cli/rpm/distribution.py
```diff
@@ -60,6 +60,7 @@ def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: s
 update_options = [
     click.option("--base-path"),
     click.option("--publication"),
+    click.option("--repository", callback=_repository_callback),
```
mdellweg, line 63: This is def more than just adjusted tests.

gerrod3: Well the option was missing, so I added it.

mdellweg: And it's needed for the autopublish operations?
Then we can reflect this as a feature in the changelog. Like "Adjusted autopublish to new version of ...".

CHANGES/251.bugfix
```diff
@@ -0,0 +1 @@
+Updated RPM auto-publish tests for future versions >=3.13.0.dev
```
daviddavis, line 1: The changelog is meant for end users so test changes don't need to be called out. I would probably just repurpose this for the repository option as @mdellweg suggests.

daviddavis (approved):

Issue #207, opened 2021-04-16
File publication list command is missing options


daviddavis:
It seems that the publication list command is missing options. See:

https://docs.pulpproject.org/pulpcore/restapi.html#operation/publications_file_file_list

PR #253, opened 2021-05-21
Added missing fields to publication list commands

gerrod3:
fixes: #207

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -335,6 +335,34 @@ base_path_contains_option = pulp_option(
     type=str,
 )
 
+content_in_option = pulp_option(
+    "--content",
+    "content__in",
+    help=_("Search for {entities} with these content hrefs in them (JSON list)"),
+    callback=load_json_callback,
+)
+
+pulp_created_gte_option = pulp_option(
+    "--created-after",
+    "pulp_created__gte",
+    help=_("Search for {entities} created at or after this ISO 8601 date"),
+    type=str,
+)
+
+pulp_created_lte_option = pulp_option(
+    "--created-before",
+    "pulp_created__lte",
+    help=_("Search for {entities} created at or before this ISO 8601 date"),
+    type=str,
+)
+
+publication_filter_options = [
+    content_in_option,
+    pulp_created_gte_option,
+    pulp_created_lte_option,
+    pulp_option("--repository-version", help=_("Search {entities} by repository version HREF")),
+]
+
 common_remote_create_options = [
     click.option("--name", required=True),
     click.option("--url", required=True),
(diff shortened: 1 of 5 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -271,6 +271,42 @@ def pulp_option(*args: Any, **kwargs: Any) -> Callable[[_F], _F]:
...
+pulp_created_lte_option = pulp_option(
+    "--created-lte",
+    "pulp_created__lte",
+    help=_("Search for {entities} created at or before this ISO 8601 date"),
+    type=str,
+)
+
+publication_filter_options = [
+    content_in_option,
+    pulp_created_exact_option,
```
mdellweg, line 304: An exact date to the millisecond does not sound like a useful filter to me.

pulpcore/cli/file/publication.py
```diff
@@ -49,8 +50,12 @@ def publication(ctx: click.Context, pulp_ctx: PulpContext, publication_type: str
     click.option(
         "--version", type=int, help=_("a repository version number, leave blank for latest")
     ),
+    click.option(
+        "--manifest",
+        help=_("Filename to use for manifest file containing metadata for all the files."),
+    ),
```
mdellweg, line 56: Not technically the same, but OK...

pulpcore/cli/common/generic.py
```diff
@@ -271,6 +271,42 @@ def pulp_option(*args: Any, **kwargs: Any) -> Callable[[_F], _F]:
...
 
+content_in_option = pulp_option(
+    "--content",
+    "content__in",
+    help=_("Search for {entities} with these content hrefs in them (JSON list)"),
+    callback=load_json_callback,
+)
+
+pulp_created_exact_option = pulp_option(
+    "--created-exact",
```
mdellweg, line 282: What about "created at"? And do you think we could add some fuzzyness? If i specify the date to the day, i might want to see publications made that day.

On a related note, the others could be named "creted before" and "created after".

pulpcore/cli/file/publication.py
daviddavis: I believe we should specify pulp_file >= 1.7.0 for this option. 

Also I don’t see this change reflected in the change log.

daviddavis: Oh sorry I thought this was the repository command. Ignore my comment about 1.7.0.

mdellweg (approved):

mdellweg:
The CI failure looks like #259 to me. Rebase should fix it.

mdellweg:
What's going on here?
https://github.com/pulp/pulp-cli/pull/253/checks?check_run_id=2765262634#step:5:1097

PR #262, opened 2021-06-02
Bypass chunked uploading if chunk size exceeds file size

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -38,22 +38,34 @@ class PulpArtifactContext(PulpEntityContext):
     HREF = "artifact_href"
     LIST_ID = "artifacts_list"
     READ_ID = "artifacts_read"
+    CREATE_ID = "artifacts_create"
 
-    def upload(self, file: IO[bytes], chunk_size: int = 1000000, check_exists: bool = True) -> Any:
+    def upload(self, file: IO[bytes], chunk_size: int = 1000000) -> Any:
         upload_ctx = PulpUploadContext(self.pulp_ctx)
         start = 0
         size = os.path.getsize(file.name)
-        sha256 = hashlib.sha256()
 
-        if check_exists:
-            for chunk in iter(lambda: file.read(chunk_size), b""):
-                sha256.update(chunk)
-            result = self.list(limit=1, offset=0, parameters={"sha256": sha256.hexdigest()})
-            if len(result) > 0:
-                click.echo("Artifact already exists.", err=True)
-                return result[0]["pulp_href"]
+        sha256 = hashlib.sha256()
+        for chunk in iter(lambda: file.read(chunk_size), b""):
+            sha256.update(chunk)
+        sha256_digest = sha256.hexdigest()
+        file.seek(0)
+
+        # check whether the file exists before uploading
+        result = self.list(limit=1, offset=0, parameters={"sha256": sha256_digest})
+        if len(result) > 0:
+            click.echo("Artifact already exists.", err=True)
+            return result[0]["pulp_href"]
 
         click.echo(f"Uploading file {file.name}", err=True)
+
+        if chunk_size > size:
+            # if chunk_size is bigger than the file size, just upload it directly
+            artifact: Dict[str, Any] = self.create(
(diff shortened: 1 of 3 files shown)


pulpcore/cli/core/context.py
```diff
@@ -38,13 +38,20 @@ class PulpArtifactContext(PulpEntityContext):
     HREF = "artifact_href"
     LIST_ID = "artifacts_list"
     READ_ID = "artifacts_read"
+    CREATE_ID = "artifacts_create"
+
+    def _create(self, file: IO[bytes]) -> Any:
```
mdellweg, line 43:
```suggestion
    def create(self, file: IO[bytes]) -> Any:
```
Why should we hide that interface?

mdellweg (approved):
Great idea!

pulpcore/cli/core/context.py
daviddavis: Mypy gets upset that the signature doesn't match [`create()` from `PulpEntityContext`](https://github.com/pulp/pulp-cli/blob/8d164d7a60321e970ff710cb0e8ebd71e4618a55/pulpcore/cli/common/context.py#L439-L444).

I could change the name of this method? Or I could use the signature from `create()` in `PulpEntityContext` and pull the file from `body` but that seems a bit awkward.

daviddavis:
I'm debating about adding an explicit option to also bypass chunk uploading. Wdyt @mdellweg ?

mdellweg:
> I'm debating about adding an explicit option to also bypass chunk uploading. Wdyt @mdellweg ?

Then i would argue, that we'd need a force-chunks option, because large files only work as chunks. I wouldn't do either without anyone asking for it.

pulpcore/cli/core/context.py
mdellweg: I would extend the `PulpEntityContext` create with `uploads` the same way as update has it.

pulpcore/cli/core/context.py
```diff
@@ -38,13 +38,20 @@ class PulpArtifactContext(PulpEntityContext):
...
+        return self.pulp_ctx.call(self.CREATE_ID, uploads={"file": file.read()})
 
     def upload(self, file: IO[bytes], chunk_size: int = 1000000, check_exists: bool = True) -> Any:
         upload_ctx = PulpUploadContext(self.pulp_ctx)
         start = 0
         size = os.path.getsize(file.name)
         sha256 = hashlib.sha256()
 
+        if chunk_size > size:
+            return self._create(file)["pulp_href"]
```
mdellweg, line 53: Is it possible to pass the sha256 to this create call to add verification as with the chunked upload?

pulpcore/cli/core/context.py
```diff
@@ -38,20 +38,28 @@ class PulpArtifactContext(PulpEntityContext):
     HREF = "artifact_href"
     LIST_ID = "artifacts_list"
     READ_ID = "artifacts_read"
+    CREATE_ID = "artifacts_create"
 
-    def upload(self, file: IO[bytes], chunk_size: int = 1000000, check_exists: bool = True) -> Any:
```
daviddavis, line 42: We're not using `check_exists=False` anywhere and maintaining it was making the code more and more complex.

pulpcore/cli/core/context.py
```diff
@@ -38,20 +38,29 @@ class PulpArtifactContext(PulpEntityContext):
...
+        if chunk_size > size:
+            # if chunk_size is bigger than the file size, just upload it directly
+            result = self.create({"sha256": sha256_digest}, uploads={"file": file.read()})
+            return result["pulp_href"]
+
+        # check whether the file exists before chunk uploading
+        result = self.list(limit=1, offset=0, parameters={"sha256": sha256_digest})
+        if len(result) > 0:
+            click.echo("Artifact already exists.", err=True)
+            return result[0]["pulp_href"]
```
mdellweg, line 63: Why not move this check before the atomic upload?

daviddavis: I felt like this check was less useful and less worthwhile it for atomic uploads but I suppose it could save users some time if they are uploading big files. Will update.

mdellweg (approved):

Issue #271, opened 2021-06-18
Add allow-uploads option to python distribution commands


gerrod3:
Added in pulp_python 3.4.0: https://docs.pulpproject.org/pulp_python/restapi.html#operation/distributions_python_pypi_create

PR #272, opened 2021-06-18
Add allows-uploads/block-uploads option to python distribution commands

gerrod3:
fixes: #271

### DIFF
--- a/pulpcore/cli/python/distribution.py
+++ b/pulpcore/cli/python/distribution.py
@@ -2,7 +2,7 @@ import gettext
 
 import click
 
-from pulpcore.cli.common.context import PulpContext, pass_pulp_context
+from pulpcore.cli.common.context import PluginRequirement, PulpContext, pass_pulp_context
 from pulpcore.cli.common.generic import (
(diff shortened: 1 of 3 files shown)


pulpcore/cli/python/distribution.py
```diff
@@ -53,6 +54,11 @@ def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: s
     click.option("--base-path"),
     click.option("--publication"),
     repository_option,
+    pulp_option(
+        "--allow-uploads/--block-uploads",
```
mdellweg, line 58: Really not sure here: Would "--no-allow-uploads" be more intuitive?

mdellweg (approved):

pulpcore/cli/python/distribution.py
mdellweg: I know, we also have "--dry-run/--force"...

ggainey: I think I favor the allow/block verbiage, from a user's standpoint.

PR #275, opened 2021-06-23
Create issue templates

daviddavis:


daviddavis:
When you go to create an issue, there is an option to create a blank (ie uncategorized issue). So I think two templates (one for bug and one for feature) should probably suffice.

Here's an example from foreman: https://github.com/theforeman/foreman-ansible-modules/issues/new/choose

daviddavis:
Closing in favor of #276

PR #276, opened 2021-06-23
Create issue templates

daviddavis:
When you go to create an issue, there is an option to create a blank (ie uncategorized issue). So I think two templates (one for bug and one for feature) should probably suffice.

Here's an example from foreman: https://github.com/theforeman/foreman-ansible-modules/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,24 @@
+---
+name: 🐛 Bug report
+about: Create a report to help us improve the cli
+labels: bug
+
+---
+
+## Summary
+<!-- A clear and concise description of what the bug is. -->
+
+## Steps to reproduce
+<!-- Steps to reproduce the behavior. Please list actual cli commands. -->
+
+## Expected behavior
+<!-- A clear and concise description of what you expected to happen. -->
+
+## Stacktrace/Error log
+<!-- If possible, add a stacktrace and/or logs -->
+
+## Pulp and pulp-cli version info
+<!-- Verbatim output of `pulp --version` and `pulp status` -->
+
+## Additonal context
+<!-- Add any other context about the problem here. -->
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,14 @@
+---
+name: ✨ Feature request
+about: Suggest an idea for this project
+labels: feature request
+
+---
+
+## Summary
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,24 @@
...
+<!-- Steps to reproduce the behavior. Please list actual cli commands. -->
+
+## Expected behavior
+<!-- A clear and concise description of what you expected to happen. -->
+
+## Stacktrace/Error log
+<!-- If possible, add a stacktrace and/or logs -->
+
+## Pulp and pulp-cli version info
+<!-- Which version of pulp-cli and also include pulp status output -->
```
mdellweg, line 21:
```suggestion
<!-- Verbatim output of `pulp --version` and `pulp status` -->
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,13 @@
...
+about: Suggest an idea for this project
+labels: feature request
+
+---
+
+## Summary
+<!-- include a summary -->
+
+## Examples
+<!-- If applicable, include some proposed cli command examples -->
```
mdellweg, line 12:
```suggestion
<!-- If applicable, include some proposed cli command examples -->
<!-- If applicable, include some related httpie call examples -->
```
Just a thought, not sure if it's a good one.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,13 @@
+---
+name: ✨ Feature request
+about: Suggest an idea for this project
+labels: feature request
+
+---
+
+## Summary
+<!-- include a summary -->
```
mdellweg, line 9: Maybe ask for the api to wrap or add a disclamer that we will not collect pulpcore feature requests here.

daviddavis:
I just found out about [issue forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/). I think I am going to use them if that makes sense?

bmbouter:
This type of pattern is going to be great for our users as more of our repos move over to GHI!

mdellweg:
> I just found out about [issue forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/). I think I am going to use them if that makes sense?

I'm not sure how much benefit a form with only freetext input is over the comment based template.
And i would not add a dropdown for versions, for example.
So what input fields would you put into the form?

We can also merge this plain template and then decide later to turn it into an input form.

mdellweg (approved):

PR #277, opened 2021-06-25
Update to click 8

### LINKED ISSUE #256: Support click 8
Tasks:

- [ ] [click-shell](https://github.com/clarkperkins/click-shell) doesn't support click 8
~- [ ] get bash tab completion working~
- [ ] mypy linting is hard to impossible to fix for 7 and 8 simultaneously

edit: The shell completion issue is actually: https://github.com/pallets/click/issues/1940

mdellweg:
This will use a compatible fork of click-shell for the time being.
Also update mypy.

fixes #256

### DIFF
--- a/pulpcore/cli/ansible/distribution.py
+++ b/pulpcore/cli/ansible/distribution.py
@@ -36,7 +36,8 @@ def _repository_callback(
 ) -> EntityFieldDefinition:
     # Pass None and "" verbatim
     if value:
-        pulp_ctx: PulpContext = ctx.find_object(PulpContext)
+        pulp_ctx = ctx.find_object(PulpContext)
+        assert pulp_ctx is not None
         return PulpAnsibleRepositoryContext(pulp_ctx, entity={"name": value})
     return value
 
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -1,6 +1,18 @@
 import gettext
 import json
-from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Mapping,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
 
 import click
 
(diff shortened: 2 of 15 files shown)


pulpcore/cli/ansible/distribution.py
```diff
@@ -36,7 +36,8 @@ def _repository_callback(
 ) -> EntityFieldDefinition:
     # Pass None and "" verbatim
     if value:
-        pulp_ctx: PulpContext = ctx.find_object(PulpContext)
+        pulp_ctx = ctx.find_object(PulpContext)
+        assert pulp_ctx is not None
```
mdellweg, line 40: These kind of changes are actually a good sign, because they mean that click declared a proper return type `Optional[PulpContext]` instead of `Any`.

pulpcore/cli/common/generic.py
```diff
@@ -82,18 +101,21 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         ).strip()
         super().__init__(*args, **kwargs)
 
-    def handle_parse_result(self, ctx: click.Context, opts: Dict[str, Any], args: Any) -> Any:
+    def handle_parse_result(
+        self, ctx: click.Context, opts: Mapping[str, Any], args: List[Any]
```
daviddavis, line 105: Why did this go from Dict to Mapping?

pulpcore/cli/ansible/distribution.py
daviddavis: What prompted this change?

pulpcore/cli/common/generic.py
mdellweg: Because the function we inherit it from is using Mapping (which is more general).

pulpcore/cli/ansible/distribution.py
mdellweg: Click8 changed the return-type of `Context.find_object` from `Any` to `Optional[WhateverTypeYouRequested]`.
So click is more accurate, and we need to make sure it's not returning `None`. Or mypy will complain.

daviddavis (approved):
LGTM

ggainey (approved):
daviddavis asked all the questions I would have asked - lg2m, too. I assume the nightly-failure is part of today's GHA Failure Fun? I think we're good to merge!

PR #279, opened 2021-06-25
Also wait of task state is "canceling"

mdellweg:
[noissue]


daviddavis:
Also wait *if* task state is "canceling"

daviddavis (approved):

PR #280, opened 2021-06-26
Prepare 0.10.0 release

mdellweg:
This is a cleanup PR where i intend to collect some housekeeping commits before we will release the next minor version.

Please look at the commits individually.


pulpcore/cli/ansible/repository.py
```diff
@@ -32,18 +33,15 @@
...
-    return value
+remote_option = resource_option(
+    "--remote",
+    default_plugin="ansible",
+    default_type="collection",
+    context_table={
+        "ansible:collection": PulpAnsibleCollectionRemoteContext,
+        "ansible:role": PulpAnsibleRoleRemoteContext,
+    },
+)
```
mdellweg, line 44: @gerrod3 
This needs an update of the ansible sync workflow:

https://github.com/pulp/pulp_ansible/blob/master/docs/_scripts/sync.sh#L2

ggainey (approved):
This all looks like reasonable housekeeping to me - lg2m!

PR #282, opened 2021-06-27
Capture container logs by test

mdellweg:
~Rather an experiment at this stage.~

~Need to find a way to only use this fixture in CI.~


ggainey (approved):
The test-failure, shows that the PR is doing what we want - lg2m!

PR #283, opened 2021-06-28
Stop using find_namespace_packages

### LINKED ISSUE #248: Don't use setuptools > 39.2.0 features in setup.py
Ohai,

currently both EL7 and EL8 ship with setuptools 39.2.0 by default, and I'd prefer not to touch such a core package of the system.

However, `pulp-cli/setup.py` uses `find_namespace_packages` which was introduced in setuptools 40.1.0. This makes it impossible to build `pulp-cli` from source on EL7/8 :(

Would it be possible to restrict the used featureset to what's in 39.2, so that we can build packages?

mdellweg:
Also provide the shell entrypoint in a not namespaced module.
This provides compatibility with older versions of setuptools.

fixes #248

Co-authored-by: Evgeni Golov <evgeni@golov.de>

### DIFF
--- a/CHANGES/248.bugfix
+++ b/CHANGES/248.bugfix
@@ -1,0 +1,1 @@
+Use ``find_packages`` instead of ``find_namespace_packages`` in setup to be compatible with ``setuptools<39.2.0``.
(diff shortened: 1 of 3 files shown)


evgeni (approved):

evgeni:
I wonder if there is a reasonable way to add a test for that into the matrix

PR #288, opened 2021-06-30
Add pulp-cli module to pypi distribution

### LINKED ISSUE #287: sdist on PyPI doesn't contain `pulp_cli` fake module
https://github.com/pulp/pulp-cli/pull/283 introduced a fake, non-namespaced module called `pulp_cli` which is needed to make old setuptools happy. sadly, this is only added to the distributed source when the tarball is built using such an old setuptools, which is usually not the case.

this results in the following error when trying to install from such an sdist:
```
error: package directory 'pulp_cli' does not exist
```

mdellweg:
fixes #287

### DIFF
--- a/CHANGES/287.bugfix
+++ b/CHANGES/287.bugfix
@@ -1,0 +1,1 @@
+Unconditionally add the fake not namespaced pulp_cli to the distribution on pypi to make it able to be consumed by setuptools<40.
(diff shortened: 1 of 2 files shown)


CHANGES/287.bugfix
```diff
@@ -0,0 +1 @@
+Add the fake not namespaced pulp_cli to the distribution on pypi to make it able to be consumed by setuptools<40
```
evgeni, line 1: "Always add" or "Unconditionally add" is probably more precise :)

evgeni (approved):

CHANGES/287.bugfix
mdellweg: Well it adds it to what will be on pypi later...

PR #292, opened 2021-07-01
Document steps to use pulp-cli with Pulp running on Katello content-p…

wbclark:
…roxy

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -87,6 +87,8 @@ cert = "/etc/pki/katello/certs/pulp-client.crt"
 key = "/etc/pki/katello/private/pulp-client.key"
 ```
 
-As Katello uses Pulp as a backend all modifying actions in Pulp should be performed via Katello.
+Katello content-proxy servers do not possess the certificates required to communicate with Pulp; therefore you should install pulp-cli only on the primary server, and override the base_url setting on the command line when you need to connect to Pulp running on a content-proxy rather than Pulp on the primary Katello server: `--base-url https://capsule.example.com`
+
+As Katello uses Pulp as a backend, all modifying actions in Pulp should be performed via Katello.
 Therefore you are also strongly encourged to set `dry_run = true`, to prevent accidentally calling into dangerous commands.
 This setting can in turn be overwritten on the command-line with the `--force` flag.


docs/install.md
```diff
@@ -88,6 +88,8 @@ key = "/etc/pki/katello/private/pulp-client.key"
 verify_ssl = false
 ```
 
-As Katello uses Pulp as a backend all modifying actions in Pulp should be performed via Katello.
+Katello content-proxy servers do not possess the certificates required to communicate with Pulp; therefore you should install pulp-cli only on the primary server, and override the base_url setting on the command line when you need to connect to Pulp running on a content-proxy rather than Pulp on the primary Katello server: `--base-url https://capsule.example.com`
```
mdellweg, line 91: Maybe the "--profile" option comes in handy here.

wbclark: Asking the user to set this up might be a bit overkill for this use case. Either way they are typing `--base-url https://capsule1.example.com` or `--config ~/.config/pulp/capsule1_settings.toml` when invoking commands from the CLI, and the base_url is the only option that would need to change (same certificates are used for authentication) so it should take about the same amount of typing either way.

Maybe it's nicer to use a config file to remember a complex URL, particularly in an environment with many Capsules. In that case, I think generating these automatically for the admin would be a nice feature, I suppose at the time of Capsule certs generation or Capsule registration makes a fair amount of sense.

In the meantime could we consider merging this as-is, since it should add a bit of needed clarity for users now, while we figure out how to deliver an even better experience?

mdellweg: My thinking was, if you have a `[cli-capsule1]` section in your config, you can run `pulp -p capsule1 ...`.
And yes, this "--profile" option is probably not well documented.

ehelms: I like that profile option, it may come down to how we manage the config file from our installer in the end for design. I think this documentation of using `--base-url` is a good starting point.

mdellweg (approved):

docs/install.md
wbclark: > My thinking was, if you have a `[cli-capsule1]` section in your config, you can run `pulp -p capsule1 ...`.
> And yes, this "--profile" option is probably not well documented.

Very helpful, thank you

PR #294, opened 2021-07-02
don't configure verify_ssl=False for Katello

evgeni:
Katello does deploy its certs to the system store properly, so once you have set `base_url` to the FQDN, you can verify the chain just fine.


evgeni:
damn linter

mdellweg:
I want's to see a `[noissue]` in the commit message...

evgeni:
> It want's to see a `[noissue]` in the commit message...

I know, I just find this requirement nonsense :woman_shrugging:

daviddavis (approved):

PR #295, opened 2021-07-02
Rotate test matrix

mdellweg:
3.14 in, 3.9 out

[noissue]


evgeni:
I always found matrix rotation rather complicated, and docker hub seems to agree: there is no 3.14 image (yet?)

PR #296, opened 2021-07-03
relax PyYAML and requests deps

evgeni:
let's assume they know how semver works

[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,8 +44,8 @@ setup(
     install_requires=[
         "click>=7.1.2,<9.0.0",
         "packaging",
-        "PyYAML~=5.4.1",
-        "requests~=2.25.1",
+        "PyYAML~=5.3",
+        "requests~=2.24",
         "toml==0.10.2",
     ],
     extras_require={


pcreech (approved):

pcreech:
Oops, didn't realize this was on the overall pulp-cli repo.  Apologies for the errant approval.

daviddavis:
I think this will trigger dependabot to attempt to update the dependencies. Might be good to add [a dependabot ignore rule](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#creating-ignore-conditions-from-dependabot-ignore).

evgeni:
> I think this will trigger dependabot to attempt to update the dependencies. Might be good to add [a dependabot ignore rule](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#creating-ignore-conditions-from-dependabot-ignore).

Will it? There are no newer PyYAML or requests that don't match the version range *today*, so I am not sure to what it would try to update to. And if there will be, it might be a good trigger to re-consider the version range?

daviddavis:
Yea, I believe dependabot will bump the floor even if the requirement doesn't preclude the new release version. Here's an example:

https://github.com/pulp/pulpcore/pull/1462

daviddavis:
We could just ignore the dependabot PRs for these deps. We might forget though so it might make sense to leave a comment.

mdellweg:
Let's merge this, and see what dependabot has to offer. WDYT?

evgeni:
Works for me!

daviddavis:
Me too :+1:

mdellweg (approved):

PR #303, opened 2021-07-14
Enable exporting of ansible repos

daviddavis:


mdellweg:
If we add a changelog, we can merge this. Just keep in mind that the interface will likely change (multiple times, maybe) in the near future.
#306

daviddavis:
@rochacbruno fyi ^

rochacbruno (approved):
Tested this patch and worked for me

mdellweg (approved):

PR #307, opened 2021-07-16
Add a global config file, rename the user config

### LINKED ISSUE #290: As a user, I want to be able to deploy a global configuration file
Right now pulp-cli only reads `~/.config/pulp/settings.toml` (btw, why doesn't it mention *cli* enywhere?).

I would like to be able to deploy a `/etc/pulp/cli/settings.toml` (or whatever other path you can think of, in `/etc/pulp`) which will be loaded *before* the one in `~/.config`, so I can deploy "global" configuration defaults that vary from the ones in the app itself, but should be valid for all users.

Background: on Katello installations, I want to deploy pulp-cli with a default of `dry-run = true`.

### LINKED ISSUE #293: inconsistent naming of "settings"
```console
# pulp config create
Created config file at '/root/.config/pulp/settings.toml'.
```

the cli param is `config` but the file is called `settings`? ;-)

mdellweg:
This will read the global (/etc/pulp/cli_config.toml) first and merge
with the user local ones (settings.toml, cli_config.toml) in that order.
Creating and editing config files default to cli_config now.

fixes #290
fixes #293


daviddavis:
LGTM. I wonder if we should call this file `cli.toml` since toml is a config file format? Sort of like cli_conf.conf vs cli.conf.

daviddavis (approved):

mdellweg:
> LGTM. I wonder if we should call this file `cli.toml` since toml is a config file format? Sort of like cli_conf.conf vs cli.conf.

I always thought Tom's Obvious Minimal Language to play in the json/yaml ballpark, but in any case, the path we are putting the file in gives the conffile away, too.

mdellweg:
@evgeni do you think, this serves your need?

evgeni:
This reads like what I'd need, yeah!

PR #309, opened 2021-07-18
Turn 'verbose' into a config value

mdellweg:
[noissue]


mdellweg:
I used this to improve debugging on failed tests. Just set verbose=3 in 'tests/cli.toml'.

ggainey (approved):

PR #312, opened 2021-07-21
Set minimum for limit to 1

### LINKED ISSUE #311: `--limit 0` behavior differs from API
Not sure if this is a problem but I noticed that `--limit 0` behaviors different than supplying limit=0 when hitting the api:

```
# returns all objects
$ http :/pulp/api/v3/repositories/file/file/ limit==0

$ pulp file repository list --limit 0
Not all -1 entries were shown.
[]
```

Also the message seems odd too (`Not all -1 entries were shown.`)

daviddavis:
fixes #311

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -311,7 +311,7 @@ def resource_option(*args: Any, **kwargs: Any) -> Callable[[_FC], _FC]:
 limit_option = pulp_option(
     "--limit",
     default=DEFAULT_LIMIT,
-    type=int,
+    type=click.IntRange(1),
     help=_("Limit the number of {entities} to show."),
 )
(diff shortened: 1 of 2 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -311,7 +311,7 @@ def _multi_option_callback(
 limit_option = pulp_option(
     "--limit",
     default=DEFAULT_LIMIT,
-    type=int,
+    type=click.IntRange(1),
```
mdellweg, line 314: This translates to 1..\infty?

mdellweg (approved):

pulpcore/cli/common/generic.py
daviddavis: Yes, exactly.

PR #313, opened 2021-07-21
Document the config profile feature

mdellweg:
[noissue]

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -24,7 +24,7 @@ pip install -e .
 ## Configuration
 
 The CLI can be configured by using a toml file.
-By default the location of this file is `~/.config/pulp/settings.toml`.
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -24,7 +24,7 @@ pip install -e .
 ## Configuration
 
 The CLI can be configured by using a toml file.
-By default the location of this file is `~/.config/pulp/settings.toml`.
+By default the location of this file is `~/.config/pulp/cli.toml`.
```
daviddavis, line 27: :+1:

daviddavis (approved):
LGTM. My only gripe is that our docs formatting is inconsistent with the rest of Pulp. But I suppose we can discuss another time.

PR #316, opened 2021-07-23
RPM sync skip_types

mdellweg:


daviddavis (approved):

mdellweg:
pulp_rpm sync seems to be broken on the 3.8 container. Is this a known fact to anyone?

https://github.com/pulp/pulp-cli/pull/316/checks?check_run_id=3149857590#step:5:797

@pulp/rpm

mdellweg:
As this is presumably happening to the skip_types fix, I will isolate the other two in a separate PR (And silently transfer your approval, @daviddavis ).

Issue #261, opened 2021-06-01
Validate when pulp-ansible and galaxy-ng are installed together


fao89:
Galaxy NG blocks pulp-ansible endpoints:
https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/webserver_snippets/nginx.conf

When Galaxy NG and Pulp Ansible are installed together, users will get 403 on pulp ansible calls.
This can confuse users with authentication errors, maybe we can add  some helpful error message like:
```
pulp-cli does not support galaxy_ng yet
```

bmbouter:
+1 to an error message when galaxy_ng is installed and pulp_ansible is attempted to be used.

pulp-cli can't fix this though, is there a galaxy_ng issue tracking the removal of [this snippet](https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/webserver_snippets/nginx.conf#L14-L32)?

mdellweg:
We may be able add a `PluginRequirement("galaxy_ng", max_version="0.0.0")` to the `pulp ansible` command group. But i am not sure if the resulting error message would serve your needs.

mdellweg:
We should use `has_plugin` to issue this error message.
This might impact more endpoints, like users and groups.

mdellweg:
I this still true?

mdellweg:
I believe the api is exposed on a redirected path which pulp-cli supports.

PR #319, opened 2021-07-26
Add "inverted" to PluginRequirement

mdellweg:
Also remove the deprecated interface from has_plugin and needs_plugin.

[noissue]

This should help to implement #261 .

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -53,27 +53,38 @@ def my_sub_command(entity_ctx):
 Each pulp CLI release is designed to support multiple pulp server versions.
 The entity contexts should be used to implement necessary server version-dependent workarounds.
 To facilitate this, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods.
+While `has_plugin` will evaluete immediately, `needs_plugin` can be seen as a deferred assertion.
+It will raise an error, once the first access to the server is attempted.
 
 ```python
 class MyEntityContext(PulpEntityContext):
     def show(self, href):
-        if not self.pulp_ctx.has_plugin("pulp_my_content", min_version="1.2.3"):
+        if self.pulp_ctx.has_plugin(PluginRequirement("my_content", min="1.2.3", inverted=True)):
+            # Versioned workaroud
+            # see bug-tracker/12345678
             return lookup_my_content_legacy(href)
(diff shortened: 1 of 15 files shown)


pulpcore/cli/common/context.py
```diff
@@ -281,13 +223,19 @@ def needs_plugin(
                 if plugin.max is not None:
                     specifier += f"{separator}<{plugin.max}"
                 feature = plugin.feature or _("this command")
-                raise click.ClickException(
-                    _(
+                if plugin.inverted:
+                    msg = _(
+                        "The server provides the pulp component '{specifier}',"
+                        " which prevents to use {feature}."
```
ggainey, line 229: ^to use^the use of

ggainey (changes requested):
One minor working-tweak, otherwiuse lg2m

docs/architecture.md
```diff
@@ -57,7 +57,7 @@ To facilitate this, the `PulpContext` provides the `needs_plugin` and `has_plugi
 ```python
 class MyEntityContext(PulpEntityContext):
     def show(self, href):
-        if not self.pulp_ctx.has_plugin("pulp_my_content", min_version="1.2.3"):
+        if not self.pulp_ctx.has_plugin(PluginRequirement("pulp_my_content", min="1.2.3")):
```
daviddavis, line 60: AFAICT, these changes are not related to the inverted feature that you're also introducing here. Is that correct?

daviddavis:
Might be worth adding some information about this inverted feature here:

https://github.com/pulp/pulp-cli/blob/develop/docs/architecture.md#version-dependent-code-paths

docs/architecture.md
mdellweg: In a way. Those functions have the "old" interface aka "passing parameters individually" and the new one, "passing the `PluginRequrement` as a single object".
The old interface was meant to go away anyway, and i didn't bother to add the `inverted` to it. So i decided for removing it here.
Why do we have the sligtly more complex "new" interface? `needs_plugin` works deferred, and it was very convenient to store and pass around the `PluginRequirement` objects.

ggainey (approved):

daviddavis (approved):

PR #321, opened 2021-07-27
Fix references to settings.toml

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/cli/common/__init__.py
+++ b/pulpcore/cli/common/__init__.py
@@ -86,7 +86,7 @@ def _config_callback(ctx: click.Context, param: Any, value: Optional[str]) -> No
 @click.option(
     "--config",
     type=click.Path(resolve_path=True),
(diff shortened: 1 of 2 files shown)


pulpcore/cli/common/__init__.py
```diff
@@ -86,7 +86,7 @@ def _config_callback(ctx: click.Context, param: Any, value: Optional[str]) -> No
 @click.option(
     "--config",
     type=click.Path(resolve_path=True),
-    help=_("Path to the Pulp settings.toml file"),
+    help=_("Path to the Pulp cli.toml file"),
```
mdellweg, line 89: Actually, the file can be named anything.
How about: "Path of a Pulp CLI settings file to use instead of the default location"

mdellweg (approved):

mdellweg (approved):

PR #325, opened 2021-07-30
Add --uuid for task commands

mdellweg:
The workers only log the uuid of tasks. This way it will be easier to
take that information and work with tasks.

### DIFF
--- a/pulpcore/cli/core/task.py
+++ b/pulpcore/cli/core/task.py
@@ -1,13 +1,37 @@
 import gettext
+from typing import Optional
 
 import click
 
-from pulpcore.cli.common.context import PulpContext, pass_entity_context, pass_pulp_context
-from pulpcore.cli.common.generic import list_command
+from pulpcore.cli.common.context import (
+    PulpContext,
+    PulpEntityContext,
+    pass_entity_context,
+    pass_pulp_context,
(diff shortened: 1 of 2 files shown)


daviddavis:
> The workers only log the uuid of tasks.

Is this like in the system logs?

mdellweg:
Yes i am talking about system logs (aka journalctl).

Also uuid is the closest thing to a (unique) name the tasks have.

daviddavis:
Is correlation id not an option?

```
$ journalctl -n 2
pulp [f35a81b318bc4d5eaac616f65bf021fc]: pulpcore.tasking.pulpcore_worker:INFO: Starting task 072ae5c9-4c0e-406c-b8f5-d6124f12e75e
pulp [f35a81b318bc4d5eaac616f65bf021fc]: pulpcore.tasking.pulpcore_worker:INFO: Task completed 072ae5c9-4c0e-406c-b8f5-d6124f12e75e

$ pulp task list --cid 
```

mdellweg:
Correlatio ID is a filter, and one user action can trigger multiple tasks. They would share the cid.

daviddavis (approved):
LGTM. A test might be nice but is not required.

PR #326, opened 2021-07-30
Add configurable filter for plugins

### LINKED ISSUE #291: As a user, I want to be able to disable certain plugins
pulp-cli supports a variety of pulpcore plugins, but currently doesn't have a way to limit the CLI to the ones that are actually installed.

I am not sure whether querying the API *before* doing anything (including tab completion) is a good idea, but having a way to list plugins that shall be enabled (or disabled, I don't care too much), would be a good middle ground

That way Katello could deploy a config (after #290 is done) where we match the CLI plugins and the Server plugins :)

mdellweg:
fixes #291

### DIFF
--- a/CHANGES/291.feature
+++ b/CHANGES/291.feature
@@ -1,0 +1,1 @@
+Added the ability to specify a list in the system wide config file to filter plugins on loading.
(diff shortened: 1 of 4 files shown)


mdellweg:
@evgeni does this match your expectations?

mdellweg:
Closing as this is not a nice solution. We can always factor plugins into their own repositories if we want to make them separately installable.

PR #327, opened 2021-08-03
Add collection version content list command

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/cli/ansible/content.py
+++ b/pulpcore/cli/ansible/content.py
@@ -1,0 +1,31 @@
+import gettext
+
+import click
+
+from pulpcore.cli.ansible.context import PulpAnsibleCollectionVersionContext, PulpAnsibleRoleContext
+from pulpcore.cli.common.context import PulpContext, pass_pulp_context
+from pulpcore.cli.common.generic import list_command
+
+_ = gettext.gettext
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "content_type",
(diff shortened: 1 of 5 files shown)


pulpcore/cli/ansible/content.py
```diff
@@ -0,0 +1,29 @@
...
+_ = gettext.gettext
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "content_type",
+    type=click.Choice(["collection-version"], case_sensitive=False),
+    default="collection-version",
```
mdellweg, line 18: I'm wondering if we should expose this as a "collection" to the user.

mdellweg: Let's go for consistency.

fao89: the closest we get to the galaxy user experience, the better

pulpcore/cli/ansible/content.py
```diff
@@ -0,0 +1,29 @@
...
+)
+@pass_pulp_context
+@click.pass_context
+def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> None:
+    if content_type == "collection-version":
+        ctx.obj = PulpAnsibleCollectionVersionContext(pulp_ctx)
+    else:
```
mdellweg, line 25: Any chance we can add the code path for roles right now? Also OK to do it later.

mdellweg (approved):

mdellweg:
The failures seem unrelated now. Have any of the containers been rebuilt?

PR #338, opened 2021-08-11
Set a limit on the content list command

daviddavis:
[noissue]


daviddavis:
This is sort of a workaround due to a bad collection with linebreaks in the pulp_ansible CI. I figure it's something we want anyway though since there's a lot of collection version and we want to limit the list to a more reasonable number.

daviddavis:
Closing in favor of https://github.com/pulp/pulp-cli/pull/339.

PR #339, opened 2021-08-11
Count collections with grep

fao89:
[noissue]

### DIFF
--- a/tests/scripts/pulp_ansible/test_sync.sh
+++ b/tests/scripts/pulp_ansible/test_sync.sh
@@ -57,4 +57,6 @@ test "$(echo "$OUTPUT" | jq -r length)" -eq 2
 expect_succ pulp ansible repository version show --repository "cli_test_ansible_repository" --version 2
 test "$(echo "$OUTPUT" | jq -r '.content_summary.present."ansible.collection_version".count')" -gt 0
 expect_succ pulp ansible content list
-test "$(echo "$OUTPUT" | jq -r length)" -gt 0
+# jq fails to parse the collection data
+# collection_version.docs_blob brings HTML snippets which may be adding linebreaks
+test "$(echo "$OUTPUT" | grep -c pulp_href)" -gt 0


fao89:
let's see if the issue is on our fixture
https://github.com/pulp/pulp_ansible/pull/615

mdellweg:
I'm not so sure, who is producing the invalid json here. FWIW, if pulp_ansible answered with invalid json, the cli should fail, and not jq.
Also can we consider this change a temporary workaround, and keep the current line as a comment?

daviddavis:
I have no objection to the cli failing when it encounters invalid json from a pulp server but I kind of thought that the cli just outputted the data it gets from pulp for the most part with no validation, etc.

I think though that pulp_ansible should not return invalid json and that's something we need to fix in pulp_ansible probably.

mdellweg:
> I have no objection to the cli failing when it encounters invalid json from a pulp server but I kind of thought that the cli just outputted the data it gets from pulp for the most part with no validation, etc.

It parses the returned data as json and formats it as json/yaml again. But maybe it's json parser is accepting things that jq doesn't. That is why i really want (someone) to fully understand, what's going on here.

> I think though that pulp_ansible should not return invalid json and that's something we need to fix in pulp_ansible probably.

agree, absolutely

fao89:
> > I have no objection to the cli failing when it encounters invalid json from a pulp server but I kind of thought that the cli just outputted the data it gets from pulp for the most part with no validation, etc.
> 
> It parses the returned data as json and formats it as json/yaml again. But maybe it's json parser is accepting things that jq doesn't. That is why i really want (someone) to fully understand, what's going on here.
> 
> > I think though that pulp_ansible should not return invalid json and that's something we need to fix in pulp_ansible probably.
> 
> agree, absolutely

what I have so far is:
- when I ssh to CI, I can get the list without problems (running pulp ansible content list);
- I can even jq it without problems (running pytest -m pulp_ansible -k test_sync);
- the issue only happens when the CLI tests run just after the functional tests

issues:
- the output of every collection_version is huge because docs_blob brings docs html
- the output is so huge that is hard to see where the linebreak come from

I have no idea how jq evaluates the output from the CLI, but when I run `pulp ansible content list` on the terminal, the output fits the size of my terminal creating linebreaks on the documentation. So it becomes pretty unclear if the linebreak was there because the docs had it or because it didn't fit my screen.

daviddavis:
Debugging some things it looks like the rendered README ends up in the json response. When I import `pulp.test_asset` for example, and run `pulp ansible content list` for docs_blob, it shows:

```
    "docs_blob": {
      "contents": [],
      "collection_readme": {
        "html": "<h1>Ansible Collection - pulp.testing_asset</h1>\n<p>Documentation for the collection.</p>",
        "name": "README.md"
      },
      "documentation_files": []
    },
```

Notice that the newline in collection_readme is properly escaped as `\n`. 

However, when I echo the output in the CI, it shows:

```
[...] "docs_blob": {"contents": [], "collection_readme": {"html": "<h1>Ansible Collection - pulp.testing_asset</h1>
2021-08-16T17:17:13.6894587Z <p>Documentation for the collection.</p>", "name": "README.md"}, "documentation_files": []}
```

Notice the newline after the closing h1 tag.

What's more strange is that when I run the cli test sync locally I get different output (with `\n`):

```
"docs_blob": {"contents": [], "collection_readme": {"html": "<h1>Ansible Collection - pulp.testing_asset</h1>\n<p>Documentation for the collection.</p>", "name": "README.md"}, "documentation_files": []}
```

fao89:
I'm running with a JSON validator: https://github.com/pulp/pulp-cli/pull/341/files
and it gives a clue where the error happens
https://github.com/pulp/pulp_ansible/runs/3342659987?check_suite_focus=true#step:11:955
```
Invalid control character at: line 1 column 3442 (char 3441)
```

fao89:
after using the JSON validator I could see it is on pulp_ansible side, so I'm closing this PR

PR #343, opened 2021-08-17
Add href support to resource_option

### LINKED ISSUE #315: For sync there's no remote href option nor help text about what's expected for the remote option
What does pulp-cli expect for `--remote` and where is the option for href?

```
$ pulp file repository sync --help
Usage: pulp file repository sync [OPTIONS]

Options:
  --name TEXT    Name of the repository
  --href TEXT    HREF of the repository
  --remote TEXT
  --help         Show this message and exit.
```

mdellweg:
fixes #315

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -1,5 +1,6 @@
 import gettext
 import json
+import re
 from typing import (
     Any,
     Callable,
@@ -248,6 +249,7 @@ def resource_option(*args: Any, **kwargs: Any) -> Callable[[_FC], _FC]:
     lookup_key: str = kwargs.pop("lookup_key", "name")
     context_table: Dict[str, Type[PulpEntityContext]] = kwargs.pop("context_table")
     capabilities: Optional[List[str]] = kwargs.pop("capabilities", None)
+    href_pattern: Optional[str] = kwargs.pop("href_pattern", None)
 
     def _option_callback(
         ctx: click.Context, param: click.Parameter, value: Optional[str]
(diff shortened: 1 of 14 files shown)


mdellweg:
This is currently only implemented in "pulp file repository *" commands to showcase.

pulpcore/cli/common/generic.py
```diff
@@ -256,10 +258,35 @@ def _option_callback(
...
             return value
 
-        split_value = value.split(":", maxsplit=2)
-        while len(split_value) < 3:
-            split_value.insert(0, "")
-        plugin, resource_type, identifier = split_value
+        pulp_href: Optional[str] = None
+        entity: Optional[EntityDefinition] = None
+
+        if value.startswith("/"):
```
daviddavis, line 264: I wonder if this shouldn't check for href_pattern instead and then just assume it's a name if it doesn't match:

```
$ pulp file remote create --name /test/ --url https://fixtures.pulpproject.org/file/
$ pulp file repository create --name /test/ --remote /test/
Error: '/test/' is not a valid href for remote.
```

mdellweg: That wouldn't solve the problem where the name actually looks like a valid href.
But you can always use: `$ pulp file repository create --name /test/ --remote ::/test/`

daviddavis (approved):

mdellweg:
@daviddavis I extended the use of this feature to all plugins sync and export{,er}s. You might want to revisit.

daviddavis (approved):

daviddavis:
I had one thought (not necessarily for this PR): what if we stored the href_pattern on the context instead of having to define it everywhere?

mdellweg:
> I had one thought (not necessarily for this PR): what if we stored the href_pattern on the context instead of having to define it everywhere?

I'm happy to move it around (to generic?). But for the more generic options, what should be _the_ Context to store it on?

daviddavis:
Maybe for generic options, the href_pattern must be supplied when calling resource_option?

mdellweg:
I moved it into the base context. It seemd to match best there.

pulpcore/cli/core/export.py
```diff
@@ -16,16 +18,34 @@
 from pulpcore.cli.core.context import PulpExportContext, PulpExporterContext
 
 _ = gettext.gettext
+REPO_HREF_PATTERN = r"^/pulp/api/v3/repositories/(?P<plugin>\w+)/(?P<resource_type>\w+)/"
```
daviddavis, line 21: Why define this hear instead of using `PulpRepositoryContext.HREF_PATTERN`?

daviddavis:
I was thinking that `_option_callback` in `resource_option` would have some logic like:

```python
if href_pattern is None and hasattr(ctx, HREF_PATTERN):
    href_pattern = ctx.HREF_PATTERN
```

pulpcore/cli/core/export.py
mdellweg: forgot about that.

mdellweg:
> I was thinking that `_option_callback` in `resource_option` would have some logic like:
> 
> ```python
> if href_pattern is None and hasattr(ctx, HREF_PATTERN):
>     href_pattern = ctx.HREF_PATTERN
> ```

The problem is, it needs the pattern before determining the Context class to instanciate.

daviddavis:
Ah, I see. That kind of defeats part of the purpose of my suggestion to have it on context then.

mdellweg:
> Ah, I see. That kind of defeats part of the purpose of my suggestion to have it on context then.

It still feels like the right place.

daviddavis (approved):

PR #345, opened 2021-08-19
Add option that takes name or href

daviddavis:
This [PR](https://github.com/pulp/pulp-cli/pull/343) got me thinking: what if we had a lookup option that took name or href?

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -146,6 +146,19 @@ def _name_callback(
     return value
 
 
+def _ref_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[str]
+) -> Optional[str]:
+    if value is not None:
+        entity_ctx = ctx.find_object(PulpEntityContext)
+        assert entity_ctx is not None
+        if value.startswith("/"):
+            entity_ctx.pulp_href = value
+        else:
+            entity_ctx.entity = {"name": value}
+    return value
+
+
 def _repository_href_callback(
     ctx: click.Context, param: click.Parameter, value: Optional[str]
 ) -> Optional[str]:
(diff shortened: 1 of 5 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -145,6 +145,19 @@ def _name_callback(
...
 
 
+def _id_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[str]
+) -> Optional[str]:
+    if value is not None:
+        entity_ctx = ctx.find_object(PulpEntityContext)
+        assert entity_ctx is not None
+        if value.startswith("/"):
+            entity_ctx.pulp_href = value
```
daviddavis, line 155: Originally I started by implementing an `id_option` that could take an `href_pattern` but that maybe seemed like overkill. I'm tempted to do it but I thought I'd start with something simple first.

pulpcore/cli/common/generic.py
```diff
@@ -145,6 +145,19 @@ def _name_callback(
...
+        else:
+            entity_ctx.entity = {"name": value}
```
daviddavis, line 157: If I got the `id_option` route, I could make lookup_key a possible option.

pulpcore/cli/common/generic.py
```diff
@@ -344,6 +357,13 @@ def _multi_option_callback(
     expose_value=False,
 )
 
+id_option = pulp_option(
+    "--id",
```
mdellweg, line 361: This is basically "name-or-href" now. At that point we could almost make it a positional argument. (Do you sense that i am not fond of the term "id" here? Don't think it is very instructive to users.)
But, what it you need two data to identify one entity? i.e. digest, rel-path...

mdellweg: That is: `pulp file repository sync <name_or_href>`.

daviddavis: Yea, I'm not fond of the option name either but it seems necessarily ambiguous/generic to capture both name and href. It's also short.

I'm a bit mixed about introducing the concept of positional arguments. More UI patterns introduce more mental tax.

I'll think about this feature some more.

mdellweg:
When we have this "ref" option, should we deprecate "--name" and "--href"?

daviddavis:
Filed issue: https://github.com/pulp/pulp-cli/issues/383

PR #354, opened 2021-08-31
Reorganize Contribution docs

melcorr:
[noissue]

As part of Pulpcore 3.15 docs day.
* Move Pulp CLI architecture to contributor's section
because it is only relevant for code contributors
* Create docs contribution section
* Add line to explain where to contribute


docs/contributing.md
```diff
@@ -1,9 +1,128 @@
...
+```
+
+The named tuple `PluginRequirement` is used to describe dependence on server components.
+It needs the `name` of the plugin and accepts optionally an inclusive `min` version,
+an exclusive `max` version, and an `inverted` flag for exclusion of a version range.
+
+Additionally, the `PulpOption` provides the `needs_plugins` keyword argument.
+It accepts a list of `PluginRequirements` to error when used.
+
+## Generics
```
mdellweg, line 100:
```suggestion
### Generics
```

docs/contributing.md
```diff
@@ -1,9 +1,128 @@
...
+
+filter_params = [
+    click.option("--name"),
+    click.option("--name-contains", "name__contains"),
+]
+my_command.add_command(list_command(decorators=filter_params))
+```
+
+
 ## Code conventions
```
mdellweg, line 126: I would put the code (styling) conventions before the architecture.

docs/contributing.md
```diff
@@ -26,7 +145,7 @@ Those include:
   - `PulpEntityContext.pulp_href`: This property can be used to specify an entity by its URI.
     It will be fetched from the server only at read access.
 
-### Compatibility to pulp versions
+## Compatibility to pulp versions
```
mdellweg, line 148: This one should probably be merged with "Version dependent codepaths". This one is the goal, the other one is the tooling for it.
We should have them adjacent at least.

docs/contributing.md
```diff
@@ -39,7 +158,7 @@ Tests are shell scripts in `tests/scripts` with names like `test_*.sh`.
 They should should focus on the cli operation and are not a replacement for pulp integration tests;
 i.e. make sure the cli translates to the right api calls, but do not care about pulp internals.
 
-### Running Tests
+## Running Tests
```
mdellweg, line 161: This makes me think, we should state that we require tests for features in the conventions section (It's better to set the expectations strait upfront.).
Or maybe we add a PR checklist.

docs/contributing.md
melcorr: Fixed!

docs/contributing.md
melcorr: I have tried to do this @mdellweg let me know if you have any suggestions

docs/contributing.md
melcorr: I added a line but I agree with a PR checklist, but perhaps that can be an additional PR that is out of the scope of this rearrangement PR. 
I was wondering if an example test for a feature would be helpful also. 
But perhaps anyone contributing to the CLI would not need such handholding. I always tend to lean towards highest level of handholding we can offer :-)

docs/contributing.md
```diff
@@ -1,15 +1,142 @@
...
+### Version dependent code paths
+
+Each Pulp CLI release is designed to support multiple Pulp server versions and the CLI itself is versioned independently of any version of the Pulp server components.
+It is supposed to be able to communicate with different combinations of server component versions at the same time.
+Because of this, it might be necessary to guard certain features and workarounds by checking against the available server plugin version.
+To divert code paths depending on plugin versions, use the `PulpContext.has_version` function.
+As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects to provide a consistent interface to the command callbacks.
+
+The entity contexts should be used to implement necessary server version-dependent workarounds.
+To facilitate this, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods.
```
mdellweg, line 87:
```suggestion
As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects to provide a consistent interface to the command callbacks.
To facilitate diverting code paths depending on plugin versions, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods.
```

docs/contributing.md
mdellweg: +1 to followup PR's.
Handholding can include that we run shell scripts located in the `tests/scripts` directory.

mdellweg (changes requested):
One last suggestion to amalgamate 4 phrases with too much repetition.

docs/contributing.md
melcorr: Added your rewrite. Thank you!

mdellweg (approved):

PR #364, opened 2021-09-02
Ansible content

gerrod3:
Still need to add tests, but this should fill out the needed content commands for ansible that are not apart of the galaxy api.

### DIFF
--- a/pulpcore/cli/ansible/content.py
+++ b/pulpcore/cli/ansible/content.py
@@ -1,12 +1,27 @@
 import gettext
+from typing import IO, Any, Optional, Union
 
 import click
 
 from pulpcore.cli.ansible.context import PulpAnsibleCollectionVersionContext, PulpAnsibleRoleContext
-from pulpcore.cli.common.context import PulpContext, pass_pulp_context
-from pulpcore.cli.common.generic import list_command
+from pulpcore.cli.common.context import PulpContext, pass_entity_context, pass_pulp_context
+from pulpcore.cli.common.generic import (
+    GroupOption,
+    chunk_size_option,
+    href_option,
+    list_command,
+    pulp_option,
+    show_command,
+)
+from pulpcore.cli.core.context import PulpArtifactContext
 
 _ = gettext.gettext
+
+
+def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
+    if value:
+        ctx.obj.entity = value  # The context should be already set based on the type option
+    return value
 
 
 @click.group()
@@ -28,4 +43,96 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
         raise NotImplementedError()
 
 
-content.add_command(list_command())
+list_options = [
+    pulp_option("--name", help=_("Name of {entity}")),
+    pulp_option("--namespace", help=_("Namespace of {entity}")),
+    pulp_option("--version", help=_("Version of {entity}")),
+    pulp_option(
+        "--only-highest",
+        "is_highest",
+        is_flag=True,
+        default=None,
+        help=_(
+            "Only show highest version of collection version (only works for collection versions)"
+        ),
+    ),
+    pulp_option(
+        "--tags",
+        help=_(
+            "Comma separated list of tags that must all match (only works for collection-versions)"
+        ),
+    ),
+]
+
+fields_options = [
+    pulp_option("--fields", help=_("String list of fields to include in the result")),
+    pulp_option(
+        "--exclude-fields",
+        default="files,manifest,docs_blob",
+        help=_("String list of fields to exclude from result"),
+    ),
+]
+
+lookup_options = [
+    click.option(
+        "--name",
+        help=_("Name of {entity}"),
+        group=["namespace", "version"],
+        expose_value=False,
+        cls=GroupOption,
+        callback=_content_callback,
+    ),
+    click.option(
+        "--namespace",
+        help=_("Namespace of {entity}"),
+        group=["name", "version"],
+        expose_value=False,
+        cls=GroupOption,
+    ),
+    click.option(
+        "--version",
+        help=_("Version of {entity}"),
+        group=["namespace", "name"],
+        expose_value=False,
+        cls=GroupOption,
+    ),
+    href_option,
+]
+
+content.add_command(list_command(decorators=list_options + fields_options))
+content.add_command(show_command(decorators=lookup_options))
+
+
+@content.command()
+@click.option("--file", type=click.File("rb"), required=True)
+@chunk_size_option
+@pulp_option("--name", help=_("Name of {entity}, only required for Roles"))
+@pulp_option("--namespace", help=_("Namespace of {entity}, only required for Roles"))
+@pulp_option("--version", help=_("Version of {entity}, only required for Roles"))
+@pass_entity_context
+@pass_pulp_context
+def upload(
+    pulp_ctx: PulpContext,
+    content_ctx: Union[PulpAnsibleRoleContext, PulpAnsibleCollectionVersionContext],
+    file: IO[bytes],
+    chunk_size: int,
+    name: Optional[str],
+    namespace: Optional[str],
+    version: Optional[str],
+) -> None:
+
+    if isinstance(content_ctx, PulpAnsibleRoleContext):
+        # Check that name, namespace, and version are supplied for Role upload
+        if not all((name, namespace, version)):
+            raise click.ClickException(
+                "Please specify name, namespace, and version when uploading a Role"
+            )
+        artifact_href = PulpArtifactContext(pulp_ctx).upload(file, chunk_size)
+        body = {"artifact": artifact_href, "name": name, "namespace": namespace, "version": version}
+        result = content_ctx.create(body=body)
+        pulp_ctx.output_result(result)
+    else:
+        # Collection upload uses name, namespace, and version for extra validation
+        body = {"expected_name": name, "expected_namespace": namespace, "expected_version": version}
+        result = content_ctx.upload(file=file, body=body)
+        pulp_ctx.output_result(result)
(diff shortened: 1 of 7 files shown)


pulpcore/cli/ansible/content.py
```diff
@@ -27,5 +35,60 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
...
+]
+
+content.add_command(list_command(decorators=list_options + fields_options))
+content.add_command(show_command(decorators=lookup_options))
+
 
-content.add_command(list_command())
+@content.command()
+@click.option("--file", type=click.File("rb"), required=True)
+@click.option(
```
mdellweg, line 64: Don't we have a generic for this?

mdellweg (commented):
No proper review...

pulpcore/cli/ansible/content.py
mdellweg: I mean the chunk-size

gerrod3: Yes, I didn't rebase when making this so copying pasting made me use the old option.

pulpcore/cli/ansible/content.py
```diff
@@ -1,14 +1,29 @@
...
+    show_command,
+)
+from pulpcore.cli.core.context import PulpArtifactContext
 
 _ = gettext.gettext
 
 
+def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
+    if value:
+        ctx.obj.entity = value  # The context should be already set based on the type option
```
mdellweg, line 23: I think, this should be a `ctx.find_object(PulpContentContext)`.

pulpcore/cli/ansible/content.py
```diff
@@ -28,4 +50,96 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
         raise NotImplementedError()
 
 
-content.add_command(list_command())
+list_options = [
```
mdellweg, line 53: These are really filter options, right? We could have gems like `--name-contains`...

pulpcore/cli/ansible/content.py
```diff
@@ -28,4 +50,96 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
...
+    pulp_option("--name", help=_("Name of {entity}")),
+    pulp_option("--namespace", help=_("Namespace of {entity}")),
+    pulp_option("--version", help=_("Version of {entity}")),
```
mdellweg, line 56: Does pulp support version sorting filters? Like `--version-gt'?

pulpcore/cli/ansible/content.py
```diff
@@ -28,4 +50,96 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
...
+    pulp_option(
+        "--only-highest",
```
mdellweg, line 58: Do you think, `--highest` or `--latest` would be a better name?

pulpcore/cli/ansible/content.py
```diff
@@ -28,4 +50,96 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
...
+        is_flag=True,
+        default=None,
+        help=_(
+            "Only show highest version of collection version (only works for collection versions)"
+        ),
+    ),
+    pulp_option(
+        "--tags",
+        help=_(
+            "Comma separated list of tags that must all match (only works for collection-versions)"
```
mdellweg, line 69:
```suggestion
            "Comma separated list of tags that must all match (only works for collection versions)"
```

pulpcore/cli/ansible/content.py
```diff
@@ -28,4 +50,96 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
...
+        if not all((name, namespace, version)):
+            raise click.ClickException(
+                "Please specify name, namespace, and version when uploading a Role"
+            )
+        artifact_href = PulpArtifactContext(pulp_ctx).upload(file, chunk_size)
+        body = {"artifact": artifact_href, "name": name, "namespace": namespace, "version": version}
+        result = content_ctx.create(body=body)
+        pulp_ctx.output_result(result)
+    else:
+        # Collection upload uses name, namespace, and version for extra validation
```
mdellweg, line 142:
```suggestion
        # Collection upload uses name, namespace, and version for server side validation
```

pulpcore/cli/ansible/context.py
```diff
@@ -21,6 +22,11 @@ class PulpAnsibleCollectionVersionContext(PulpContentContext):
     LIST_ID = "content_ansible_collection_versions_list"
     READ_ID = "content_ansible_collection_versions_read"
     CREATE_ID = "content_ansible_collection_versions_create"
+    UPLOAD_ID = "upload_collection"
```
mdellweg, line 25: Wow, this is unexpected. Is this endpoint not part of the usual pulp rest api?
Just curiosity. We can proabably not do any thing about it here.

pulpcore/cli/ansible/repository.py
```diff
@@ -49,6 +56,31 @@
 )
 
 
+CONTENT_LIST_SCHEMA = s.Schema(
+    [{"name": s.And(str, len), "namespace": s.And(str, len), "version": s.And(str, len)}]
+)
+
+
+def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
+    print(value)
```
mdellweg, line 65:
```suggestion
```

pulpcore/cli/ansible/repository.py
```diff
@@ -49,6 +56,31 @@
...
+    if value:
+        ctx.obj.entity = value  # The context is set by the type parameter on the content commands
+    return value
+
+
+def _content_type_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
+    # This is eagerly ran
```
mdellweg, line 72:
```suggestion
    # This needs to run eagerly
```

pulpcore/cli/ansible/repository.py
```diff
@@ -119,6 +229,3 @@ def sync(
         href=repository_href,
         body=body,
     )
-
-
-# TODO Finish 'add' and 'remove' commands when role and collection contexts are implemented
```
mdellweg, line 124: :tada:

pulpcore/cli/common/generic.py
```diff
@@ -109,7 +109,9 @@ def handle_parse_result(
     ) -> Any:
         assert self.name is not None
         all_options = self.group + [self.name]
-        if all(x in opts for x in all_options):
+        options_present = [x for x in all_options if x in opts]
+        num_options = len(options_present)
+        if not num_options and not self.required or num_options == len(all_options):
```
mdellweg, line 114: This is really of hard to read. How about:
```python
if num_options != len(all_options) and (num_options != 0 or self.required):
    raise click.UsageError(...)
self.prompt = None
```

pulpcore/cli/common/generic.py
```diff
@@ -119,7 +121,7 @@ def handle_parse_result(
             )
         value = opts.get(self.name)
         if self.callback is not None:
-            value = self.callback(ctx, self, {o: opts[o] for o in all_options})
+            value = self.callback(ctx, self, {o: opts[o] for o in options_present})
```
mdellweg, line 124: I don't think we should call the callback if `options_present` is empty.

pulpcore/cli/common/generic.py
```diff
@@ -217,11 +219,12 @@ def load_json_callback(
 
 
 def create_content_json_callback(
-    content_ctx: Type[PulpContentContext], schema: s.Schema = None
+    content_ctx: Optional[Type[PulpContentContext]] = None, schema: s.Schema = None
```
mdellweg, line 222: Are we breaking any interface if we properly rename this parameter to `ctx_class`?

pulpcore/cli/ansible/content.py
gerrod3: `--latest` sounds good.

pulpcore/cli/ansible/content.py
gerrod3: Sadly, no

pulpcore/cli/ansible/context.py
gerrod3: There is a create endpoint for collection versions, but it requires you know and supply every piece of metadata for the upload. This other endpoint is apart of the galaxy api and is much simpler to use. I think they both use the same upload task code, it's just their serializers are different.

pulpcore/cli/ansible/content.py
gerrod3: There is only exact filters on these fields.

pulpcore/cli/ansible/context.py
mdellweg: So this endpoint is meant to be used by the galaxy cli? In pulp_container we intentionally excluded the live/push-pull api from the openapi schema...

mdellweg:
```suggestion
    UPLOAD_ID: ClassVar[str] = "upload_collection"
```

pulpcore/cli/common/generic.py
```diff
@@ -217,11 +218,12 @@ def load_json_callback(
 
 
 def create_content_json_callback(
-    content_ctx: Type[PulpContentContext], schema: s.Schema = None
+    ctx_class: Optional[Type[PulpContentContext]] = None, schema: s.Schema = None
 ) -> Any:
     def _callback(
         ctx: click.Context, param: click.Parameter, value: Optional[str]
     ) -> Optional[List[PulpContentContext]]:
+        nonlocal ctx_class
```
mdellweg, line 226: This I don't understand. I think this allows to overwrite the value in the closure (that is the scope created by calling `create_content_json_callback`. And I think, this scope may live longer than we expect here when using the `shell` command. So where I expect this to fail is: Inside `pulp shell` call a content modify command for collections and then one for roles. In the second call, we might have the `CollectionContext` still in the `ctx_class`. For single one shot commands, this is probably just not correct, but no problem.

Issue #284, opened 2021-06-30
As a user I want to be able to create, upload, and list rpm packages


ajsween:
Essentially I'd like to be able to add rpm packages in the same way I can with file content. I'd also like to be able to directly upload the artifact as part of the cli create operation (skip the ```pulp artifact create``` step).

mdellweg:
#205 may fix this.

PR #347, opened 2021-08-24
Add support for ACS

daviddavis:
fixes #324

PR #365, opened 2021-09-07
Add rpm package content support

Tjev:
I've implemented the rpm content functionality as has been once requested (#284), and once attempted (#205) by my colleague.

### DIFF
--- a/pulpcore/cli/rpm/repository.py
+++ b/pulpcore/cli/rpm/repository.py
@@ -2,6 +2,7 @@ import gettext
 from typing import Any, Dict, Iterable, Optional
 
 import click
+import schema as s
 
 from pulpcore.cli.common.context import (
     EntityFieldDefinition,
@@ -15,6 +16,7 @@ from pulpcore.cli.common.context import (
 )
 from pulpcore.cli.common.generic import (
     create_command,
+    create_content_json_callback,
     destroy_command,
     href_option,
     label_command,
@@ -22,6 +24,7 @@ from pulpcore.cli.common.generic import (
     list_command,
     name_option,
     pulp_option,
+    repository_content_command,
     repository_href_option,
     repository_option,
     resource_option,
@@ -32,7 +35,11 @@ from pulpcore.cli.common.generic import (
 )
 from pulpcore.cli.core.generic import task_command
 from pulpcore.cli.rpm.common import CHECKSUM_CHOICES
-from pulpcore.cli.rpm.context import PulpRpmRemoteContext, PulpRpmRepositoryContext
+from pulpcore.cli.rpm.context import (
+    PulpRpmPackageContext,
+    PulpRpmRemoteContext,
+    PulpRpmRepositoryContext,
+)
 
 _ = gettext.gettext
 
@@ -48,6 +55,17 @@ remote_option = resource_option(
         "Remote used for synching in the form '[[<plugin>:]<resource_type>:]<name>' or by href."
     ),
 )
+
+
+def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
+    if value:
+        pulp_ctx = ctx.find_object(PulpContext)
+        assert pulp_ctx is not None
+        ctx.obj = PulpRpmPackageContext(pulp_ctx, pulp_href=value)
+    return value
+
+
+CONTENT_LIST_SCHEMA = s.Schema([{"pulp_href": str}])
 
 
 @click.group()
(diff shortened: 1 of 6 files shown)


gerrod3 (commented):
This is looking pretty good. The CI isn't going to pass without some lint changes, run `make black & make lint`, this should format the files nicely and tell you if you have any mypy issues. The CI also needs a changelog file and reference in the commit message. See this PR for an example: https://github.com/pulp/pulp-cli/pull/347. Finally we like to add tests for new features, something similar to this: https://github.com/pulp/pulp-cli/blob/develop/tests/scripts/pulp_python/test_content.sh#L1-L23 would be good.

The commands & options look good, but I don't know much about the RPM plugin, so I can't advise on what could be changed. Maybe @ggainey or @daviddavis could provide more insight.

daviddavis:
@ggainey would you mind reviewing?

Tjev:
Hello :)

in order to test the newly implemented ```rpm_content``` command group in some reasonable manner, I attempted to implement the ```rpm repository content``` command group.

I tried to follow the implementation from the ```file``` and ```python```, but am still struggling to get the commands to work.
If I attempt to add a newly uploaded rpm package into an existing repository as follows: 

```pulp rpm repository content add --package-href "some_package_href" --repository "some_repo"```

I get following error:

```Error: Parameter [package_href] not available for content_rpm_packages_list.```

Could I ask for some guidance, please?

File: pulpcore/cli/rpm/repository.py [hunk 1]
```diff
@@ -50,6 +57,17 @@
 )
 
 
+def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
+    if value:
+        pulp_ctx = ctx.find_object(PulpContext)
+        assert pulp_ctx is not None
+        ctx.obj = PulpRpmPackageContext(pulp_ctx, entity={"package_href": value})
```
mdellweg, line 64: This may solve your problem here.

```suggestion
        ctx.obj = PulpRpmPackageContext(pulp_ctx, pulp_href=value})
```

But it's probably not meant that way: This callback was meant to lookup the content by its "natural key" (is it nevra here?), while you created an href-lookup, which is another valid approach currently missing on the file commands.

Tjev:
I managed to resolve the issue. Thanks for the help.

Could I ask for a review, please?

pulpcore/cli/rpm/repository.py
```diff
@@ -67,8 +85,39 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
         raise NotImplementedError()
 
 
+package_options = [
+    click.option(
+        "--pulp-href",
```
mdellweg, line 90:
```suggestion
        "--package-href",
```

pulpcore/cli/rpm/repository.py
```diff
@@ -50,6 +57,17 @@
...
 
+def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
+    if value:
+        pulp_ctx = ctx.find_object(PulpContext)
+        assert pulp_ctx is not None
+        ctx.obj = PulpRpmPackageContext(pulp_ctx, pulp_href=value)
+    return value
+
+
+CONTENT_LIST_SCHEMA = s.Schema([{"pulp_href": str}])
```
mdellweg, line 68: I still think we should see a natural key here.

mdellweg:
Can you squash to one commit and add a "fixes #284" to the commit message?

pulpcore/cli/rpm/repository.py
ggainey: "natural key" for RPMs gets complicated in the wild - you would think it would just 'nevra in repo', but you can have RPMs with the same NEVRA but different checksums, and RPMs with the same NEVRA/checksum but different relative_paths. This is because the "rules" for rpm-repo-layout are...more like a set of **guidelines**, really.

I would eventually like to see a more 'natural' approach available to CLI users - maybe an "add to repo" that takes a JSON file of just path-to-rpm-file and relative-path? - that hides things like sha256 and hrefs and the content-modify call details from the casual CLI user. However, the approach here is def. valid, and not having this API covered is absolutely a Gaping Hole in rpm-support in the CLI - so I'm happy to have it! We can add "higher level" commands at a later date.

pulpcore/cli/rpm/repository.py
Tjev: I've changed it back to the `--package-href` as you suggested. However, I can't seem to get the `package_href` working for the json content. For now, I kept the `pulp_href` as the option keyword for the json. Is this way okay?

mdellweg:
Your test shell scripts fail the shellcheck linting.
You know you can run all the lingers locally: `make lint` or even `make black lint`.

tests/scripts/pulp_rpm/test_content.sh
```diff
@@ -0,0 +1,55 @@
...
+expect_succ pulp rpm remote show --name "cli_test_rpm_remote"
+REMOTE_HREF=$(echo "$OUTPUT" | jq -r .pulp_href)
+expect_succ pulp rpm repository create --name "cli_test_rpm_repository" --remote "$REMOTE_HREF"
+expect_succ pulp rpm repository show --name "cli_test_rpm_repository"
+
+expect_succ pulp rpm repository content modify \
+--repository "cli_test_rpm_repository" \
+--add-content "[{\"pulp_href\": \"$PACKAGE_HREF\"}]"
+expect_succ pulp rpm repository content list --repository "cli_test_rpm_repository"
+test "(echo \"$OUTPUT\" | jq -r '.[0].pulp_href')" = "$PACKAGE_HREF"
```
mdellweg, line 31:
```suggestion
test "$(echo "$OUTPUT" | jq -r '.[0].pulp_href')" = "$PACKAGE_HREF"
```

mdellweg:
@ggainey you want to give this a final review?

ggainey:
> @ggainey you want to give this a final review?

Installed into my dev-env and played with. Went back over the changes, and the conversations, and I'm good to go with this. As discussed, we will probably want to add some higher-level/more "natural" command-options to make things easier on users, but this is great to have!

ggainey (approved):
This is a great addition, thank you for all the work!

PR #377, opened 2021-09-22
Add pulp file ACS refresh command

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -89,6 +89,7 @@ class PulpContext:
         self.format: str = format
         self.background_tasks: bool = background_tasks
         self.timeout: int = timeout
+        self.start_time: Optional[datetime.datetime] = None
 
     @property
     def api(self) -> OpenAPI:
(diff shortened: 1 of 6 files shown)


pulpcore/cli/common/context.py
```diff
@@ -148,41 +148,97 @@ def call(self, operation_id: str, non_blocking: bool = False, *args: Any, **kwar
...
-                elif task["state"] == "canceled":
-                    raise click.ClickException("Task canceled")
-                elif task["state"] in ["waiting", "running", "canceling"]:
+                if task_group["all_tasks_dispatched"] is True:
+                    for task in task_group["tasks"]:
+                        task = self.api.call(
+                            "tasks_read", parameters={"task_href": task["pulp_href"]}
+                        )
+                        click.echo(f"Waiting for task {task['pulp_href']}", err=True)
+                        self._poll_task(task, timeout)
```
mdellweg, line 226: This is not accounting the tasks timeout as part of the whole timeout. Maybe counting seconds is not the best approach and we should be starting to compare wall clock times.

pulpcore/cli/core/task.py
```diff
@@ -107,6 +108,6 @@ def cancel(
         try:
             pulp_ctx.wait_for_task(entity)
         except Exception as e:
-            if str(e) != "Task canceled":
+            if not re.match("Task /pulp/api/v3/tasks/[-0-9a-f]*/ canceled", str(e)):
```
mdellweg, line 111: +1

pulpcore/cli/file/context.py
```diff
@@ -23,6 +24,13 @@ class PulpFileACSContext(PulpEntityContext):
     CREATE_ID = "acs_file_file_create"
     UPDATE_ID = "acs_file_file_partial_update"
     DELETE_ID = "acs_file_file_delete"
+    REFRESH_ID = "acs_file_file_refresh"
```
mdellweg, line 27:
```suggestion
    REFRESH_ID: ClassVar[str] = "acs_file_file_refresh"
```

mdellweg (commented):
I'd like to have one if the ACS experts to judge the necessary test coverage here.

pulpcore/cli/common/context.py
daviddavis: One question I had is whether timeout should apply to each task or the entire task group?

pulpcore/cli/common/context.py
```diff
@@ -148,41 +148,97 @@ def call(self, operation_id: str, non_blocking: bool = False, *args: Any, **kwar
             click.echo(f"Started background task {task_href}", err=True)
             if not non_blocking:
                 result = self.wait_for_task(result)
+        if isinstance(result, dict) and ["task_group"] == list(result.keys()):
```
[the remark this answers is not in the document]
mdellweg: No action needed here.
Thinking further: If we could pass this as some kind of annotation to the result without breaking the `call` interface backwards incompatible ...

pulpcore/cli/common/context.py
mdellweg: That is indeed a valid question. I do not know.

mdellweg: I agree completely. In the best case we can shield the user from thinking about tasks. So it's one operation that takes time.

daviddavis:
@pavelpicka can you review?

pavelpicka (approved):
LGTM
just "[cli-bad-repo](https://github.com/pulp/pulp-cli/pull/377/files#diff-c3b3f2f409d8e4fcecd980edb6314e598fd501fe9cc3c3bdf2bf6b1ad35e5c05R14)" in tests invokes malformed repo, this case sounds like incomplete or metadata only (if we expect success)

daviddavis:
I renamed the repo `cli-repo-manifest-only` and added a failing refresh test.

mdellweg (approved):

mdellweg:
Are the necessary bits for this merged to pulpcore and pulp_file or do we need to wait with merging to not break their respective ci?

daviddavis:
> Are the necessary bits for this merged to pulpcore and pulp_file or do we need to wait with merging to not break their respective ci?

The only thing this relies on is the pulp_file refresh command which got merged: https://github.com/pulp/pulp_file/pull/536

PR #308, opened 2021-07-16
Add validation to file content json parameters

mdellweg:
fixes #255

PR #381, opened 2021-09-30
Bump pulpcore version requirement to >= 3.11

mdellweg:
Require pulpcore>=3.11 and drop all workarounds for older versions.
Transitive plugin requirements:
  pulp_container>=2.3

fixes #308

### DIFF
--- a/CHANGES/380.removal
+++ b/CHANGES/380.removal
@@ -1,0 +1,1 @@
+Bumped required pulpcore version to >=3.11 and removed legacy workarounds.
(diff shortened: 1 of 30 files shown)


mdellweg:
Line count statistics: +24 −221

ggainey (approved):
Tested against dev-env and a 3.7 container, looks good. The 3.9/nightly failure is "Object of type PythonRepository is not JSON serializable" , which needs to be fixed in pulp-python I think.

PR #384, opened 2021-10-03
Add debug openapi commands

mdellweg:


pulpcore/cli/common/debug.py
```diff
@@ -43,3 +51,48 @@ def task_summary(pulp_ctx: PulpContext) -> None:
...
+def schema(pulp_ctx: PulpContext, jq_expr: Optional[str]) -> None:
+    """
+    Print the openapi schema of the server.
+    """
+    result = pulp_ctx.api.api_spec
+    if jq_expr:
+        if jq is None:
+            raise click.ClickException(_("`jq` is not installed."))
+        else:
+            result = jq.compile(jq_expr).input(result).first()
```
mdellweg, line 73: I just now realized, we could make this a global option to transform any old (json/yaml based) output. WDYT?

ggainey: I like this idea. I'd divorce it from this PR - open an issue and expand on this once it's merged maybe?

ggainey (commented):
I like this a lot!

The older tests are failing because 'canceling' wasn't added to pulpcore until 0e1c62, but pulp-cli "knows about it". What can we do about that?

mdellweg:
@ggainey removed the jq idea for now.
Ci failure is reduced to pulp_ansible bug: https://github.com/pulp/pulp_ansible/pull/659

pulpcore/cli/common/debug.py
```diff
@@ -37,9 +40,50 @@ def task_summary(pulp_ctx: PulpContext) -> None:
     """
     from pulpcore.cli.core.context import PulpTaskContext
 
+    TASK_STATES = ["waiting", "skipped", "running", "completed", "failed", "canceled"]
+    if pulp_ctx.has_plugin(PluginRequirement("core", min="3.14")):
+        TASK_STATES.append("canceling")
     result = {}
-    for state in ["waiting", "skipped", "running", "completed", "failed", "canceled", "canceling"]:
+    for state in TASK_STATES:
```
ggainey, line 47: :+1:

ggainey (approved):
lg2m!

ggainey:
> @ggainey removed the jq idea for now. Ci failure is reduced to pulp_ansible bug: [pulp/pulp_ansible#659](https://github.com/pulp/pulp_ansible/pull/659)

Oh - I didn't mean to ask that the jq-work be removed from here, so much as "don't do the global-option as part of this PR". But it's prob better to fully-separate the two.

Looks good - do we wait for the ansible PR to merge and a clean run before merging here?

mdellweg:
I don't see a reason to wait.

PR #386, opened 2021-10-05
Add jq processing to output

mdellweg:
[noissue]


mdellweg:
The benefit of this is negligible. Closing!

PR #390, opened 2021-10-12
Update distribution update calls.

ipanova:
[noissue]

### DIFF
--- a/pulpcore/cli/container/context.py
+++ b/pulpcore/cli/container/context.py
@@ -32,6 +32,7 @@ class PulpContainerDistributionContext(PulpEntityContext):
     CREATE_ID = "distributions_container_container_create"
     UPDATE_ID = "distributions_container_container_partial_update"
     DELETE_ID = "distributions_container_container_delete"
+    NULLABLES = {"repository_version", "repository"}
 
     def preprocess_body(self, body: EntityDefinition) -> EntityDefinition:
         body = super().preprocess_body(body)
--- a/tests/scripts/pulp_container/test_distribution.sh
+++ b/tests/scripts/pulp_container/test_distribution.sh
@@ -25,10 +25,14 @@ expect_succ pulp container distribution create --name "cli_test_container_distro
 
 expect_succ pulp container distribution list
 expect_succ pulp container distribution show --name "cli_test_container_distro"
-# Not sure what's going on here. Maybe a bug in the distribution serializer.
-# expect_succ pulp container distribution update --name "cli_test_container_distro" --version 0
-# expect_succ pulp container distribution update --name "cli_test_container_distro" --repository ""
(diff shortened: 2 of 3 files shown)


mdellweg:
It would be interesting, if we could still get these tests to pass.
Could you rebase please?

ipanova:
@mdellweg it's in draft because we first need to fix this in pulp-container, since the unsafe async flag was removed from content app https://pulp.plan.io/issues/9454#note-1

mdellweg:
@ipanova Looks like the upstream issue is fixed. Are we ready to move forward with this?

pulpcore/cli/container/context.py
```diff
@@ -32,13 +32,17 @@ class PulpContainerDistributionContext(PulpEntityContext):
...
 
     def preprocess_body(self, body: EntityDefinition) -> EntityDefinition:
         body = super().preprocess_body(body)
         version = body.pop("version", None)
         if version is not None:
             repository_href = body.pop("repository")
             body["repository_version"] = f"{repository_href}versions/{version}/"
+        if not self.pulp_ctx.has_plugin(PluginRequirement("container", min="2.10.0.dev")):
+            if body.get("repository_version") == "":
+                del body["repository_version"]
```
mdellweg, line 45: Now that i think of this, at this point, line 38 would have transformed the `""` into `None` already, since you specified `"repository_version"` in `NULLABLES`.

tests/scripts/pulp_container/test_distribution.sh
```diff
@@ -25,10 +25,12 @@ expect_succ pulp container distribution create --name "cli_test_container_distro
 
 expect_succ pulp container distribution list
 expect_succ pulp container distribution show --name "cli_test_container_distro"
-# Not sure what's going on here. Maybe a bug in the distribution serializer.
-# expect_succ pulp container distribution update --name "cli_test_container_distro" --version 0
-# expect_succ pulp container distribution update --name "cli_test_container_distro" --repository ""
-# expect_succ pulp container distribution update --name "cli_test_container_distro_ver" --repository "cli_test_container_repository"
-# expect_succ pulp container distribution update --name "cli_test_container_distro_ver" --repository ""
 
+if pulp debug has-plugin --name "container" --min-version "2.10.0.dev"
```
mdellweg, line 29: So we accept that this is a bug we could not work around in earlier versions? :+1:
Optional: Leave a link to the issue in `pulp_container` here.

mdellweg (approved):

tests/scripts/pulp_container/test_distribution.sh
ipanova: yeah, unfortunately. There was a fix in core *and* plugin, so this combination basically makes it impossible to make the tests pass against earlier versions :/

PR #391, opened 2021-10-12
WIP Fix CI failure from distribution validation

mdellweg:
- [x] adjust the option help texts to reflect the they replace the other one


gerrod3:
> Hmm, looks like we need some version guards (at least for python) for versions where repository is not an allowed parameter for distributions. cc. @gerrod3 Was this introduced with autopublish?

Yes, before 3.13, auto-distribute worked by having both publication and repository set. Two plugins (file and rpm) released with this version of auto-distribute. When we switch to only needing repository to be set in 3.13 for compatibility we only put the strict guards in the plugin's serializers else the already released file and rpm wouldn't be compatible with 3.13.

mdellweg:
This may fix the tests now, but i am not sure it's a nice solution.

PR #392, opened 2021-10-13
Disable ACS tests for RPM plugin temporarily

dralley:
RPM plugin is backing out the patches so that the 3.16 release can be
pulpcore 3.15 compatible.  It can be re-enabled immediately after the
release gets branched.

[noissue]


dralley:
@mdellweg How is this?

mdellweg:
> @mdellweg How is this?

It depends on the pulp_rpm version this feature is supposed to land.

PR #393, opened 2021-10-13
Make log capturing a session fixture

mdellweg:
[noissue]


mdellweg:
This should be an improvement to the log capturing, as is does not create a new subprocess for each individual test.

Issue #398, opened 2021-10-14
Add orphan cleanup options to command


gerrod3:
## Summary
orphan cleanup has two options that can be passed in: 
`content_hrefs` - a list of content hrefs to choose to cleanup from if they are orphans
`orphan_protection_time` - from pulpcore 3.15, amount of protection time before removing any orphaned content

## Examples
`pulp orphan cleanup --protection-time 0`
`pulp orphan cleanup --content-hrefs '["/content_href/1/", "/content_href/2"]'`

PR #399, opened 2021-10-14
Add new options to orphan clenaup command

gerrod3:
fixes: #398


pulpcore/cli/core/orphan.py
```diff
@@ -16,13 +18,39 @@ def orphan() -> None:
...
+    body: Dict[str, Any] = {}
+    if content_hrefs is not None:
+        pulp_ctx.needs_plugin(
+            PluginRequirement("core", "3.14", feature=_("the content_hrefs option"))
+        )
+        body["content_hrefs"] = content_hrefs
+    if protection_time is not None:
+        pulp_ctx.needs_plugin(
+            PluginRequirement("core", "3.15", feature=_("the protection_time option"))
+        )
```
mdellweg, line 50: I think this can be more elegant when using a pulp_option.

gerrod3: I tried that, it needed an entity context object. Should I make one for the command?

mdellweg: I see. Then it's probably not worth it.

Is the entity context only used for the help text? We can probably opt out of that...

pulpcore/cli/common/generic.py
```diff
@@ -88,8 +88,8 @@ def get_help_record(self, ctx: click.Context) -> Optional[Tuple[str, str]]:
             return None
         synopsis, help_text = tmp
         entity_ctx = ctx.find_object(PulpEntityContext)
-        assert entity_ctx is not None
-        help_text = help_text.format(entity=entity_ctx.ENTITY, entities=entity_ctx.ENTITIES)
+        if entity_ctx is not None:
+            help_text = help_text.format(entity=entity_ctx.ENTITY, entities=entity_ctx.ENTITIES)
```
mdellweg, line 92: :+1:

tests/scripts/pulpcore/test_orphans.sh
```diff
@@ -6,3 +6,40 @@
 expect_succ pulp orphan cleanup
 
 test "$(echo "${OUTPUT}" | jq -r '.state' )" = "completed"
+
+pulp debug has-plugin --name "file" || exit 3
+
+if pulp debug has-plugin --name "core" --max-version "3.13.0"
```
mdellweg, line 12:
```suggestion
if ! pulp debug has-plugin --name "core" --min-version "3.14.0"
```
We would handle 3.13.1 wrong...

tests/scripts/pulpcore/test_orphans.sh
```diff
@@ -6,3 +6,40 @@
...
+  expect_succ pulp file content upload --file test_3.txt --relative-path orphan_test/test_3.txt
+
+  content_href="$(echo "$OUTPUT" | jq -r .pulp_href)"
+  expect_succ pulp file content list
+  count="$(echo "$OUTPUT" | jq -r length)"
+  expect_succ pulp orphan cleanup --content-hrefs "[\"$content_href\"]"
+  expect_succ pulp file content list
+  if pulp debug has-plugin --name "core" --min-version "3.15.0"
+  then
+    test "$(echo "$OUTPUT" | jq -r length)" -eq "$count"
```
mdellweg, line 31: Wouldn't this branch be a bug in pulpcore 3.14?

tests/scripts/pulpcore/test_orphans.sh
```diff
@@ -6,3 +6,40 @@
...
+  else
+    test "$(echo "$OUTPUT" | jq -r length)" -eq "$((count-1))"
+    expect_succ pulp orphan cleanup
+  fi
+fi
+
+if pulp debug has-plugin --name "core" --max-version "3.14.0"
```
mdellweg, line 38:
```suggestion
if ! pulp debug has-plugin --name "core" --min-version "3.15.0"
```

tests/scripts/pulpcore/test_orphans.sh
```diff
@@ -6,3 +6,40 @@
 expect_succ pulp orphan cleanup
 
 test "$(echo "${OUTPUT}" | jq -r '.state' )" = "completed"
+
+pulp debug has-plugin --name "file" || exit 3
+
+if [ "$(pulp debug has-plugin --name "core" --min-version "3.14.0")" = "true" ]
```
mdellweg, line 12:
```suggestion
if pulp debug has-plugin --name "core" --min-version "3.14.0"
```
This command is meant to be used as a condition.

tests/scripts/pulpcore/test_orphans.sh
gerrod3: Yeah you're right, I had to find the file under .ci/settings/settings.py, but we do set protection time to 0 for the tests. Changed the tests to accommodate for that.

PR #400, opened 2021-10-15
Add string-or-file option.

### LINKED ISSUE #220: As a user I can specify a file path for cert options such as --ca-cert for `pulp rpm remote create`

ggainey:
json_option is now built on top of this.

closes #220.

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -1,19 +1,7 @@
 import gettext
 import json
 import re
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    Iterable,
-    List,
-    Mapping,
-    Optional,
-    Tuple,
-    Type,
-    TypeVar,
-    Union,
-)
+from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Tuple, Type, TypeVar
 
 import click
 import schema as s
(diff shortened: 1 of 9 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -183,30 +171,46 @@ def _version_callback(
...
+
+def load_json_callback(
+    ctx: click.Context,
+    param: click.Parameter,
+    value: Optional[str],
+) -> Any:
+    """Load JSON from input string or from file if string starts with @."""
+    json_object: Any
+    json_string: str
+    json_string = load_file_or_string_callback(ctx, param, value)
```
mdellweg, line 211:
```suggestion
    json_string: str = load_file_or_string_callback(ctx, param, value)
```

pulpcore/cli/common/generic.py
```diff
@@ -183,30 +171,46 @@ def _version_callback(
...
 
 
-def load_json_callback(
+# TODO: would be great to have enable this to take a validator, rather than having
+# to build "on top of" it like I'm doing now w/ json_callback
+def load_file_or_string_callback(
     ctx: click.Context,
     param: click.Parameter,
     value: Optional[str],
 ) -> Any:
```
mdellweg, line 180: This should be Optional[str] now.

pulpcore/cli/common/generic.py
```diff
@@ -183,30 +171,46 @@ def _version_callback(
...
+    ctx: click.Context,
+    param: click.Parameter,
+    value: Optional[str],
+) -> Any:
+    """Load JSON from input string or from file if string starts with @."""
+    json_object: Any
+    json_string: str
+    json_string = load_file_or_string_callback(ctx, param, value)
+    if not json_string:
+        return value
```
mdellweg, line 213: `""` would evaluate as proper json, so only `None` is really interesting and we can catch that before calling the other hook.

test_requirements.txt
```diff
@@ -5,7 +5,7 @@ isort
 mypy
 pygments
 pytest
-
+shellcheck-py
```
mdellweg, line 8: Ha, haven't known that. I always thought it was not a python related tool.

tests/scripts/pulp_file/mock.crt
```diff
@@ -0,0 +1,3 @@
+-----BEGIN CERTIFICATE-----
+This file is vaguely cert-like.==
```
mdellweg, line 2: The certificate duck! :duck:

mdellweg (commented):
I like this very much!
Idea for the future: Do we need an almighty callback class, that can naturally chain multiple transformers, validators and presenters together?

tests/scripts/pulp_file/mock.crt
ggainey: indeed :)

PR #402, opened 2021-10-18
Have dependabot use '[noissue]'

daviddavis:
[noissue]


daviddavis:
> BTW You noticed, we changed the default branch name?

I did 👍  👍  👍

Issue #395, opened 2021-10-13
a way to know what plugins work in which version of pulp-cli.


fkuep:
### Excuse
Thanks for all the work!   
I am very blunt here and I hope You can take it as my despair.  
### reference
#394 ,  https://github.com/pulp/pulpproject.org/issues/361 
## Summary
As a user I don't know what plugins exsist and if they work together.  
I don't know if there is a much cleverer way, but I will just start to suggest:

* You take issues for plugin-writers to "register" their pypi links.
Then You could:
  * have a ci script install those into a current version pulp-cli
  * execute some sort of "list-commands"
  * Make the Known plugins table, a "known to work with this version of pulp-cli "-table.

My Problem as user, right now :   
*I am not even convinced I don't end up with multiple venvs of cli to support Your release process. Not something I intend to deal with as a user. *

PR #404, opened 2021-10-19
Add workflows documentation page

gerrod3:
fixes: #395

### DIFF
--- a/docs/supported_workflows.md
+++ b/docs/supported_workflows.md
@@ -1,0 +1,160 @@
+# Supported Workflows
+
+`pulp-cli` is still in beta, so the features and workflows listed here are subject to change.
+Currently `pulp-cli` supports 5 versions of `pulpcore` from 3.11-3.16 and 5 of Pulp's plugins:
+`pulp_ansible`, `pulp_container`, `pulp_file`, `pulp_python` and `pulp_rpm`. Some of Pulp's other
+plugins can be added to the CLI through CLI plugins, check out [CLI plugins](index) for more information.
+
+## pulpcore
+
+Pulpcore commands require a minimum version of 3.11. Not every command is supported for all
+`pulpcore` versions. Run `pulp status` to check your connection to Pulp and see currently 
+installed plugins.
+
+### Workflows
+
+The CLI currently supports the following workflows:
+
+* [**Uploading** Artifacts](https://docs.pulpproject.org/pulpcore/workflows/upload-publish.html)
+* [**Creating, Syncing, Distributing** Content](https://docs.pulpproject.org/pulpcore/workflows/exposing-content.html)
+* [**Exporting** Content](https://docs.pulpproject.org/pulpcore/workflows/import-export.html#exporting)
+* [**Labeling** Resources](https://docs.pulpproject.org/pulpcore/workflows/labels.html)
+
+### Features
+
+The CLI currently supports the following operations on these `pulpcore` objects
+(C = Create, R = Read, U = Update, D = Delete):
+
+* Access Policies - **RU**
+* Artifacts - **CR**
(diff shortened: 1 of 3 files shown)


mdellweg:
Actually, i am concerned this level of detail may be too much (both for digesting and for keeping up to date).
What i really like about this is that we list all the workflow links for plugin specific docs, as well as the "use `--help` to find yourself around". The latter has the benefit that it is always accurate to the users installation.

docs/supported_workflows.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+## pulp_ansible
+
+Ansible commands require minimum version of 0.7.0. The commands can be found under the `ansible`
+subgroup.
+
+### Workflows
+
+The CLI currently supports the following workflows:
```
ggainey, line 44: Same here as above - in fact, all the Workflows sections. I think they need a blank line before the lists?

ggainey (changes requested):
Would still like to see the CRUD indicators in bold, as well as fixing the Workflows-lists.

gerrod3:
> Would still like to see the CRUD indicators in bold, as well as fixing the Workflows-lists.

My bad, I wasn't thinking of the CRUD indicators as verbs. Should the letters in the keys also be bolded?

ggainey:
> > Would still like to see the CRUD indicators in bold, as well as fixing the Workflows-lists.
> 
> My bad, I wasn't thinking of the CRUD indicators as verbs. Should the letters in the keys also be bolded?

Hm - I'm not sure we need to. What I'm mostly after here, is to pull the eye to the end of the line and separate the "these are implemented" from "this is the object" - does that make sense?

ggainey (approved):
Changes look great, thanks!

fkuep:
Thanks @gerrod3 !!
Looking over the document, I can see links to the workflows in the plugins documentation , which most times use the http binary instead of pulp-cli. 
Since the server plugins are out-of-tree of pulp-cli pulp-cli/tree/main/tests/scripts serve as integration tests right ?
Should the tests like tests/scripts/pulp_ansible/test_content.sh , have a comment like : * `docs/supported_workflows.md#pulp_ansible : Role content / Collection content` 
to track what is really known to work, with a certain release ? 


@mdellweg with the addition of `docs/supported_workflows.md` now has *authorative ,versioned information on what it can do* ...
does @melcorr know already she can point to it from https://github.com/pulp/pulpproject.org/blob/gh-pages/content-plugins.md#pulp-3-cli-coverage ?

gerrod3:
> Thanks @gerrod3 !! Looking over the document, I can see links to the workflows in the plugins documentation , which most times use the http binary instead of pulp-cli. Since the server plugins are out-of-tree of pulp-cli pulp-cli/tree/main/tests/scripts serve as integration tests right ? Should the tests like tests/scripts/pulp_ansible/test_content.sh , have a comment like : `docs/supported_workflows.md#pulp_ansible : Role content / Collection content` to track what is really known to work, with a certain release ?

Yes it is up to the plugin writers to change their workflows to use the CLI. Currently `pulp_ansible`, `pulp_file`, and `pulp_python`, have CLI workflows. Some of the `pulpcore` workflows have been converted, but none of the `pulp_container` or `pulp_rpm` have been. It's my goal to work on converting those last 2 plugins next. The tests are currently used as integration tests if they have [`test_cli` set](https://github.com/pulp/pulp_rpm/blob/master/template_config.yml#L69). The tests don't have much visibility to regular users, so I'm not sure how helpful putting comments in them will be.

fkuep:
> . The tests are currently used as integration tests if they have [`test_cli` set](https://github.com/pulp/pulp_rpm/blob/master/template_config.yml#L69).

Yes - OK - Great!
.. thats way better !

PR #409, opened 2021-10-27
Make commands and options polymorphic

gerrod3:
fixes: TBD

Started playing around with a different version of command/option polymorphism that is hopefully much easier to use. Have two examples; option polymorphism for ansible remote options, and command polymorphism for container repository commands.  The poly options is pretty simplistic, but the poly commands require a new `type_callback` that is eagerly ran so that it properly updates in the help menu.

WIP, so tell me what you think of these methods.

### DIFF
--- a/pulpcore/cli/ansible/context.py
+++ b/pulpcore/cli/ansible/context.py
@@ -24,9 +24,8 @@ class PulpAnsibleCollectionVersionContext(PulpContentContext):
     CREATE_ID = "content_ansible_collection_versions_create"
     UPLOAD_ID: ClassVar[str] = "upload_collection"
 
-    def upload(self, file: IO[bytes], body: Any) -> Any:
-        body = self.preprocess_body(body)
-        return self.pulp_ctx.call(self.UPLOAD_ID, body=body, uploads={"file": file.read()})
+    def upload(self, file: IO[bytes]) -> Any:
+        return self.pulp_ctx.call(self.UPLOAD_ID, uploads={"file": file.read()})
 
 
 class PulpAnsibleRoleContext(PulpContentContext):
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -25,11 +25,20 @@ from pulpcore.cli.common.context import (
 
 _ = gettext.gettext
 
-_F = TypeVar("_F")
-_FC = TypeVar("_FC", Callable[..., Any], click.Command)
+_F = Callable[..., Any]
+_FC = TypeVar("_FC", _F, click.Command)
 
 
 class PulpCommand(click.Command):
+    def __init__(
+        self,
+        *args: Any,
+        allowed_with_contexts: Optional[Tuple[Type[PulpEntityContext]]] = None,
+        **kwargs: Any,
+    ):
+        self.allowed_with_contexts = allowed_with_contexts
+        super().__init__(*args, **kwargs)
+
     def get_short_help_str(self, limit: int = 45) -> str:
         return self.short_help or ""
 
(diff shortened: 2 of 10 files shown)


pulpcore/cli/ansible/context.py
```diff
@@ -24,9 +24,8 @@ class PulpAnsibleCollectionVersionContext(PulpContentContext):
     CREATE_ID = "content_ansible_collection_versions_create"
     UPLOAD_ID: ClassVar[str] = "upload_collection"
 
-    def upload(self, file: IO[bytes], body: Any) -> Any:
-        body = self.preprocess_body(body)
-        return self.pulp_ctx.call(self.UPLOAD_ID, body=body, uploads={"file": file.read()})
+    def upload(self, file: IO[bytes]) -> Any:
+        return self.pulp_ctx.call(self.UPLOAD_ID, uploads={"file": file.read()})
```
mdellweg, line 28: `body` was never used?

pulpcore/cli/common/generic.py
```diff
@@ -30,6 +30,15 @@
 
 
 class PulpCommand(click.Command):
+    def __init__(
+        self,
+        *args: Any,
+        needs_contexts: Optional[Tuple[Type[PulpEntityContext]]] = None,
+        **kwargs: Any,
+    ):
+        self.needs_contexts = needs_contexts if needs_contexts else None
```
mdellweg, line 39:
```suggestion
        self.needs_contexts = needs_contexts
```
In the worst case this could be an empty tuple. And if you think we _need_ to normalize, it would be `needs_contexts or None`.

pulpcore/cli/common/generic.py
```diff
@@ -42,15 +51,58 @@ def format_help_text(
...
         super().format_help_text(ctx, formatter)
 
+    def get_params(self, ctx: click.Context) -> List[click.Parameter]:
+        params = super().get_params(ctx)
+        new_params: List[click.Parameter] = []
+        # Is it more efficient to go through and just remove options from params?
+        for param in params:
+            if isinstance(param, PulpOption):
+                if param.needs_contexts is not None:
+                    if not isinstance(ctx.obj, param.needs_contexts):
```
mdellweg, line 61: This will refer to the immediate context. Are you sure, we don't need to check parent contexts too? Or maybe we must not?
Also you should document that `needs_contexts` are not _all_ but only _one_ is needed.

pulpcore/cli/common/generic.py
```diff
@@ -42,15 +51,58 @@ def format_help_text(
...
+                        continue
+            new_params.append(param)
+        return new_params
+
+
+class PulpGroup(click.Group):
+    def command(self, *args: Any, **kwargs: Any) -> Callable[[_FC], click.Command]:
```
mdellweg, line 68: I figured, we can be more specific on the type definitions used here.
Can you replace them?
```
_F = Callable[..., Any]
_FC = TypeVar("_FC", _F, click.Command)

File: pulpcore/cli/common/generic.py [hunk 5]
```diff
@@ -42,15 +51,58 @@ def format_help_text(
...
+        kwargs["cls"] = kwargs.get("cls", PulpCommand)
+        return super().command(*args, **kwargs)
+
+    def group(self, *args: Any, **kwargs: Any) -> Callable[[_FC], click.Group]:
+        kwargs["cls"] = kwargs.get("cls", PulpGroup)
+        return super().group(*args, **kwargs)
+
+    def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Command]:
+        # Overwriting this removes the command, while overwriting list_commands just hides it
```
mdellweg, line 77: Is `list_command` still listing it? Then we need to overwrite both.

pulpcore/cli/common/generic.py
```diff
@@ -42,15 +51,58 @@ def format_help_text(
...
 class PulpOption(click.Option):
     def __init__(
         self,
         *args: Any,
         needs_plugins: Optional[List[PluginRequirement]] = None,
+        needs_contexts: Optional[Tuple[Type[PulpEntityContext]]] = None,
         **kwargs: Any,
     ):
         self.needs_plugins = needs_plugins
+        self.needs_contexts = needs_contexts if needs_contexts else None
```
mdellweg, line 105:
```suggestion
        self.needs_contexts = needs_contexts
```

pulpcore/cli/common/generic.py
```diff
@@ -370,6 +422,39 @@ def _multi_option_callback(
...
+        "cls": PulpOption,
+        "default": type_names[0],
+        "is_eager": True,
+        "expose_value": False,
+    }
+
+    def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[str]) -> str:
+        pulp_ctx = ctx.find_object(PulpContext)
+        assert pulp_ctx
+        if value in choices:
```
mdellweg, line 441: I think we can assume click will provide either one choice or `None`.
```suggestion
        if value is not None:
```
And `None` should not even be possible given that we have a default.

mdellweg (commented):
I really like, what you did there.

I want to see @ggainey 's judgment about the amount of magic for contributors. As far as i can see, there is no breaking change introduced by this however.

pulpcore/cli/common/generic.py
gerrod3: Should I rename it to `allowed_contexts`?

pulpcore/cli/common/generic.py
gerrod3: Oh no, my comment wasn't clear. `list_commands` calls `get_command` when listing out the commands in the help message. If you overwrite `list_commands` you can hide the command from the help message, but it will still be callable since `get_command` can still access it.

pulpcore/cli/common/generic.py
gerrod3: Technically you can still override the default to be a value not inside choices, but I haven't tested this to see if it errors or not.

pulpcore/cli/ansible/context.py
gerrod3: It contained optional fields, which I removed.

pulpcore/cli/common/generic.py
mdellweg: Good call. What about `allowed_with_contexts` since the contexts are supposed to be there first.
Either is fine for me.

pulpcore/cli/common/generic.py
mdellweg: Thanks for the clarification. That last step of calling `get_command` for everything in `list_command` was not obvious.

pulpcore/cli/common/generic.py
gerrod3: Also, I think only immediate context should be check. Since parent contexts are set by the previous subcommands I don't think it should be necessary to check them since there should only be one variation of them for the current command executing, e.g. if they were different we would be executing a different command with different options.

mdellweg: Yeah, maybe i was worrying too much. You still get the "nearest" object that was assigned to a context. So this can still traverse multiple subcommands if they did not introduce a new EntityContext.

ggainey (approved):
I don't see any breaking changes here, and I can see this being really useful for, say, the rest of pulp_rpm content entities! (which makes me hopeful that some of the 'magic' is starting to stick... :) )

Great work gerrod!

PR #411, opened 2021-10-30
Update CI

mdellweg:
[noissue]


ggainey (changes requested):
I **think** I understand what you're doing here - but it took a bit. Since there isn't an issue, can you add some commentary in the commit-msg so future Us will know what this change is for?

ggainey (approved):

PR #412, opened 2021-10-30
Add rpm-comps-upload support.

### LINKED ISSUE #406: Add support for comps.xml uploads to rpm
## Summary

PR https://github.com/pulp/pulp_rpm/pull/2159 adds support for uploading comps.xml files. Add support for same to cli
## Examples

`pulp rpm content comps upload --file --repository --replace`

ggainey:
closes #406
Required PR: https://github.com/pulp/pulp_rpm/pull/2159

### DIFF
--- a/pulpcore/cli/rpm/comps.py
+++ b/pulpcore/cli/rpm/comps.py
@@ -1,0 +1,46 @@
+import gettext
+from typing import IO, Optional
+
+import click
+
+from pulpcore.cli.common.context import (
+    EntityFieldDefinition,
+    PluginRequirement,
+    PulpContext,
+    PulpEntityContext,
+    pass_pulp_context,
+)
+from pulpcore.cli.common.generic import resource_option
+from pulpcore.cli.rpm.context import PulpRpmCompsXmlContext, PulpRpmRepositoryContext
+
+_ = gettext.gettext
+
+repository_option = resource_option(
+    "--repository",
+    default_plugin="rpm",
+    default_type="rpm",
+    context_table={"rpm:rpm": PulpRpmRepositoryContext},
+    href_pattern=PulpRpmRepositoryContext.HREF_PATTERN,
+    help=_("Repository to associate the comps-units to, takes <name> or href."),
+)
+
+
+@click.command()
+@click.option("--file", type=click.File("rb"), required=True)
+@repository_option
+@click.option("--replace", type=bool, default=False)
+@pass_pulp_context
+def comps_upload(
+    pulp_ctx: PulpContext,
+    file: IO[bytes],
+    repository: Optional[EntityFieldDefinition],
+    replace: Optional[bool],
+) -> None:
+    """Create comps.xml content-units by uploading a comps.xml-formatted file"""
+    pulp_ctx.needs_plugin(PluginRequirement("rpm", min="3.17.0dev"))
+    entity_ctx = PulpRpmCompsXmlContext(pulp_ctx)
+    href = None
+    if isinstance(repository, PulpEntityContext):
+        href = repository.pulp_href
+    result = entity_ctx.upload_comps(file, href, replace)
+    pulp_ctx.output_result(result)
--- a/pulpcore/cli/rpm/context.py
+++ b/pulpcore/cli/rpm/context.py
@@ -1,5 +1,7 @@
 import gettext
-from typing import Any
+from typing import IO, Any, ClassVar, Optional
+
+import click
 
 from pulpcore.cli.common.context import (
     EntityDefinition,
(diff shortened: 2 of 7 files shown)


mdellweg:
I see the RPM PR is still in flight. Should we draft this?

pulpcore/cli/rpm/context.py
```diff
@@ -33,6 +35,21 @@ def refresh(self, href: str) -> Any:
         )
 
 
+class PulpRpmCompsXmlContext(PulpEntityContext):
+    UPLOAD_COMPS_ID = "upload_comps"
```
mdellweg, line 39:
```suggestion
    UPLOAD_COMPS_ID: ClassVar[str] = "upload_comps"
```
Is there any reason not to call this `UPLOAD_ID`?

ggainey:
> I see the RPM PR is still in flight. Should we draft this?

Prob a good idea, I still working to get a response from the stakeholder that the API in 2159 meets their needs.

pulpcore/cli/rpm/context.py
ggainey: > Is there any reason not to call this `UPLOAD_ID`?

Mmm, I believe the original intent was to differentiate from the uploading of model-entities like Package and Advisory. Prob not useful, will change.

mdellweg: Actually, i kind of think it should be `rpm_comps_upload` on the server side...

ggainey: Hrm, that was my initial id, but then I changed it to conform to copy - I should have stuck w/ my first thought, I suppose. Well, the PR is waiting for final stakeholder approval, so we can fixit!

ggainey:
@mdellweg ready for final review

pulpcore/cli/rpm/context.py
```diff
@@ -33,6 +35,21 @@ def refresh(self, href: str) -> Any:
...
+    UPLOAD_COMPS_ID: ClassVar[str] = "rpm_comps_upload"
+
+    def upload_comps(
+        self, file: IO[bytes], repo_href: Optional[str], replace: Optional[bool]
+    ) -> Any:
+        click.echo(_("Uploading file {filename}").format(filename=file.name), err=True)
+        file.seek(0)
+        return self.pulp_ctx.call(
+            self.UPLOAD_COMPS_ID,
+            uploads={"file": file.read()},
```
mdellweg, line 48: Are we quite confident, these never need to go through the uploads api for being too big?
(From the CLI view, i kind of think, there should only have been a single way to upload data into pulp, but at least every other serializer that accepts files, should also accept uploads or artifacts.)

pulpcore/cli/rpm/comps.py
```diff
@@ -0,0 +1,56 @@
...
+    if repo_type == "rpm":
+        pulp_ctx.needs_plugin(PluginRequirement("rpm", min="3.17.0dev"))
+        ctx.obj = PulpRpmCompsXmlContext(pulp_ctx)
+    else:
+        raise NotImplementedError()
+
+
+@comps.command()
+@click.option("--file", type=click.File("rb"), required=True)
+@click.option("--repository", type=str)
```
mdellweg, line 37: You should turn this into a `resource_option`.

pulpcore/cli/rpm/comps.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@click.group()
+@click.option(
+    "-t",
+    "--type",
+    "repo_type",
+    type=click.Choice(["rpm"], case_sensitive=False),
+    default="rpm",
+)
```
mdellweg, line 24: With the resource option, we don't do this anymore.

pulpcore/cli/rpm/context.py
ggainey: comps.xml isn't recognized by pulp_rpm as an entity itself - it's a convenient way of getting the PackageCatalog/Environment/Langpack/Groups uploaded, but there is no 'comps.xml' file anywhere. As far as size goes, the largest comps.xml I know of is Fedora-Updates; F34 unzipped is 1.6Mb. That's def Large; you're right, maybe I need to rethink "it's just a file".

pulpcore/cli/rpm/comps.py
ggainey: :+1:

pulpcore/cli/rpm/comps.py
mdellweg: I meant the whole "-t" option seems superfluous at this point. Also as you just explained, the "comps" is not a "thing" in Pulp. So i would be ok to say we can merge the command in to "`comps-upload`". WDYT?

pulpcore/cli/rpm/comps.py
```diff
@@ -0,0 +1,65 @@
...
+@pass_entity_context
+@pass_pulp_context
+def upload(
+    pulp_ctx: PulpContext,
+    entity_ctx: PulpRpmCompsXmlContext,
+    file: IO[bytes],
+    repository: Optional[EntityFieldDefinition],
+    replace: Optional[bool],
+) -> None:
+    """Create an rpm package content unit by uploading a file"""
```
mdellweg, line 60: This is probably wrong. Keep in mind that this docstring is user facing.

pulpcore/cli/rpm/context.py
ggainey: The chunked-upload-api doesn't play very well with either "convenience API" or "not-an-artifact". Given that the largest comps.xml I know if is 1.7Mb, comps.xml isn't its own artifact, and the pulp_rpm create calls all have --file= parameters, I think this is the way to go.

If at some point we decide to, say, add a comps-Artifact and extends the POST /comps/ call  to include an --artifact option, we can extend the CLI to handle it using chunked uploads. But for the purposes of filling the RFE, I think this is sufficient, and consistent.

pulpcore/cli/rpm/comps.py
ggainey: Ooo, nice catch - copypasta strikes again!

pulpcore/cli/rpm/comps.py
ggainey: I think this all sounds like a fine idea. Ready for review.

mdellweg (approved):

PR #414, opened 2021-11-03
Extract the architecture's description

lubosmj:
This PR makes the architecture section more visible. Before this change, the description of the pulp CLI architecture was cluttered inside the section `Contributing`.


mkdocs.yml
```diff
@@ -1,7 +1,7 @@
 site_name: Pulp 3 CLI
 theme: readthedocs
 repo_url: https://github.com/pulp/pulp-cli
-edit_uri: edit/develop/docs/
+edit_uri: edit/main/docs/
```
mdellweg, line 4: good catch!

mdellweg (commented):
I see your request to make the section more visible, but everything written in this section is only interesting to developers wanting to contribute, no?
Can we at least move it after the contributing part?

lubosmj:
I have moved it to after the contributing part.

mdellweg (approved):

PR #419, opened 2021-11-07
Refactor plugin loading

mdellweg:
Deferred the plugin loading to the entry point to prevent some chances
of circular imports. Loaded plugins should provide a `mount` method to
register their commands to the main group.

### DIFF
--- a/pulp_cli/__init__.py
+++ b/pulp_cli/__init__.py
@@ -1,1 +1,20 @@
-from pulpcore.cli.common import main  # noqa: F401
+from types import ModuleType
+from typing import Any, Dict
+
+import click
+import pkg_resources
+
+
+def main() -> Any:
+    ##############################################################################
+    # Load plugins
+    # https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata
+    discovered_plugins: Dict[str, ModuleType] = {
+        entry_point.name: entry_point.load()
+        for entry_point in pkg_resources.iter_entry_points("pulp_cli.plugins")
+    }
+    _main: click.Group = discovered_plugins["common"].main
+    for plugin in discovered_plugins.values():
+        if hasattr(plugin, "mount"):
+            plugin.mount(_main, discovered_plugins=discovered_plugins)
+    return _main()
--- a/pulpcore/cli/file/__init__.py
+++ b/pulpcore/cli/file/__init__.py
@@ -1,5 +1,9 @@
-from pulpcore.cli.common import main
+from typing import Any
+
+import click
+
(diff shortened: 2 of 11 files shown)


pulp_cli/__init__.py
```diff
@@ -1 +1,19 @@
...
+
+def main() -> Any:
+    ##############################################################################
+    # Load plugins
+    # https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata
+    discovered_plugins: Dict[str, ModuleType] = {
+        entry_point.name: entry_point.load()
+        for entry_point in pkg_resources.iter_entry_points("pulp_cli.plugins")
+    }
+    _main: Callable[[], Any] = discovered_plugins["common"].main
```
gerrod3, line 15: What happens if a plugin also declares a common folder? Does it overwrite it?

pulpcore/cli/file/__init__.py
```diff
@@ -12,15 +15,17 @@
 _ = translation.gettext
 
 
-@main.group(name="file")
+@click.group(name="file")
```
gerrod3, line 18: Should we make these `PulpGroup`s?

gerrod3 (commented):
I'm confused, how are the other plugin commands being loaded now? I only see support for the core and file commands.

*Edit* Nevermind, they are being "loaded" (commands added) just like before, there is now an option that a plugin can load itself through the new `mount` method. So, current plugins should continue to work with this new method.

pulp_cli/__init__.py
mdellweg: No problem unless the plugin is named "common". Maybe i am missing something.
Are you saying that multiple packages could define the same `entry_point` in `"pulp_cli.plugins"`. I suspect this would break the system already today.

pulpcore/cli/file/__init__.py
mdellweg: Probably yes.

pulp_cli/__init__.py
mdellweg: Also in the long run we might pull the main function out of the common plugin, or make common not-a-plugin. wdyt?

gerrod3: Right now all the other *external* plugins are separate packages containing just that plugin, but I wonder what would happen if you had a package that contained multiple plugins with one of them also being named "common". It would probably break today as well, but I think this question could be expanded to any name conflicts between the packages. How badly would it break and is this a scenario we should be concern about?

I think having common not be a plugin makes sense. All the other plugins rely on it, so it isn't optional like the "plugin" nomenclature would suggest.

mdellweg: But it was not so straight forward to do, so i started small.

PR #420, opened 2021-11-10
Ignore content commands when there are no options

lubosmj:
The rationale behind this change is that we do not want to expose commands which are not currently implemented. Specific content commands may not be supported by all plugin writers.

The only exception occurs for the 'content_list' command which is configured with content types' values by default. Therefore, this command is created even though no decorator options were previously specified.

[noissue]

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -864,11 +864,11 @@ def repository_content_command(**kwargs: Any) -> click.Group:
         rc = [unit.pulp_href for unit in remove_content] if remove_content else None
         repo_ctx.modify(repo_ctx.pulp_href, ac, rc, base_version.pulp_href)
 
-    command_decorators = {
+    command_decorators: Dict[click.Command, Optional[List[Callable[[_FC], _FC]]]] = {
         content_list: kwargs.pop("list_decorators", []),
-        content_add: kwargs.pop("add_decorators", []),
-        content_remove: kwargs.pop("remove_decorators", []),
-        content_modify: kwargs.pop("modify_decorators", []),
+        content_add: kwargs.pop("add_decorators", None),
+        content_remove: kwargs.pop("remove_decorators", None),
+        content_modify: kwargs.pop("modify_decorators", None),
     }
     if not kwargs.get("name"):
         kwargs["name"] = "content"
@@ -880,8 +880,9 @@ def repository_content_command(**kwargs: Any) -> click.Group:
         ctx.obj = PulpContentContext(pulp_ctx)
 
     for command, options in command_decorators.items():
-        for option in options:
-            command = option(command)
-        content_group.add_command(command)
+        if options is not None:
+            for option in options:
+                command = option(command)
+            content_group.add_command(command)
 
     return content_group


pulpcore/cli/common/generic.py
```diff
@@ -879,9 +878,15 @@ def content_modify(
...
+        content_group.add_command(content_command)
+
+    _add_command(content_list, kwargs.pop("list_decorators", []))
+
     for command, options in command_decorators.items():
-        for option in options:
-            command = option(command)
-        content_group.add_command(command)
+        if options:
+            _add_command(command, options)
```
mdellweg, line 890:
```suggestion
        if options is not None:
            for option in options:
                command = option(command)
            content_group.add_command(command)
```

pulpcore/cli/common/generic.py
```diff
@@ -879,9 +878,15 @@ def content_modify(
     def content_group(ctx: click.Context, pulp_ctx: PulpContext) -> None:
         ctx.obj = PulpContentContext(pulp_ctx)
 
+    def _add_command(content_command: Any, command_options: List[Any]) -> None:
+        for option in command_options:
+            content_command = option(content_command)
+        content_group.add_command(content_command)
+
+    _add_command(content_list, kwargs.pop("list_decorators", []))
+
```
mdellweg, line 887: This seems unnecessary.
```suggestion
```

pulpcore/cli/common/generic.py
```diff
@@ -865,7 +865,6 @@ def content_modify(
         repo_ctx.modify(repo_ctx.pulp_href, ac, rc, base_version.pulp_href)
 
     command_decorators = {
-        content_list: kwargs.pop("list_decorators", []),
         content_add: kwargs.pop("add_decorators", []),
         content_remove: kwargs.pop("remove_decorators", []),
         content_modify: kwargs.pop("modify_decorators", []),
```
mdellweg, line 870: Keep that line, and default the others to `None`.

pulpcore/cli/common/generic.py
```diff
@@ -865,7 +865,6 @@ def content_modify(
         repo_ctx.modify(repo_ctx.pulp_href, ac, rc, base_version.pulp_href)
 
     command_decorators = {
```
mdellweg, line 867:
```suggestion
    command_decorators: Dict[click.Command, Optional[list[_FC]]] = {
```

mdellweg (commented):
Let's make this a bit simpler.

pulpcore/cli/common/generic.py
mdellweg: Sorry the MetaType is `List[...]` (capital `L`).

lubosmj: Easy peasy and straighforward:
```python
    command_decorators: Dict[click.Command, Optional[List[Callable[[_FC], _FC]]]] = {
```

mdellweg: Right. That's the one we need.

mdellweg: Do we see `Callable[[_FC], _FC]` elsewhere? Then we should put a name on it.

lubosmj: Yep, we do. Bunch of times:
```
[lmjachky@localhost pulp-cli]$ git grep "_FC"
pulpcore/cli/common/generic.py:_FC = TypeVar("_FC", Callable[..., Any], click.Command)
pulpcore/cli/common/generic.py:def pulp_option(*args: Any, **kwargs: Any) -> Callable[[_FC], _FC]:
pulpcore/cli/common/generic.py:def resource_option(*args: Any, **kwargs: Any) -> Callable[[_FC], _FC]:
pulpcore/cli/common/generic.py:    command_decorators: Dict[click.Command, Optional[List[Callable[[_FC], _FC]]]] = {
```

I am adding a constant for that.

mdellweg (approved):

PR #427, opened 2021-11-12
Add a user agent header to REST calls

### LINKED ISSUE #426: pulp-cli should have its own user-agent header on requests
It would be great to be able to tell which API calls were hitting a pulp-instance from the CLI (as opposed to coming from the bindings or directly from, say, httpie) so we could start measuring uptake!

mdellweg:
fixes #426


ggainey (approved):
shippit!

Issue #423, opened 2021-11-11
Add tagging/untagging for container repositories


gerrod3:
## Summary
Both types of container repositories can be tagged and untagged. Should the commands be grouped under a tag subcommand under repository or just added as two separate commands?

## Examples
`pulp container repository tag --repository foo --name "custom_tag" --digest "manifest_digest"`
`pulp container repository -t push untag --repository phoo --name "another_tag"`

or

`pulp container repository tag add --repository foo --name "custom_tag" --digest "manifest_digest"`
`pulp container repository -t push tag remove --repository phoo --name "another_tag"`

or maybe it's common enough to warrant it's own subgroup in container commands

`pulp container tag add --repository foo --name "custon_tag" --digest "manifest_digest"`
`pulp container tag remove --repository push:phoo --name "another_tag"`
`pulp container tag list --repository foo` *overlaps with `content -t tag list` and `repository content list -t tag`*

Tell me which command structure would be preferred.

lubosmj:
I am good with the first option.

ipanova:
I'd go with option 1 since these are actions which are performed under the repository

PR #425, opened 2021-11-12
Add content commands to container

gerrod3:
Needs polymorphic PR to work, marking as draft till merged.

PR #429, opened 2021-11-12
Add tag/untag commands for container repositories

gerrod3:
fixes: #423

See issue for discussion around command structure. I would like to have #425 merged first so I can use those commands for creating tests for these commands. I'll update the `suppported_workflows.md` after I'm done with all the new container commands to avoid merge conflicts.

### DIFF
--- a/pulpcore/cli/container/context.py
+++ b/pulpcore/cli/container/context.py
@@ -83,6 +83,7 @@ class PulpContainerRepositoryContext(PulpRepositoryContext):
     CAPABILITIES = {
         "sync": [PluginRequirement("container")],
         "pulpexport": [PluginRequirement("container", "2.8.0.dev")],
+        "tag": [PluginRequirement("container", "2.3.0")],
     }
 
 
@@ -90,6 +91,7 @@ class PulpContainerPushRepositoryContext(PulpRepositoryContext):
     HREF = "container_container_push_repository_href"
     ID_PREFIX = "repositories_container_container_push"
     VERSION_CONTEXT = PulpContainerPushRepositoryVersionContext
+    CAPABILITIES = {"tag": [PluginRequirement("container", "2.3.0")]}
 
 
 registered_repository_contexts["container:container"] = PulpContainerRepositoryContext
--- a/pulpcore/cli/container/repository.py
+++ b/pulpcore/cli/container/repository.py
@@ -1,3 +1,4 @@
+import re
 from typing import Any, Dict
 
 import click
@@ -37,6 +38,16 @@ from pulpcore.cli.core.generic import task_command
 
 translation = get_translation(__name__)
 _ = translation.gettext
+VALID_TAG_REGEX = r"^[A-Za-z0-9][A-Za-z0-9._-]*$"
+
+
+def _tag_callback(ctx: click.Context, param: click.Parameter, value: str) -> str:
+    if len(value) == 0:
+        raise click.ClickException("Please pass a non empty tag name.")
+    if re.match(VALID_TAG_REGEX, value) is None:
+        raise click.ClickException("Please pass a valid tag.")
+
+    return value
 
 
 remote_option = resource_option(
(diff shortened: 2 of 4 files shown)


pulpcore/cli/container/context.py
```diff
@@ -75,6 +76,8 @@ class PulpContainerRepositoryContext(PulpRepositoryContext):
     UPDATE_ID = "repositories_container_container_partial_update"
     DELETE_ID = "repositories_container_container_delete"
     SYNC_ID = "repositories_container_container_sync"
+    TAG_ID: ClassVar[str] = "repositories_container_container_tag"
+    UNTAG_ID: ClassVar[str] = "repositories_container_container_untag"
     VERSION_CONTEXT = PulpContainerRepositoryVersionContext
     CAPABILITIES = {
         "sync": [PluginRequirement("container")],
```
gerrod3, line 83: If adding/removing tags between the two repository types is supported from different versions of pulp_container then I can add capabilities here. Might be worth to do it anyway since I don't know which version tagging was added in.

pulpcore/cli/container/repository.py
```diff
@@ -119,3 +119,42 @@ def sync(
...
+@pass_repository_context
+def add_tag(repository_ctx: PulpRepositoryContext, digest: str, tag: str,) -> None:
+    if len(tag) == 0:
+        raise click.ClickException("Please pass a non empty tag name.")
+
+    digest = digest.strip()
+    if not digest.startswith("sha256:"):
+        digest = f"sha256:{digest}"
+    if len(digest) != 71:  # len("sha256:" + 64
+        raise click.ClickException("Improper SHA256, please provide a valid 64 digit digest.")
```
gerrod3, line 138: Tried doing some bare minimum checks and leaving the rest to the endpoint's serializer. If there is anything else needed please tell me.

gerrod3 (commented):
Commands are pretty simple, but the main question I have is about their placement. I would like for @pulp/container-plugin to be in consensus so it doesn't have to be changed later.

pulpcore/cli/container/context.py
lubosmj: Support for tagging manifests within push repositories was added in version 2.3.0 (2021-02-08). According to the pulp-cli [docs](https://github.com/pulp/pulp_container/blob/2.3/requirements.txt#L1), we support `5 versions of pulpcore from 3.11-3.16 and 5 of Pulp's plugins: pulp_ansible, pulp_container`. Version 2.3.0 falls into the pulpcore requirement when looking at [requirements.txt](https://github.com/pulp/pulp_container/blob/b81d233f0ce950153cfea9165cd7b4c2aad3286f/requirements.txt#L1).

pulpcore/cli/container/context.py
```diff
@@ -101,10 +103,13 @@ class PulpContainerRepositoryContext(PulpRepositoryContext):
     UPDATE_ID = "repositories_container_container_partial_update"
     DELETE_ID = "repositories_container_container_delete"
     SYNC_ID = "repositories_container_container_sync"
+    TAG_ID: ClassVar[str] = "repositories_container_container_tag"
+    UNTAG_ID: ClassVar[str] = "repositories_container_container_untag"
```
mdellweg, line 107: This looks like a rebase is needed.

gerrod3: Are you talking about your new PR refactoring the IDs? #448 I can wait till after it's merged to rebase.

mdellweg: Oh, yes, i get confused what's merged or not.

pulpcore/cli/container/repository.py
```diff
@@ -37,6 +38,16 @@
 
 translation = get_translation(__name__)
 _ = translation.gettext
+VALID_TAG_REGEX = r"^[A-Za-z0-9][A-Za-z0-9._-]*$"
+
+
+def _tag_callback(ctx: click.Context, param: click.Parameter, value: str) -> str:
+    if len(value) == 0:
+        raise click.ClickException("Please pass a non empty tag name.")
```
mdellweg, line 46: FWIW, an empty string would not pass the regex. But i see this check provide a better error condition.

pulpcore/cli/container/repository.py
```diff
@@ -122,3 +133,46 @@ def sync(
...
+@click.option("--tag", help=_("Name to tag an image with"), required=True, callback=_tag_callback)
+@click.option("--digest", help=_("SHA256 digest of the Manifest file"), required=True)
+@pass_repository_context
+def add_tag(
+    repository_ctx: PulpRepositoryContext,
+    digest: str,
+    tag: str,
+) -> None:
+    if not repository_ctx.capable("tag"):
+        raise click.ClickException(_("pulp_container 2.3.0 is required to tag images"))
```
mdellweg, line 150: (Note for later:) Ha, we should have a `needs_capability` that raises itself.
Rationale: All mentions of Pulp versions should be in `context.py`.

mdellweg (approved):

PR #431, opened 2021-11-13
Add support for the "tasks purge" command.

### LINKED ISSUE #430: Support task-purge command
See https://pulp.plan.io/issues/8554 for details of the API.

ggainey:
closes #430
Required PR: https://github.com/pulp/pulpcore/pull/1721

### DIFF
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -278,6 +278,7 @@ class PulpTaskContext(PulpEntityContext):
     READ_ID = "tasks_read"
     DELETE_ID = "tasks_delete"
     CANCEL_ID: ClassVar[str] = "tasks_cancel"
+    PURGE_ID = "tasks_purge"
 
     resource_context: Optional[PulpEntityContext] = None
 
@@ -294,6 +295,17 @@ class PulpTaskContext(PulpEntityContext):
             return {"reserved_resources_record": self.resource_context.pulp_href}
         else:
             return {}
+
+    def purge(self, finished_before: Optional[str], states: Optional[List[str]]) -> Any:
+        body: Dict[str, Any] = {}
+        if finished_before:
+            body["finished_before"] = finished_before
+        if states:
+            body["states"] = states
+        return self.pulp_ctx.call(
+            self.PURGE_ID,
+            body=body,
+        )
 
 
 class PulpTaskGroupContext(PulpEntityContext):
--- a/pulpcore/cli/core/task.py
+++ b/pulpcore/cli/core/task.py
@@ -1,11 +1,13 @@
 import gettext
 import re
 from contextlib import suppress
-from typing import Optional
+from datetime import datetime
+from typing import Optional, Tuple
 
 import click
 
 from pulpcore.cli.common.context import (
+    PluginRequirement,
     PulpContext,
     PulpEntityContext,
     pass_entity_context,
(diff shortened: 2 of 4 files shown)


pulpcore/cli/core/context.py
```diff
@@ -262,6 +263,17 @@ def scope(self) -> Dict[str, Any]:
         else:
             return {}
 
+    def purge(self, finished_before: Optional[str], states: Optional[List[str]]) -> Any:
+        body: Dict[str, Union[str, List[str]]] = {}
```
mdellweg, line 267: I appreciate the effort, but for these, we usually do `body: Dict[str, Any]`, because `Any` is the closest thing we have to `JSONSerializable`, or `ConstructOfPythonPrimitives`.

pulpcore/cli/core/task.py
```diff
@@ -112,3 +114,33 @@ def cancel(
...
+
+DATETIME_FORMATS = ["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S"]
+
+
+@task.command()
+@click.option(
+    "--finished-before",
+    "finished",
+    help=_("Purge task-records whose 'finished' time is **before** the time specified."),
+    type=click.DateTime(formats=DATETIME_FORMATS),
```
mdellweg, line 127: Is click injecting the formatting info into the help text?

pulpcore/cli/core/task.py
```diff
@@ -112,3 +114,33 @@ def cancel(
...
+)
+@click.option(
+    "--states",
```
mdellweg, line 130:
```suggestion
    "--state",
    "states",
```

pulpcore/cli/core/task.py
```diff
@@ -112,3 +114,33 @@ def cancel(
...
+    help=_("Only purge tasks in the specified state(s)."),
```
mdellweg, line 131:
```suggestion
    help=_("Only purge tasks in the specified state(s). Can be specified multiple times."),
```

pulpcore/cli/core/task.py
```diff
@@ -112,3 +114,33 @@ def cancel(
...
+)
+@pass_entity_context
+@pass_pulp_context
+def purge(
+    pulp_ctx: PulpContext,
+    task_ctx: PulpTaskContext,
+    finished: Optional[datetime],
+    states: Optional[List[str]],
+) -> None:
+    pulp_ctx.needs_plugin(PluginRequirement("core", "3.17.0.dev"))
```
mdellweg, line 143: :+1: Thanks for not forgetting about that.

pulpcore/cli/core/task.py
```diff
@@ -112,3 +114,33 @@ def cancel(
...
+    state_list = list(states) if states else None
```
mdellweg, line 144: According to the typing info, states is already a List[str] or None.
But i think, click passes it as a Tuple. So either this transformation is unnecessary, or the typing info is incorrect (and yes, sadly mypy does not account for the click decorators that rigorously).

mdellweg (commented):
Let's try to get this into the 0.13.0 release. Please add to the milestone if you agree.

pulpcore/cli/core/task.py
ggainey: Yes it is

mdellweg (approved):

PR #432, opened 2021-11-16
Container content management commands

gerrod3:
Mega WIP

### DIFF
--- a/pulpcore/cli/container/context.py
+++ b/pulpcore/cli/container/context.py
@@ -124,6 +124,25 @@ class PulpContainerRepositoryContext(PulpContainerBaseRepositoryContext):
         if add_content:
             self.call("add", parameters={self.HREF: href}, body={"content_units": add_content})
 
+    def copy_tag(self, source_href: str, tags: Optional[List[str]]) -> Any:
+        body = {"source_repository_version": source_href, "names": tags}
+        body = self.preprocess_body(body)
+        return self.call("copy_tags", parameters={self.HREF: self.pulp_href}, body=body)
+
+    def copy_manifest(
+        self,
+        source_href: str,
+        digests: Optional[List[str]],
+        media_types: Optional[List[str]],
+    ) -> Any:
+        body = {
+            "source_repository_version": source_href,
+            "digests": digests,
+            "media_types": media_types,
+        }
+        body = self.preprocess_body(body)
+        return self.call("copy_manifests", parameters={self.HREF: self.pulp_href}, body=body)
+
 
 class PulpContainerPushRepositoryContext(PulpContainerBaseRepositoryContext):
     HREF = "container_container_push_repository_href"
--- a/pulpcore/cli/container/repository.py
+++ b/pulpcore/cli/container/repository.py
@@ -1,5 +1,5 @@
 import re
-from typing import Any, Dict
+from typing import Any, Dict, List, Optional
 
 import click
 
@@ -50,12 +50,33 @@ VALID_TAG_REGEX = r"^[A-Za-z0-9][A-Za-z0-9._-]*$"
 
 def _tag_callback(ctx: click.Context, param: click.Parameter, value: str) -> str:
     if len(value) == 0:
-        raise click.ClickException("Please pass a non empty tag name.")
+        raise click.ClickException(_("Please pass a non empty tag name."))
     if re.match(VALID_TAG_REGEX, value) is None:
-        raise click.ClickException("Please pass a valid tag.")
+        raise click.ClickException(_("Please pass a valid tag."))
 
     return value
 
+
+source_option = resource_option(
+    "--source",
+    default_plugin="container",
+    default_type="container",
+    context_table={
+        "container:container": PulpContainerRepositoryContext,
+        "container:push": PulpContainerPushRepositoryContext,
+    },
+    href_pattern=PulpRepositoryContext.HREF_PATTERN,
+    help=_(
+        "Source repository to copy content from in the form `[[<plugin>:]<resource_type>:]<name>' "
+        "or by href."
+    ),
+    required=True,
+)
+
+
+version_option = click.option(
+    "--version", help=_("Version of the source repository to use"), type=int
+)
 
 remote_option = resource_option(
     "--remote",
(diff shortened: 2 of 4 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -326,7 +326,7 @@ def _callback(
 
 
 # based on https://stackoverflow.com/a/42865957/2002471
-units = {"B": 1, "KB": 10**3, "MB": 10**6, "GB": 10**9, "TB": 10**12}
+units = {"B": 1, "KB": 10 ** 3, "MB": 10 ** 6, "GB": 10 ** 9, "TB": 10 ** 12}
```
mdellweg, line 329: Did black change their mind again, or is your version outdated?

pulpcore/cli/common/generic.py
```diff
@@ -1002,29 +1000,30 @@ def version_callback(
         )
         return repo_ver_ctx
 
-    @click.command("list")
-    @click.option("-t", "--type", "type", type=click.Choice(names), default=names[0])
+    @pulp_command("list")
+    @click.option("--all-types", is_flag=True)
```
gerrod3, line 1004: I've changed the default behavior of this command group. It might make more sense to pull this out into a separate PR.

pulpcore/cli/container/repository.py
```diff
@@ -44,22 +51,62 @@
...
+        # Check that the version makes sense
+        latest_version = int(repo_ctx.entity["latest_version_href"].split("/")[-2])
+        if not (0 < int(version) <= latest_version):
+            raise click.ClickException(
+                _(
+                    "Please specify a version that is less then the latest version {} and greater"
+                    " than 0"
+                ).format(latest_version)
+            )
+        repo_ctx.meta["version"] = version
```
gerrod3, line 99: Can you just store what you want in meta? Because that's what I doing.

pulpcore/cli/common/generic.py
mdellweg: Well, black said, they are now stable and won't change incompatible anymore...

pulpcore/cli/container/repository.py
mdellweg: Hmm good question, i don't even see it being used somewhere else.

mdellweg: One concern i have with such things is the lifetime, once we get around to more pulp-cli-shell operations, where such a meta object could leak from one command into the next. But we are not there...

gerrod3: I can pop the value from the meta object so that it goes back to its original state.

pulpcore/cli/ansible/repository.py
```diff
@@ -121,13 +120,6 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
...
-    type_option(
-        choices={
-            "collection-version": PulpAnsibleCollectionVersionContext,
-            "role": PulpAnsibleRoleContext,
-        },
-        default="collection-version",
-    ),
     href_option,
 ]
 content_json_callback = create_content_json_callback(schema=CONTENT_LIST_SCHEMA)
```
mdellweg, line 125: I am a bit lost here. This schema is supposed to handle both collections and roles?

pulpcore/cli/ansible/repository.py
```diff
@@ -150,13 +142,6 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
     The argument prefixed with the '@' can be the path to a JSON file with a list of objects."""
         ),
     ),
-    type_option(
-        choices={
-            "collection-version": PulpAnsibleCollectionVersionContext,
-            "role": PulpAnsibleRoleContext,
-        },
-        default="collection-version",
-    ),
```
mdellweg, line 159: So these commands are supposed to handle all types of content in unison now?

pulpcore/cli/common/generic.py
mdellweg: What motivates the change?

pulpcore/cli/common/generic.py
```diff
@@ -1002,29 +1000,30 @@ def version_callback(
...
-        type: Optional[str],
+        all_types: Optional[bool],
         **params: Any,
     ) -> None:
         parameters = {k: v for k, v in params.items() if v is not None}
         parameters.update({"repository_version": version.pulp_href})
-        result = content_contexts[type](pulp_ctx).list(
-            limit=limit, offset=offset, parameters=parameters
-        )
+        ctx_obj = PulpContentContext(pulp_ctx) if all_types else ctx.obj
```
mdellweg, line 1022: So this will either result in an object of type `PulpContentContext` or `PulpRepositroy`? That sounds wrong to me.
ALso what is the desired behaviour if `all_types` is `False`?

pulpcore/cli/ansible/repository.py
gerrod3: Yes, since collection and roles have the same identifiers (namespace, name, version).

pulpcore/cli/ansible/repository.py
gerrod3: This option is now apart of the content command group. I've moved it back one spot in the command.

pulpcore/cli/common/generic.py
gerrod3: No it will always be a subtype of `PulpContentContext`, I should probably use `pass_content_context` since this group's context object will always be set to a content context due to the type option.

pulpcore/cli/container/repository.py
mdellweg: Shouldn't we return a RepoVersionContext from this parameter?

gerrod3: I did want to, but `meta`'s type definition is `Dict[str, str]`.

mdellweg: Sorry, i meant the whole callback.

gerrod3: Since the API takes either repository or repository version I decided attaching the version to a `RepositoryContext` was easiest way to determine which the user was specifying.

mdellweg: Is specifying no version different from specifying the latest version?

gerrod3: Looking at the serializer, no they are the same (except you can't specify a repository with 0 versions): https://github.com/pulp/pulp_container/blob/main/pulp_container/app/serializers.py#L527

gerrod3: Updated to return `PulpContainerRepositoryVersionContext`.

pulpcore/cli/container/context.py
```diff
@@ -110,6 +110,39 @@ class PulpContainerRepositoryContext(PulpContainerBaseRepositoryContext):
...
+        add_content: Optional[List[str]] = None,
+        remove_content: Optional[List[str]] = None,
+        base_version: Optional[str] = None,
+    ) -> Any:
+        if remove_content:
+            self.call(
+                "remove", parameters={self.HREF: href}, body={"content_units": remove_content}
+            )
+        if add_content:
+            self.call("add", parameters={self.HREF: href}, body={"content_units": add_content})
```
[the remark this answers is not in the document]
gerrod3: To which class, the `PulpContainerBaseRepositoryContext`? I don't think so since only the regular container repository can call the add and remove endpoints. Also this modify call is needed since that is what the content factory uses to create the add and remove commands.

pulpcore/cli/container/repository.py
```diff
@@ -45,22 +53,66 @@
...
+    pulp_ctx = ctx.find_object(PulpContext)
+    assert pulp_ctx is not None
+    if len(value) == 0:
+        raise click.ClickException(_("Please pass in a non empty source repository."))
+
+    version: Optional[str] = None
+    pulp_href: Optional[str] = None
+    entity: Optional[EntityDefinition] = None
+    if value.startswith("/"):
+        pattern = rf"^{pulp_ctx.api_path}{PulpContainerRepositoryContext.HREF_PATTERN}"
```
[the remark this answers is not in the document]
gerrod3: Technically, but it shouldn't, I'll fix that.

pulpcore/cli/container/repository.py
```diff
@@ -45,22 +53,66 @@
...
+                _("'{value}' is not a valid href for {option_name}.").format(
+                    value=value, option_name=param.name
+                )
+            )
+        pulp_href = value
+        if "versions" in value:
+            pulp_href = value.partition("versions")[0]
+            version = value.split("/")[-2]
+    else:
+        split_value = value.split(":", maxsplit=1)
```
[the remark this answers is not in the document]
gerrod3: I'm not sure, the serializer for this operation only allows container repositories to be specified, but doesn't have the same limit on the repository version field. So you couldn't supply a push repository, but you could supply one of its versions. I feel like that has to be an oversight.

pulpcore/cli/container/context.py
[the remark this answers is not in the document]
mdellweg: I was really thinking about `PulpRepositoryContext` now that you say that push-repos cannot modify content, this would add the need for a capability. Maybe not worth the trouble now.

pulpcore/cli/container/repository.py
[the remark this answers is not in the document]
mdellweg: Yes, i think you must be able to push content to a push repo and subsequently copy it to a "normal" one.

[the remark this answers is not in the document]
gerrod3: Ok, added support for copying from push repositories.

pulpcore/cli/container/repository.py
```diff
@@ -165,3 +247,77 @@ def add_tag(
...
+
+
+@repository.command(allowed_with_contexts=container_context)
+@name_option
+@href_option
+@click.option(
+    "--source",
+    help=_(
+        "Source repository to copy manifests from. Specify in format of href or [<resource_type>:]"
+        "<name>[:<version>]"
```
mdellweg, line 286: :dragon: optional parts in two places sounds real dangerous. Also i had expected that wherever we have a resource parameter, we'd allow for `[[<plugin>:]<type>:]<name>`.
Also we need to make sure we still allow ":" in the name of the repository with `::reponame:with:colons`.
Is it very unintuitive to use a separate separator for the version?

mdellweg (approved):
I think, this is it.

PR #436, opened 2021-12-02
Add workaround for change in openapi for Group

mdellweg:
In pulpcore 3.17 the Group model is proxied into the core app. This
introduces subtle changes in the openapi docs.

[noissue]

This is required to fix the change introduced by
https://github.com/pulp/pulpcore/pull/1731

### DIFF
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -6,7 +6,12 @@ from typing import IO, Any, ClassVar, Dict, List, Optional
 
 import click
 
-from pulpcore.cli.common.context import EntityDefinition, PulpContext, PulpEntityContext
+from pulpcore.cli.common.context import (
+    EntityDefinition,
+    PluginRequirement,
+    PulpContext,
+    PulpEntityContext,
+)
 
 _ = gettext.gettext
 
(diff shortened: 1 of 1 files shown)


pulpcore/cli/core/context.py
```diff
@@ -112,13 +117,20 @@ def scope(self) -> Dict[str, Any]:
 class PulpGroupContext(PulpEntityContext):
     ENTITY = _("user group")
     ENTITIES = _("user groups")
-    HREF = "auth_group_href"
+    # Handled by a workaround
+    # HREF = "group_href"
```
bmbouter, line 121: I'm not a cli person so feel free to ignore. Should this stay in the final PR?

gerrod3 (commented):
Looks good, I'll approve once the pulpcore pr is merged.

pulpcore/cli/core/context.py
mdellweg: Yes, because someone may miss this line. Also its supposed to be easy to remove the workaround again eventually.

mdellweg:
Admin merging this to unbreak the pulpcore CI

PR #438, opened 2021-12-09
Add reset command to access-policy

mdellweg:
fixes #tba


CHANGES/438.feature
```diff
@@ -0,0 +1,2 @@
+Added `reset` command to access_policies.
+Changed `--permissions-assignment` to `--creation-hooks` to follow pulpcore 3.17 naming.
```
gerrod3, line 2: pulp-cli uses a different docs building process that doesn't need double marks right?

gerrod3 (approved):
Looks good.

CHANGES/438.feature
gerrod3: Also, do we not require issues be filed for new features? I realized that the changelog is the PR number itself.

mdellweg: Yes, It's Markdown, not RST.

PR #444, opened 2022-01-10
Use ID_PREFIX for operation_ids

mdellweg:
This probably needs a changelog.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,11 @@ name = "Translations"
 showcontent = true
 
 [[tool.towncrier.type]]
+directory = "devel"
+name = "Developer Notes"
+showcontent = true
+
+[[tool.towncrier.type]]
 directory = "misc"
 name = "Misc"
 showcontent = false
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -32,6 +32,11 @@ directory = "translation"
 name = "Translations"
 showcontent = true
 
+[[tool.towncrier.type]]
+directory = "api"
+name = "API changes"
+showcontent = true
+
```
mdellweg, line 39: I was tempted to put the changelog in "misc", but that is supposed to be hidden. Also i don't want to mix those with user facing changes.

ggainey: The problem I have with 'api' is when I hear it, I assume something that *is* user-facing. I'm not sure what a better word would be, to describe "changes to CLI toolkit/design that make it easier to add new commands", tho :(

ggainey (approved):
If we can think of a towncrier section that's better than 'api', I'd like to do that. But the code/approach I like, *a lot*, and looks fine!

PR #448, opened 2022-01-16
Make broad use of PREFIX_ID

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -508,11 +508,15 @@ class PulpEntityContext:
         uploads: Optional[Dict[str, Any]] = None,
         non_blocking: bool = False,
     ) -> Any:
+        # Workaround for plugins that do not have ID_PREFIX in place
+        if not hasattr(self, "ID_PREFIX") and not hasattr(self, "PARTIAL_UPDATE_ID"):
+            self.PARTIAL_UPDATE_ID = getattr(self, "UPDATE_ID")
+        # ----------------------------------------------------------
         _parameters = {self.HREF: href}
         if parameters:
             _parameters.update(parameters)
         return self.call(
-            "update",
+            "partial_update",
             parameters=_parameters,
             body=body,
             uploads=uploads,
--- a/pulpcore/cli/common/debug.py
+++ b/pulpcore/cli/common/debug.py
@@ -42,14 +42,7 @@ def task_summary(pulp_ctx: PulpContext) -> None:
     """
     from pulpcore.cli.core.context import PulpTaskContext
 
-    TASK_STATES = ["waiting", "skipped", "running", "completed", "failed", "canceled"]
-    if pulp_ctx.has_plugin(PluginRequirement("core", min="3.14")):
-        TASK_STATES.append("canceling")
-    result = {}
-    for state in TASK_STATES:
-        payload = {"limit": 1, "state": state}
-        answer = pulp_ctx.call(PulpTaskContext.LIST_ID, parameters=payload)
-        result[state] = answer["count"]
+    result = PulpTaskContext(pulp_ctx).summary()
     pulp_ctx.output_result(result)
 
 
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -21,13 +21,10 @@ class PulpAccessPolicyContext(PulpEntityContext):
     ENTITY = _("access policy")
     ENTITIES = _("access policies")
     HREF = "access_policy_href"
-    LIST_ID = "access_policies_list"
-    READ_ID = "access_policies_read"
-    UPDATE_ID = "access_policies_partial_update"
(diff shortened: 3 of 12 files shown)


pulpcore/cli/core/context.py
```diff
@@ -379,24 +321,20 @@ class PulpTaskGroupContext(PulpEntityContext):
...
     HREF = "upload_href"
-    LIST_ID = "uploads_list"
-    READ_ID = "uploads_read"
-    CREATE_ID = "uploads_create"
-    UPDATE_ID = "uploads_update"
-    DELETE_ID = "uploads_delete"
-    COMMIT_ID: ClassVar[str] = "uploads_commit"
+    ID_PREFIX = "uploads"
+    # TODO: Not so nice
+    PARTIAL_UPDATE_ID = "uploads_update"
```
mdellweg, line 333: This works, but i am unsure whether it's a sign that we should not use `update()`, but implement a `upload_chunk()` on this context.

pulpcore/cli/common/debug.py
```diff
@@ -48,7 +48,7 @@ def task_summary(pulp_ctx: PulpContext) -> None:
     result = {}
     for state in TASK_STATES:
         payload = {"limit": 1, "state": state}
-        answer = pulp_ctx.call(PulpTaskContext.LIST_ID, parameters=payload)
+        answer = PulpTaskContext(pulp_ctx).call("list", parameters=payload)
```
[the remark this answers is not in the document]
mdellweg: `def summary(self)` it is.

pulpcore/cli/core/context.py
mdellweg: `upload_chunk` it is.

pulpcore/cli/core/context.py
```diff
@@ -369,34 +308,56 @@ def purge(self, finished_before: Optional[str], states: Optional[List[str]]) ->
...
         )
 
+    def summary(self) -> Dict[str, int]:
+        task_states = ["waiting", "skipped", "running", "completed", "failed", "canceled"]
+        if self.pulp_ctx.has_plugin(PluginRequirement("core", min="3.14")):
+            task_states.append("canceling")
+        result = {}
+        for state in task_states:
+            payload = {"limit": 1, "state": state}
+            result[state] = self.call("list", parameters=payload)["count"]
```
ggainey, line 323: Oh, *that* is clever - :100:

pyproject.toml
```diff
@@ -51,7 +51,8 @@ line_length = 100
 
 [tool.pytest.ini_options]
 markers = [
-  "script: Marks tests provided as shell scripts",
+  "script: tests provided as shell scripts",
+  "help_page: tests that render help pages",
```
ggainey, line 55: Does this belong in this PR? Or should it be its own thing?

ggainey (commented):
A question for clarity: will plugins you can't refactor here (e.g., deb-cli) continue working as-is? Or does this break them?

pyproject.toml
mdellweg: It's not enough for me to bother making a separate PR for it...

pulpcore/cli/common/context.py
```diff
@@ -512,7 +512,7 @@ def update(
         if parameters:
             _parameters.update(parameters)
         return self.call(
-            "update",
+            "partial_update",
```
mdellweg, line 515: This is the only change with a risk to break anyone. But i don't suspect anyone to use `update` aka `PUT` for any good reason.
(Yes our upload api uses it to create chunks. You'd kind of expect a POST call for that, no? And that weird case is covered explicitely in this PR.)
...is what i wanted to say. But thinking again, this is breaking all contexts using UPDATE_ID today and defining neither PARTIAL_UPDATE_ID or nor ID_PREFIX.

gerrod3: I think that is a fine assumption to make. I looked at the other two plugins (deb & ostree) and they both use partial_update, so I think we are good.

mdellweg: I added a workaround. Please look again. @ggainey

gerrod3: Yes, you are right. I had to go look at #444 to understand how this works.

pulpcore/cli/common/context.py
```diff
@@ -508,11 +508,15 @@ def update(
         uploads: Optional[Dict[str, Any]] = None,
         non_blocking: bool = False,
     ) -> Any:
+        # Workaround for plugins that do not have ID_PREFIX in place
```
ggainey, line 511: :+1:

ggainey (approved):
"Assuming" that we know the current state of things no longer breaks pulp_deb, looks good to me!

gerrod3 (approved):
Looks good. Linter is complaining about the workaround, but after that's fixed LGTM.

PR #456, opened 2022-01-26
Add api_path to connect to rerouted pulp

### LINKED ISSUE #453: As a user, I can have an option to interact with a rerooted Pulp
## Summary
The ability to reroot Pulp is tracked here. https://github.com/pulp/pulpcore/issues/2148

It would be great if a setting could be added that would apply to all commands when set to use a different URL instead of `/pulp/api/v3`.

mdellweg:
fixes #453

### DIFF
--- a/pulpcore/cli/core/task.py
+++ b/pulpcore/cli/core/task.py
@@ -35,7 +35,7 @@ def _uuid_callback(
     if value is not None:
         entity_ctx = ctx.find_object(PulpEntityContext)
         assert entity_ctx is not None
-        entity_ctx.pulp_href = f"/pulp/api/v3/tasks/{value}/"
+        entity_ctx.pulp_href = f"{entity_ctx.pulp_ctx.api_path}tasks/{value}/"
     return value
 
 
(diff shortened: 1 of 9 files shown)


mdellweg:
The tests still use hardcoded paths. I am wondering if we can start the container with a rerouted api...

mdellweg:
The tests still use hardcoded paths. I am wondering if we can start the container with a rerouted api...

pulpcore/cli/core/task.py
```diff
@@ -125,7 +125,7 @@ def cancel(
         try:
             pulp_ctx.wait_for_task(entity)
         except Exception as e:
-            if not re.match("Task /pulp/api/v3/tasks/[-0-9a-f]*/ canceled", str(e)):
+            if not re.match(rf"Task {pulp_ctx.api_path}tasks/[-0-9a-f]*/ canceled", str(e)):
```
gerrod3, line 128: Neat! I did not know you could combine r & f strings.

gerrod3 (approved):
Looks good, we can add tests in a seprarate PR or wait for @bmbouter's reroute work to be complete.

pulpcore/cli/core/task.py
mdellweg: Yep, i learned that too.

mdellweg:
We'd need a rerouted container to really test this.

PR #461, opened 2022-02-01
Fix error reporting on tasks

mdellweg:
If a task fails not due to an Exception, we do not get the description field.
Try reason filed instead.


ggainey (changes requested):
Nice - do we have an issue for this? Can we attach one?

mdellweg:
@ggainey Ticket referenced. Please merge when approved.

ggainey (approved):
lg2m!

PR #463, opened 2022-02-02
Fetch the config profile from environment

mdellweg:
[noissue]


pulpcore/cli/common/__init__.py
```diff
@@ -80,6 +80,7 @@ def _config_callback(ctx: click.Context, param: Any, value: Optional[str]) -> No
 @click.option(
     "--profile",
     "-p",
+    envvar="PULP_PROFILE",
```
gerrod3, line 83: What do you think about "PULP_CLI_PROFILE"?

mdellweg: +1

gerrod3 (approved):
Nice!

PR #467, opened 2022-02-03
Add needs_capability to EntityContext

### LINKED ISSUE #465: Refactor: add needs_capability to context
## Summary

We should have a needs_capability that raises itself.
Rationale: All mentions of Pulp versions should be in context.py.

See attached PR for an example of what we could move into context

mdellweg:
This also makes use of it for tagging and untagging in container
repositories as well as for all roles related subcommands.

fixes #465


ggainey (approved):
I really like how much cleaner/more-obvious the code becomes with this change!

PR #468, opened 2022-02-04
Container roles

mdellweg:
We probably want _some_ tests for this...

depends on #465


ggainey (changes requested):
I would like to see tests for this before we merge it, please!

ggainey (approved):
Thanks for the tests - lg2m!

PR #469, opened 2022-02-04
Test Pulp against a rerouted container

### LINKED ISSUE #465: Refactor: add needs_capability to context

mdellweg:


mdellweg:
Hmmm ... docker with, work it does not.

ipanova:
> Hmmm ... docker with, work it does not.

what docker say does?

mdellweg:
> > Hmmm ... docker with, work it does not.
> 
> what docker say does?

My `--env` argument ignoring docker is.

fao89 (approved):
LGTM

PR #471, opened 2022-02-08
Add default help text for options taking a generic resource argument.

### LINKED ISSUE #387: Improve help text for options accepting generic resources
## Summary
The `pulp rpm distribution create` command has a `--repository` option. This options takes a generic resource as argument, i.e., `[[<plugin>:]<resource_type>:]<name>'` (albeit it would default to, and only accept `rpm:rpm:`). [NB: it does not accept a href either, not sure whether that could be added easily.]

Other commands however, e.g., `pulp rpm repository content add` only take the bare name of a repository as argument for `--repository`.

This seems to be a bit inconsistent and/or confusing.

If we do want to keep it the way it is, we should at least improve help texts.

Would a patch similar to this be acceptable?
```patch
diff --git a/pulpcore/cli/common/generic.py b/pulpcore/cli/common/generic.py
index 4271185..deef36a 100644
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -361,6 +361,13 @@ def resource_option(*args: Any, **kwargs: Any) -> Callable[[_FC], _FC]:
         kwargs["callback"] = _multi_option_callback
     else:
         kwargs["callback"] = _option_callback
+
+    if "help" not in kwargs:
+        kwargs["help"] = _(
+            "Referenced resource, in the form [[<plugin>:]<resource_type>:]<name>. "
+            "'<plugin>' defaults to '{plugin}', '<resource> defaults to {resource_type}."
+        ).format(plugin=default_plugin or "", resource_type=default_type or "")
+
     return click.option(*args, **kwargs)
 
 

```
Still need to refine the help text in case the command in question accepts a href.

## pulp-cli version info
pulp-cli 0.11.0-31-g591d6a6

thmo:
fixes #387


pulpcore/cli/common/generic.py
```diff
@@ -457,6 +457,22 @@ def _multi_option_callback(
...
     else:
         kwargs["callback"] = _option_callback
+
+    if "help" not in kwargs:
+        kwargs["help"] = _(
+            "Referenced resource, in the form {plugin_form}{type_form}<name> or by href. "
+            "{plugin_default}{type_default}"
+        ).format(
+            plugin_form=_("[<plugin>:]") if default_plugin else _("<plugin>"),
+            type_form=_("[<resource_type>:]") if default_type else _("<resource_type>"),
```
mdellweg, line 467: I believe we still need the colons:
```suggestion
            plugin_form=_("[<plugin>:]") if default_plugin else _("<plugin>:"),
            type_form=_("[<resource_type>:]") if default_type else _("<resource_type>:"),
```

mdellweg (commented):
This is a great thing!

pulpcore/cli/common/generic.py
thmo: Good catch - I have updated the commit.

PR #474, opened 2022-02-12
Add a missing sub-command

lubosmj:
[noissue]


mdellweg (approved):
In my version of towncrier, build is marked as the default sub command. But including it here eases the use of `--help`. So Ack!

PR #480, opened 2022-02-25
WIP Add a list-content command to reository versions

mdellweg:
[noissue]


mdellweg:
we already have something very similar `pulp repository content list --version --type`.

PR #483, opened 2022-03-10
Add ansible signature commands

gerrod3:
[noissue]

Implemented the signature task command. Should I just put the collection signature list/read/create commands under the ansible content commands or create its own sub-group?

TODO:

- [x] Add issues
- [x] Signature list/read/create
- [x] Tests
- [x] Update capabilities doc

### DIFF
--- a/.ci/container_setup.d/10-install-signing-service.sh
+++ b/.ci/container_setup.d/10-install-signing-service.sh
@@ -2,14 +2,32 @@
 
 set -eu
 
-if pulp --format none --refresh-api --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "deb"
+pulp --refresh-api --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "deb" && HAS_DEB=true || HAS_DEB=""
+pulp --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "ansible" && HAS_ANSIBLE=true || HAS_ANSIBLE=""
+if [ "$HAS_DEB" ] || [ "$HAS_ANSIBLE" ]
 then
-  echo "Setup a signing service"
-  "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" bash -c "cat > /root/sign_deb_release.sh" < "${BASEPATH}/../tests/assets/sign_deb_release.sh"
-  "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" bash -c "cat > /tmp/setup_signing_service.py" < "${BASEPATH}/../tests/assets/setup_signing_service.py"
+  echo "Setup the signing services"
+  # Setup key on the Pulp container
   curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-KEY-pulp-qe | "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" bash -c "cat > /tmp/GPG-KEY-pulp-qe"
   curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-PRIVATE-KEY-pulp-qe | "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" gpg --import
   echo "6EDF301256480B9B801EBA3D05A5E6DA269D9D98:6:" | "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" gpg --import-ownertrust
-  "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" chmod a+x /root/sign_deb_release.sh /tmp/setup_signing_service.py
-  "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" /tmp/setup_signing_service.py /root/sign_deb_release.sh /tmp/GPG-KEY-pulp-qe
+  # Setup key on the test machine
+  curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-KEY-pulp-qe | cat > /tmp/GPG-KEY-pulp-qe
+  curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-PRIVATE-KEY-pulp-qe | gpg --import
+  echo "6EDF301256480B9B801EBA3D05A5E6DA269D9D98:6:" | gpg --import-ownertrust
+  if [ "$HAS_DEB" ]
+  then
+    echo "Setup deb release signing service"
+    "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" bash -c "cat > /root/sign_deb_release.sh" < "${BASEPATH}/../tests/assets/sign_deb_release.sh"
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" chmod a+x /root/sign_deb_release.sh
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" bash -c "pulpcore-manager add-signing-service --class deb:AptReleaseSigningService sign_deb_release /root/sign_deb_release.sh 'Pulp QE'"
+  fi
+  if [ "$HAS_ANSIBLE" ]
+  then
+    echo "Setup ansible signing service"
+    "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" bash -c "cat > /root/sign_detached.sh" < "${BASEPATH}/../tests/assets/sign_detached.sh"
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" chmod a+x /root/sign_detached.sh
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" bash -c "pulpcore-manager add-signing-service --class core:AsciiArmoredDetachedSigningService sign_ansible /root/sign_detached.sh 'Pulp QE'"
+  fi
 fi
+
--- a/pulpcore/cli/ansible/content.py
+++ b/pulpcore/cli/ansible/content.py
@@ -2,7 +2,11 @@ from typing import IO, Any, Union
 
 import click
 
-from pulpcore.cli.ansible.context import PulpAnsibleCollectionVersionContext, PulpAnsibleRoleContext
+from pulpcore.cli.ansible.context import (
+    PulpAnsibleCollectionVersionContext,
+    PulpAnsibleCollectionVersionSignatureContext,
+    PulpAnsibleRoleContext,
+)
 from pulpcore.cli.common.context import (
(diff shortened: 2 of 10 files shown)


pulpcore/cli/ansible/repository.py
```diff
@@ -211,3 +222,21 @@ def sync(
         href=repository_href,
         body=body,
     )
+
+@repository.command()
+@name_option
+@href_option
+@click.option("--signing-service", required=True, callback=_signing_service_callback)
+@click.option("--content-units", callback=load_json_callback)
```
gerrod3, line 230: Is there a better way to add these options? We have so many helper functions that I'm starting to forget what they can do...

mdellweg: I think, they are just fine here.

mdellweg:
> Should I just put the collection signature list/read/create commands under the ansible content commands or create its own sub-group?

Well from a pulp view, signatures are just content. I have no better answer to reason another place.

tests/assets/setup_signing_service.py
```diff
@@ -1,46 +1,54 @@
 #!/usr/bin/env python3
```
mdellweg, line 1: FYI, yes we have a manage command to setup signing services, but this script still hangs around for the versions of pulpcore that predate it (< 3.13).
Actually we don't test those anymore...

pulpcore/cli/ansible/content.py
```diff
@@ -11,7 +15,6 @@
 )
 from pulpcore.cli.common.generic import (
     GroupOption,
-    chunk_size_option,
```
mdellweg, line 14: Please explain...

gerrod3: The collection upload wasn't using it (like it actually didn't do anything even if you specified it).  I can change this to use `allow_with_context` to better reflect its use only with roles.

.ci/container_setup.d/10-install-signing-service.sh
```diff
@@ -2,14 +2,25 @@
 
 set -eu
 
-if pulp --format none --refresh-api --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "deb"
+has_deb=$(pulp --format none --refresh-api --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "deb")
+has_ansible=$(pulp --format none --refresh-api --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "ansible")
```
mdellweg, line 6: Not sure this will work. You are interested in the exit code of these commands, not their stdout.

.ci/container_setup.d/10-install-signing-service.sh
```diff
@@ -2,14 +2,25 @@
...
+  then
+    "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" bash -c "cat > /root/sign_deb_release.sh" < "${BASEPATH}/../tests/assets/sign_deb_release.sh"
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" chmod a+x /root/sign_deb_release.sh
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" bash -c "pulpcore-manager add-signing-service --class deb:AptReleaseSigningService sign_deb_release /root/sign_deb_release.sh 'Pulp QE'"
+  fi
+  if $has_ansible
+  then
+    "${CONTAINER_RUNTIME}" exec -i "pulp-ephemeral" bash -c "cat > /root/sign_ansible.sh" < "${BASEPATH}/../tests/assets/sign_ansible.sh"
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" chmod a+x /root/sign_ansible.sh
+    "${CONTAINER_RUNTIME}" exec "pulp-ephemeral" bash -c "pulpcore-manager add-signing-service --class core:AsciiArmoredDetachedSigningService sign_ansible /root/sign_ansible.sh 'Pulp QE'"
```
mdellweg, line 23: so pleased by this...

pulpcore/cli/ansible/content.py
```diff
@@ -130,22 +175,35 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
...
+        PulpAnsibleCollectionVersionContext,
+        PulpAnsibleCollectionVersionSignatureContext,
+    ],
     file: IO[bytes],
-    chunk_size: int,
     **kwargs: Any,
 ) -> None:
 
     if isinstance(content_ctx, PulpAnsibleRoleContext):
-        artifact_href = PulpArtifactContext(pulp_ctx).upload(file, chunk_size)
```
mdellweg, line 144: Are you saying this didn't work?

pulpcore/cli/ansible/content.py
```diff
@@ -130,22 +175,35 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
...
         body = {"artifact": artifact_href}
         body.update(kwargs)
         result = content_ctx.create(body=body)
         pulp_ctx.output_result(result)
+    elif isinstance(content_ctx, PulpAnsibleCollectionVersionSignatureContext):
+        body = {"signed_collection": kwargs.get("collection")}
+        uploads = {"file": file.read()}
+        pulp_ctx.output_result(content_ctx.create(body=body, uploads=uploads))
     else:
         result = content_ctx.upload(file=file)
```
mdellweg, line 208: Now this is the case for `PulpAnsibleCollectionVersion`, maybe make that more explicit with elif and raise an NotImplemented in the final else.
Also i think, collections can get arbitrarily large. We should support chunked upload (probably in a separate PR).

.ci/container_setup.d/10-install-signing-service.sh
gerrod3: You are correct, I removed the format option so now true or false is stored in the variables.

pulpcore/cli/ansible/content.py
gerrod3: It did work, I'll change it back.

pulpcore/cli/ansible/content.py
gerrod3: +1 for being more explicit, and yes for supporting chunked upload. Going to need a new option on the upload endpoint in pulp_ansible as currently it only supports file uploads.

.ci/container_setup.d/10-install-signing-service.sh
```diff
@@ -2,14 +2,32 @@
 
 set -eu
 
-if pulp --format none --refresh-api --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "deb"
+has_deb=$(pulp --refresh-api --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "deb")
+has_ansible=$(pulp --base-url "http://localhost:8080" ${PULP_API_ROOT:+--api-root "${PULP_API_ROOT}"} --username "admin" --password "password" debug has-plugin --name "ansible")
+if $has_deb || $has_ansible
```
mdellweg, line 7: Now you are executing the output of these commands. I see why it works, but my instinct whines "this is unsafe".
What if you do
`pulp ... has-plugin ... && HAS_DEB=true || HAS_DEB=""`
and then check with
`if [ "$HAS_DEB" ]; then`
?

pulpcore/cli/ansible/content.py
```diff
@@ -114,7 +161,13 @@ def content(ctx: click.Context, pulp_ctx: PulpContext, content_type: str) -> Non
 
 @content.command()
 @click.option("--file", type=click.File("rb"), required=True)
-@chunk_size_option
+@pulp_option(
+    "--chunk-size",
+    help=_("Chunk size to break up {entity} into. Defaults to 1MB"),
+    default="1MB",
+    callback=parse_size_callback,
+    allowed_with_contexts=role_context,
```
mdellweg, line 169: :+1: I was about to ask why we need to redefine it...

tests/assets/sign_ansible.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+FILE_PATH=$1
+SIGNATURE_PATH="$1.asc"
+
+GPG_KEY_ID="Pulp QE"
```
mdellweg, line 6: Don't we also get the key id passed down from the signing service?
I'm wondering if we could write a generic script for at least simple gpg signing.

.ci/container_setup.d/10-install-signing-service.sh
gerrod3: I tested these commands with a custom script, and it properly knows the difference between true and false. I can change it if you really want as I'm not an expert in bash scripting.

tests/assets/sign_ansible.sh
gerrod3: Yes, we can. (I copied this from pulp_ansible) I'll change it to be generic to allow for more signing-services to be created.

.ci/container_setup.d/10-install-signing-service.sh
mdellweg: Yes, i know, why it works. `true` and `false` are standard unix commands. And we call the content of the variable here which just so happens to be either of them. This is on the same level as `eval` in python (for some reason very close to "evil"). And to me that is a anti pattern. So yes, i'd like it to be changed.
`[` on the other hand is an alias for the bash builtin `test` so we only pass the content of the variable as a parameter.

tests/assets/sign_ansible.sh
mdellweg: Well actually i was thinking that we should ship the very basic script with pulpcore at some point.

tests/assets/sign_detached.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+FILE_PATH=$1
+SIGNATURE_PATH="$1.asc"
+: "${PULP_SIGNING_KEY_FINGERPRINT:="Pulp QE"}"
+
+# Create a detached signature
+gpg --quiet --batch --homedir ~/.gnupg/ --detach-sign --default-key "${PULP_SIGNING_KEY_FINGERPRINT}" \
```
mdellweg, line 8: Yes, yes, yes!!!
Can you maybe take a tour and update the docs for signing services everywhere to use that style?

tests/assets/sign_detached.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+FILE_PATH=$1
+SIGNATURE_PATH="$1.asc"
+: "${PULP_SIGNING_KEY_FINGERPRINT:="Pulp QE"}"
```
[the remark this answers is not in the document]
gerrod3: Yes.

mdellweg (approved):

Issue #470, opened 2022-02-07
Add support for ULN remotes


quba42:
## Summary
Currently the CLI does not know how to handle ULN remotes.
See: https://discourse.pulpproject.org/t/unable-to-sync-with-oracle-uln/357

It should be fairly simple to mirror the code for the RPM remote for ULN remotes as well, while also making sure the CLI accepts all relevant ULN types.

PR #486, opened 2022-03-11
Add Support for ULN Remote in Pulp RPM

Manisha15:


pulpcore/cli/rpm/uln_remote.py
```diff
@@ -0,0 +1,56 @@
...
+    type=click.Choice(["rpm"], case_sensitive=False),
+    default="rpm",
+)
+@pass_pulp_context
+@click.pass_context
+def uln_remote(ctx: click.Context, pulp_ctx: PulpContext, remote_type: str) -> None:
+    if remote_type == "rpm":
+        ctx.obj = PulpUlnRemoteContext(pulp_ctx)
+    else:
+        raise NotImplementedError()
```
mdellweg, line 38: I would have expected, that we could add a new type to the `rpm remote` command instead.

Manisha15: So you mean `rpm remote uln <options>` ?

mdellweg: Almost: `pulp rpm remote --type uln <subcommands> <options>`

pulpcore/cli/rpm/remote.py
```diff
@@ -44,6 +46,7 @@ def remote(ctx: click.Context, pulp_ctx: PulpContext, remote_type: str) -> None:
         "--policy", type=click.Choice(["immediate", "on_demand", "streamed"], case_sensitive=False)
     ),
     click.option("--sles-auth-token"),
+    click.option("--uln-server-base-url", default="https://linux-update.oracle.com/")
```
mdellweg, line 49: This should be a `pulp_option`, with a help text and `allowed_with_contexts` specifier.

mdellweg (commented):
Can you please add tests for this?

mdellweg:
Also, we need a changelog snippet (can point to this PR instead of an issue).

pulpcore/cli/rpm/uln_remote.py
Manisha15: Fixed.

pulpcore/cli/rpm/remote.py
Manisha15: Added the tests and `pulp_option`.

pulpcore/cli/rpm/remote.py
```diff
@@ -44,6 +47,12 @@ def remote(ctx: click.Context, pulp_ctx: PulpContext, remote_type: str) -> None:
         "--policy", type=click.Choice(["immediate", "on_demand", "streamed"], case_sensitive=False)
     ),
     click.option("--sles-auth-token"),
```
mdellweg, line 49: I think this option needs to be scoped for the other context now.

tests/scripts/pulp_rpm/test_uln_remote.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
+
+pulp debug has-plugin --name "rpm" || exit 3
```
mdellweg, line 6: This probably needs a version specifier for the rpm plugin.

pulpcore/cli/rpm/context.py
```diff
@@ -102,6 +102,17 @@ class PulpRpmRemoteContext(PulpRemoteContext):
         "sles_auth_token",
     }
 
+class PulpUlnRemoteContext(PulpRemoteContext):
+    ENTITY = _("uln remote")
+    ENTITIES = _("uln remotes")
+    HREF = "rpm_uln_remote_href"
+    ID_PREFIX = "remotes_rpm_uln"
+    NULLABLES = {
```
mdellweg, line 110: Any reason not to extend the set of the base class here?

pulpcore/cli/rpm/context.py
```diff
@@ -102,6 +102,17 @@ class PulpRpmRemoteContext(PulpRemoteContext):
         "sles_auth_token",
     }
 
+class PulpUlnRemoteContext(PulpRemoteContext):
```
mdellweg, line 105: At some place i expect a `needs_plugin(PluginRequirement)` on the pulp_rpm version to introduce the uln remote. Probably in the `__init__` of this context.

tests/scripts/pulp_rpm/test_uln_remote.sh
Manisha15: Added the minimum version.

pulpcore/cli/rpm/context.py
Manisha15: Extended the RPm Remote base class and fixed Nullables as well.

pulpcore/cli/rpm/remote.py
Manisha15: Changed it to pulp_option to allow it with context.

pulpcore/cli/rpm/remote.py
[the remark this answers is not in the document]
Manisha15: Fixed.

pulpcore/cli/rpm/context.py
Manisha15: I added the requirement to remote function so that it only uses Uln when its above `3.12.0` version. https://github.com/ATIX-AG/pulp-cli/blob/0f2df1610202da5fe8274d10c1bacb901797a7cd/pulpcore/cli/rpm/remote.py#L39

pulpcore/cli/rpm/context.py
mdellweg: OK, sorry i didn't express this properly. We can still safely inherit from `PulpRemoteContext`, while we would define `NULLABLES = PulpRemoteContext.NULLABLES + {"uln_server-base_url"}`.

Manisha15: But the question is, can we have nullable `"uln_server-base_url"`? In settings we set to default value and although its not a required field but we cannot set it to Null. Or am I understanding `Nullables` wrong?

mdellweg: OK that i don't know. `NULLABLES` is basically to translate the empty string provided if you specify `--opt ""` to `None` for submitting to the api.
Whether `uln_server_base_url` can be "unset" is a question for someone else. If it is not, i'd say maybe you don't need to redefine `NULLABLES` at all...
Also think about the update command. Maybe you need to provide a version of that option without default, or it will unintentionally be overwritten.

pulpcore/cli/rpm/context.py
```diff
@@ -103,6 +103,14 @@ class PulpRpmRemoteContext(PulpRemoteContext):
     }
 
 
+class PulpUlnRemoteContext(PulpRemoteContext):
+    ENTITY = _("uln remote")
+    ENTITIES = _("uln remotes")
+    HREF = "rpm_uln_remote_href"
+    ID_PREFIX = "remotes_rpm_uln"
+    NULLABLES = PulpRpmRemoteContext.NULLABLES + {"uln_server-base_url"}
```
mdellweg, line 111: Pending on the outcome of the other discussion, but we should either not have this line or use that version:
```suggestion
    NULLABLES = PulpRemoteContext.NULLABLES + {"uln_server-base_url"}
```

mdellweg:
You still need to spec `fixes #470` in the commit message.

mdellweg:
@ggainey can i kindly ask you to review this work for rpm-ness?

tests/scripts/pulp_rpm/test_uln_remote.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
ggainey, line 1: The tests will fail unless/until we add ULN_REMOTE_URL, user, and pass, to project-secrets. As a project, pulp doesn't have ULN access - and it wouldn't be appropriate to use a contributor's access to drive our CI :)

I'd like to see the test a) rely on env-vars for user/pass, b) some brief commentary on "you need to set variables ULN_REMOTE_URL, USER, and PASSWORD before this test will work, and c) at the top of the test, if url/user/poass are NOT set, just return true. That'll let us keep the test to run locally when/as we can, but won't hold up CI.

ggainey (changes requested):
Looks great! If we can get the requested changes to the test, I think we're good to go.

tests/scripts/pulp_rpm/test_uln_remote.sh
mdellweg: I think, we made `exit 3` let the test appear as skipped.

mdellweg: Also i do not see the tests _use_ the remote, so i think just CRUD on the remote we can do without a problem.

quba42 (changes requested):
I found one issue while trying to locally test these changes, which is that the `--remote` option will not accept a ULN remote for either syncing an RPM repo, or for associating a RPM repo with a ULN remote. My test workflow goes as follows:
```
ENTITIES_NAME='test'
REMOTE_OPTIONS='--url=uln://ovm2_2.1.1_i386_patch --username=redacted --password=redacted'
pulp rpm remote --type=uln create --name=${ENTITIES_NAME} ${REMOTE_OPTIONS}
pulp rpm repository create --name=${ENTITIES_NAME} --remote=${ENTITIES_NAME}
or
pulp rpm repository create --name=${ENTITIES_NAME}
pulp rpm repository sync --name=${ENTITIES_NAME} --remote=${ENTITIES_NAME}
```
Both these workflows result in the following error:
```
Error: Could not find rpm remote with {'name': 'test'}.
```

Another less serious issue, is if we could provide some help text, for the various `--url` options used by the subcommands of ULN remote? Right now it is:
```
$ pulp rpm remote --type=uln create --help
Usage: pulp rpm remote create [OPTIONS]

  Create a uln remote.

Options:
  --name TEXT                     [required]
  --url TEXT                      [required]
```
But since for ULN remotes `--url` requires the ULN channel name e.g.: `uln://ovm2_2.1.1_i386_patch` (which is not very intuitive) some help text like the following would be great:
```
  --url TEXT                      [required] Use the ULN channel name starting with uln:// here.
```
Alternatively (perhaps even better), the CLI might check that any user supplied `--url` parameters start with `uln://`.

mdellweg:
> I found one issue while trying to locally test these changes, which is that the `--remote` option will not accept a ULN remote for either syncing an RPM repo, or for associating a RPM repo with a ULN remote. My test workflow goes as follows:
> 
> ```
> ENTITIES_NAME='test'
> REMOTE_OPTIONS='--url=uln://ovm2_2.1.1_i386_patch --username=redacted --password=redacted'
> pulp rpm remote --type=uln create --name=${ENTITIES_NAME} ${REMOTE_OPTIONS}
> pulp rpm repository create --name=${ENTITIES_NAME} --remote=${ENTITIES_NAME}
> or
> pulp rpm repository create --name=${ENTITIES_NAME}
> pulp rpm repository sync --name=${ENTITIES_NAME} --remote=${ENTITIES_NAME}
> ```
> 
> Both these workflows result in the following error:
> 
> ```
> Error: Could not find rpm remote with {'name': 'test'}.
> ```

I think, what you need to use is `--remote uln:test` or `--remote rpm:uln:test`.
I agree this is not well documented, but the default plugin:type combination for these repositories is `rpm:rpm:...`.

> 
> Another less serious issue, is if we should provide some help text, for the various `--url` options used by the subcommands of ULN remote? Right now it is:
> 
> ```
> $ pulp rpm remote --type=uln create --help
> Usage: pulp rpm remote create [OPTIONS]
> 
>   Create a uln remote.
> 
> Options:
>   --name TEXT                     [required]
>   --url TEXT                      [required]
> ```
> 
> But since for ULN remotes `--url` requires the ULN channel name e.g.: `uln://ovm2_2.1.1_i386_patch` (which is not very intuitive) some help text like the following would be great:
> 
> ```
>   --url TEXT                      [required] Use the ULN channel name starting with uln:// here.
> ```
> 
> Alternatively (perhaps even better), the CLI might check that any user supplied `--url` parameters start with `uln://` for ULN remotes.

I'd say, let's go for both document and validate.

pulpcore/cli/rpm/remote.py
```diff
@@ -43,7 +53,23 @@ def remote(ctx: click.Context, pulp_ctx: PulpContext, remote_type: str) -> None:
...
+        allowed_with_contexts=(PulpRpmRemoteContext,),
+    ),
+    pulp_option(
+        "--uln-server-base-url",
+        default="https://linux-update.oracle.com/",
+        help=_("ULN Server base URL, default is 'https://linux-update.oracle.com/'"),
+        allowed_with_contexts=(PulpUlnRemoteContext,),
+    ),
+    pulp_option(
+        "--url",
```
Manisha15, line 67: I have added the check for uln `url` as well as help_text. The only problem here now is we have two `url` options/fields in `--help` of uln remote. Is there a way to remove the base class url option? 

```
Options:
  --name TEXT                     [required]
  --url TEXT                      [required]
  --ca-cert TEXT                  a PEM encoded CA certificate or @file
                                  containing same
  .
  .
  .
  --policy [immediate|on_demand|streamed]
  --uln-server-base-url TEXT      ULN Server base URL, default is
                                  'https://linux-update.oracle.com/'
  --url TEXT                      Use the ULN channel name starting with
                                  uln:// here.  [required]
  --help                          Show this message and exit.
```

mdellweg: Well, the other `--url` option comes via the `common_remote_{create,update}_options` which are plain lists.
So maybe we can create a filtered copy of these `[option for option in ... if option.name != "url"]`, and then add one `--url` option for the `RpmRemoteContext` and one for the `UlnRemoteContext`.
@gerrod3 may know whether they turn up correctly in the corresponding help outputs.

pulpcore/cli/rpm/remote.py
```diff
@@ -11,29 +11,39 @@
...
     update_command,
 )
 from pulpcore.cli.common.i18n import get_translation
-from pulpcore.cli.rpm.context import PulpRpmRemoteContext
+from pulpcore.cli.rpm.context import PulpRpmRemoteContext, PulpUlnRemoteContext
 
 translation = get_translation(__name__)
 _ = translation.gettext
 
+def _url_callback(ctx: click.Context, param: click.Parameter, value: str) -> str:
```
mdellweg, line 24: Should this be named `_uln_channel_callback` or `_uln_url_callback`?

pulpcore/cli/rpm/remote.py
Manisha15: The `common_remote_{create,update}_options` are lists of decorators that come from `click` so they don't have argument `name` or at least callable arguments.

pulpcore/cli/rpm/remote.py
Manisha15: I would suggest `_uln_url_callback` as the callback is for `url` option.

pulpcore/cli/rpm/remote.py
mdellweg: Then maybe it it impossible to use that shortcut here.

Manisha15: So where do you suggest to add this condition? Adding it to the base class doesn't look like a good idea.

pulpcore/cli/rpm/remote.py
Manisha15: Changed name from `_url_callback` to `_uln_url_callback`.

pulpcore/cli/rpm/remote.py
mdellweg: Yeah, it's not so much a condition. I think we need to specify the used parameters explicitly here. Not particularly happy about this, but probably the only way to get this forward now.

Manisha15: You mean to remove generic `common_remote_{create/update}_options` options and define them explicitly in `remote.py`?

mdellweg: Yes, this is the way forward i can see today. And i know it's bad for consistency.

pulpcore/cli/rpm/remote.py
```diff
@@ -43,14 +54,93 @@ def remote(ctx: click.Context, pulp_ctx: PulpContext, remote_type: str) -> None:
...
+    click.option("--password"),
+    click.option("--proxy-url"),
+    click.option("--proxy-username"),
+    click.option("--proxy-password"),
+    click.option("--rate-limit", type=int, help=_("limit download rate in requests per second")),
+    click.option("--sock-connect-timeout", type=float),
+    click.option("--sock-read-timeout", type=float),
+    click.option("--tls-validation", type=bool),
+    click.option("--total-timeout", type=float),
+    click.option("--username"),
```
mdellweg, line 136: Now you can merge the `rpm_remote_options` into these two lists with the proper values of `required`.

Manisha15: The options list will be repetitive. I can instead create list of  `unique` `remote_create_options`, `remote_update_options`, and another list which have common options for `create/update`. What do you think?

mdellweg: Sounds good.

Manisha15: Modified the options list.

pulpcore/cli/rpm/remote.py
```diff
@@ -1,56 +1,124 @@
...
+    click.option("--sock-connect-timeout", type=float),
+    click.option("--sock-read-timeout", type=float),
+    click.option("--tls-validation", type=bool),
+    click.option("--total-timeout", type=float),
+    pulp_option(
+        "--uln-server-base-url",
+        default="https://linux-update.oracle.com/",
+        help=_("ULN Server base URL, default is 'https://linux-update.oracle.com/'"),
+        allowed_with_contexts=(PulpUlnRemoteContext,),
+    ),
```
mdellweg, line 91: I think, this needs to be split for create and update, because providing a default with update will always set this value, and you don't want that in the update call on an existing customized remote.

Manisha15: Added the changes.

Manisha15:
Fixed the tests.

quba42 (approved):
I was now able to fully verify the functionality (using this branch rebased to the latest pulp-cli main branch) using the following workflow:
```
Requires adding "sha1" to the allowed checksums in /etc/pulp/settings.py:
ALLOWED_CONTENT_CHECKSUMS = ["sha224", "sha256", "sha384", "sha512", "sha1"]
pstop
pulpcore-manager handle-artifact-checksums
pstart


ENTITIES_NAME='test'
REMOTE_OPTIONS='--url=uln://ovm2_2.1.1_i386_patch --username=redacted --password=redacted'
pulp rpm remote --type=uln create --name=${ENTITIES_NAME} ${REMOTE_OPTIONS}
pulp rpm repository create --name=${ENTITIES_NAME} --remote=uln:${ENTITIES_NAME}
pulp rpm repository sync --name=${ENTITIES_NAME}
RPM_PUBLICATION_HREF=$(pulp rpm publication create --repository=${ENTITIES_NAME} | jq -r '.pulp_href')
pulp rpm distribution create --name=${ENTITIES_NAME} --base-path=${ENTITIES_NAME} --publication=${RPM_PUBLICATION_HREF}
```
I also checked that the CLI now throws an error if the `--url` parameter does not start with `uln://`.

The `--help` option now also contains more explanations, thanks!

From a usage point of view this looks good to go to me!

PR #488, opened 2022-03-21
Fix ansible ci with no signing service configured

mdellweg:
[noissue]

https://github.com/pulp/pulp_ansible/runs/5617249077?check_suite_focus=true


fao89 (approved):

mdellweg:
@gerrod3 FYI

PR #491, opened 2022-03-30
Lint with click~=8.0.0

mdellweg:
Click 8.1.0 introduced incompatible type annotations.


mdellweg:
Question: Is it wise to lint against one version and test against the other? Should we change some of the test runs to keep running against 8.0.0?

gerrod3 (approved):

ggainey (approved):

PR #498, opened 2022-04-17
Refactor PulpEntityContext

mdellweg:


pulpcore/cli/common/context.py
```diff
@@ -71,6 +66,12 @@ class PulpContext:
     It is an abstraction layer for api access and output handling.
     """
 
+    def echo(self, message: str, nl: bool = True, err: bool = False) -> None:
+        print(message)  # we can do better...
+
+    def prompt(self, text: str, hide_input: bool = False) -> Any:
```
gerrod3, line 72: Should these just be `NotImplementedError`?

gerrod3, line 72: Should these just be `NotImplementedError`?

gerrod3:
This work is for extracting out the shared logic between the CLI and ansible squeezer package?

pulpcore/cli/common/context.py
mdellweg: I dind't want to make it an abstract class.

ggainey (approved):
Looking good - thanks!

PR #507, opened 2022-05-18
RPM optional ``relative_path``

pavelpicka:
``relative_path`` for package become optional in RPM plugin.

[noissue]

### DIFF
--- a/pulpcore/cli/rpm/content.py
+++ b/pulpcore/cli/rpm/content.py
@@ -204,7 +204,6 @@ content.add_command(show_command(decorators=lookup_options))
 create_options = [
     pulp_option(
         "--relative-path",
-        required=True,
         allowed_with_contexts=(PulpRpmPackageContext,),
     ),
     pulp_option(
@@ -237,7 +236,6 @@ content.add_command(
 @chunk_size_option
 @pulp_option(
     "--relative-path",
-    required=True,
     help=_("Relative path within a distribution of the entity"),
     allowed_with_contexts=(PulpRpmPackageContext,),
 )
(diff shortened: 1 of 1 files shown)


mdellweg:
@ggainey , do you think we need to make this version aware?

ggainey:
> @ggainey , do you think we need to make this version aware?

def needs to be version (and PR) aware, since until https://github.com/pulp/pulp_rpm/pull/2537 is merged, it's absolutely required on upload.

mdellweg:
@pavelpicka and I just discussed the version check in the upload code. I do however have no good idea to add a corresponding one to the create call.

pavelpicka (commented):
This must check if this option is missing - that is possible from 3.18. Otherwise, we should raise an error. (Will update)

mdellweg:
> > @pavelpicka and I just discussed the version check in the upload code. I do however have no good idea to add a corresponding one to the create call.
> 
> What's the restriction on the help messages, they can't call the API right? I think you could set the `required` parameter to a `has_plugin` call in order to enforce the version checking. If that doesn't work we could update `PluginRequirement` and the `needs_plugin` check to include a `required`parameter for these cases.

Another idea that i had, was to pass the action performed to the preprocess_body, It could then "lint" for this check on "create". And also we would move the version dependent code into the contexts (a goal i have anyway).

ggainey:
@pavelpicka Hey Pavel - if you would rebase this, we will approve and merge it. Sorry we've let it sit for so long.

pulpcore/cli/rpm/content.py
```diff
@@ -272,6 +270,10 @@ def upload(
     body: Dict[str, Any] = {}
     uploads: Dict[str, IO[bytes]] = {}
     if isinstance(entity_ctx, PulpRpmPackageContext):
+        if kwargs.get("relative_path") is None:
+            pulp_ctx.needs_plugin(PluginRequirement("rpm", min="3.18"))
+        else:
+            ValueError("--relative-path must be provided")
```
mdellweg, line 276: I intend to move this into `preprocess_entity` once https://github.com/pulp/pulp-cli/pull/544 landed.

mdellweg (approved):

PR #508, opened 2022-05-27
Add a note about trimming

lubosmj:
[noissue]

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -655,10 +655,20 @@ common_remote_create_options = [
     click.option(
         "--download-concurrency", type=int, help=_("total number of simultaneous connections")
     ),
-    click.option("--password"),
+    click.option(
+        "--password",
+        help=_(
+            "The password to authenticate to the remote (can contain leading and trailing spaces)."
+        ),
+    ),
     click.option("--proxy-url"),
     click.option("--proxy-username"),
-    click.option("--proxy-password"),
+    click.option(
+        "--proxy-password",
+        help=_(
+            "The password to authenticate to the proxy (can contain leading and trailing spaces)."
+        ),
+    ),
     click.option("--rate-limit", type=int, help=_("limit download rate in requests per second")),
     click.option("--sock-connect-timeout", type=float),
     click.option("--sock-read-timeout", type=float),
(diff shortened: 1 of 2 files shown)


lubosmj:
This is a follow-up PR to https://github.com/pulp/pulpcore/pull/2767 to let end-users know that passwords are stored as they are provided. Thus, no additional operations like trimming are run against the provided passwords.

pulpcore/cli/common/generic.py
```diff
@@ -655,10 +655,22 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
     click.option(
         "--download-concurrency", type=int, help=_("total number of simultaneous connections")
     ),
-    click.option("--password"),
+    click.option(
+        "--password",
+        help=_(
+            "The password to authenticate to the proxy. "
```
ipanova, line 661: no proxy here

ipanova (dismissed):

pulpcore/cli/common/generic.py
```diff
@@ -655,10 +655,22 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
     click.option(
         "--download-concurrency", type=int, help=_("total number of simultaneous connections")
     ),
-    click.option("--password"),
+    click.option(
+        "--password",
+        help=_(
+            "The password to authenticate to the remote. "
+            "Extra leading and trailing whitespace characters are not being trimmed."
```
mdellweg, line 662: It's a bit strange to talk about something that is not being done, unless it were usually expected.
Maybe "The password including surrounding whitespace to authenticate with ..."

lubosmj: The `surrounding whitespace` may suggest adding whitespace characters to the original password. I would rather leave it unchanged. Or, I can altogether remove the help message because the trimming is not expected by default ...

mdellweg: The added complexity here is that you usually issue those commands in bash, and that _will_, if not quoted correctly, trim your passwords again. Also tabs and linebreaks in passwords are really problematic for any UI.
How about: "The password to authenticate to the remote (can contain leading and trailing spaces)."?

mdellweg (approved):

PR #510, opened 2022-06-07
Fix config file loading due to lack of permissions

### LINKED ISSUE #509: pulp-cli fails to start if a config file location is inaccessible (permissions)
## Summary
On freshly setup Pulp installation (using the official ansible installer), the /etc/pulp directory has mode
drwxr-x--- 3 root     pulp      4096 Jun  2 14:50 pulp

As /etc/pulp/cli.toml is the first location that is checked for the config file, this fails for all users besides root and pulp.

## Steps to reproduce
Try to use pulp-cli when the /etc/pulp directory is not readable for your user.

Maybe the fact that /etc/pulp is checked first is a problem on its own? Shouldn't this be a fallback, if no personal config file is found?

A fix is to not only catch FileNoFound exceptions but also PermissionExceptions when trying to load the config file.

zapp42:
Fix config file loading on systems where one of the config paths is not accessible due to lack of permissions. Fixes #509.

### DIFF
--- a/pulpcore/cli/common/__init__.py
+++ b/pulpcore/cli/common/__init__.py
@@ -57,6 +57,8 @@ def _config_callback(ctx: click.Context, param: Any, value: Optional[str]) -> No
                             config[key] = new_config[key]
                 except FileNotFoundError:
                     pass
+                except PermissionError:
+                    pass
         profile: str = "cli"
         if PROFILE_KEY in ctx.meta:
             profile = "cli-" + ctx.meta[PROFILE_KEY]


pulpcore/cli/common/__init__.py
```diff
@@ -57,6 +57,8 @@ def _config_callback(ctx: click.Context, param: Any, value: Optional[str]) -> No
                             config[key] = new_config[key]
                 except FileNotFoundError:
                     pass
+                except PermissionError:
+                    pass
```
mdellweg, line 61:
```suggestion
                except (FileNotFoundError, PermissionError):
                    pass
```
I think, that's the proper syntax. Also we could consider using https://docs.python.org/3/library/contextlib.html#contextlib.suppress .

mdellweg:
This needs a changelog entry. Can you drop a corresponding file in CHANGES please?

mdellweg:
I took the liberty and completed your PR in #533 .
Thank you anyway for your contribution in filing, analyzing and fixing the issue.

PR #512, opened 2022-06-12
Added content redirect content guard

mdellweg:


gerrod3 (approved):
So nice to see how easy it is to add new object commands.

PR #514, opened 2022-06-18
Use upload to create big file content

mdellweg:
[noissue]


gerrod3 (approved):
Looks good, do we need a test for this?

mdellweg:
It will be tested already. The beauty of refactoring is that you do not change tests. ;)

PR #516, opened 2022-06-24
Add global publication list

gerrod3:
fixes: #515

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -792,7 +792,11 @@ def list_command(**kwargs: Any) -> click.Command:
         """
         Show the list of optionally filtered {entities}.
         """
-        parameters = {k: v for k, v in kwargs.items() if v is not None}
+        parameters = {
+            key: entity_ctx._preprocess_value(key, value)
+            for key, value in kwargs.items()
+            if value is not None
+        }
         result = entity_ctx.list(limit=limit, offset=offset, parameters=parameters)
         pulp_ctx.output_result(result)
 
--- a/pulpcore/cli/rpm/publication.py
+++ b/pulpcore/cli/rpm/publication.py
@@ -1,6 +1,6 @@
 import click
 
-from pulpcore.cli.common.context import PulpContext, pass_pulp_context
+from pulpcore.cli.common.context import PluginRequirement, PulpContext, pass_pulp_context
 from pulpcore.cli.common.generic import (
     create_command,
     destroy_command,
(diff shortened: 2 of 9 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -742,7 +742,7 @@ def callback(
         """
         Show the list of optionally filtered {entities}.
         """
-        parameters = {k: v for k, v in kwargs.items() if v is not None}
+        parameters = entity_ctx.preprocess_body(kwargs)
```
mdellweg, line 745: Are you sure, this is what we want here? I expect the parameters to be filters mostly (exclusively?) in case of list commands.

pulpcore/cli/rpm/publication.py
```diff
@@ -37,18 +45,21 @@ def publication(ctx: click.Context, pulp_ctx: PulpContext, publication_type: str
...
+        "--version", type=int, help=_("a repository version number, leave blank for latest")
+    ),
+]
+filter_options = publication_filter_options + [
     resource_option(
         "--repository",
         default_plugin="rpm",
         default_type="rpm",
         context_table={"rpm:rpm": PulpRpmRepositoryContext},
-    ),
```
mdellweg, line 45: Should this be the same `repository_option` now?

tests/scripts/pulp_file/test_publication.sh
```diff
@@ -20,4 +20,9 @@ PUBLICATION_HREF="$(echo "$OUTPUT" | jq -r .pulp_href)"
 expect_succ pulp file publication destroy --href "$PUBLICATION_HREF"
 expect_succ pulp file publication create --repository "cli_test_file_repository" --version 0
 PUBLICATION_HREF="$(echo "$OUTPUT" | jq -r .pulp_href)"
+if pulp debug has-plugin --name "core" --min "3.20.0"
+then
+  expect_succ pulp file publication list --repository "cli_test_file_repository"
+  test "$(echo "$OUTPUT" | jq -r length)" -eq 1
```
mdellweg, line 26: If i understand your change correctly...

```suggestion
  expect_succ pulp file publication list --repository "cli_test_file_repository"
  test "$(echo "$OUTPUT" | jq -r length)" -eq 1
  expect_succ pulp publication list --repository "file:file:cli_test_file_repository"
  test "$(echo "$OUTPUT" | jq -r length)" -eq 1
```

pulpcore/cli/common/generic.py
gerrod3: I just wanted to use a `resource_option` as a filter option, and `preprocess_body` calls `_preprocess_value` which will convert a context to a href in the body. Maybe I should switch the call since `preprocess_body` can change arguments (usually for create calls) which might not be ideal.

pulpcore/cli/rpm/publication.py
gerrod3: Not sure where this comment is pointing to.

gerrod3: The git diff is a bit wacky, but the difference between the two is that the filter option requires pulpcore 3.20, while the create option does not.

mdellweg: Ah, thank you.

mdellweg (dismissed):

mdellweg (approved):

PR #527, opened 2022-07-08
Add global distribution commands

gerrod3:
fixes: #517


pulpcore/cli/core/context.py
```diff
@@ -227,6 +227,13 @@ class PulpContentRedirectContentGuardContext(PulpEntityContext):
     ID_PREFIX = "contentguards_core_content_redirect"
 
 
+class PulpDistributionContext(PulpEntityContext):
```
mdellweg, line 230: Should we have all the Distributions inherit from this? (I know there is little value in this unless we have some fancy extra methods or NULLABLES here.)

mdellweg (approved):

pulpcore/cli/core/context.py
gerrod3: I would say we leave it as is now and in the future we find extra bits that need to be shared we move it to common and have others inherit.

PR #533, opened 2022-07-17
Fix config file loading due to lack of permissions

### LINKED ISSUE #509: pulp-cli fails to start if a config file location is inaccessible (permissions)

mdellweg:
Config files that cannot be accessed due to the lack of permissions are
now simply ignored.

fixes #509

Co-Authored-By: Matthias Dellweg <dellweg@redhat.com>

### DIFF
--- a/CHANGES/509.bugfix
+++ b/CHANGES/509.bugfix
@@ -1,0 +1,2 @@
+Fixed bug, where the failure to load config file due to the lack of file permissions lead to a crash.
+Now those files are simply ignored.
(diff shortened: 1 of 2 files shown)


ggainey (approved):
Looks good, modulo the 3.9 CI failure

gerrod3 (approved):
CI error was fixed by https://github.com/pulp/pulp-cli/pull/532, merging.

PR #540, opened 2022-08-03
Added --optimize and --sync-policy options to rpm repo sync.

### LINKED ISSUE #462: RPM sync ought to provide `sync_policy` option, mark `mirror` as deprecated
The RPM plugin provides a `sync_policy` parameter to replace the `mirror` parameter and has marked the `mirror` parameter as deprecated.  The reason is that `mirror = bool` doesn't capture the full range of potentially desired behaviors, and is ambiguous when there could be multiple different meanings of "mirror".

The current choices for `sync_policy` are "additive", "mirror_content_only", and "mirror_complete"

https://github.com/pulp/pulp_rpm/pull/2146
https://docs.pulpproject.org/pulp_rpm/workflows/create_sync_publish.html#sync-repository-foo-using-remote-bar

ggainey:
fixes #462.

Co-authored by: Grant Gainey <ggainey@redhat.com>

### DIFF
--- a/pulpcore/cli/rpm/repository.py
+++ b/pulpcore/cli/rpm/repository.py
@@ -48,6 +48,7 @@ _ = translation.gettext
 
 
 SKIP_TYPES = ["srpm"]
+SYNC_TYPES = ["additive", "mirror_complete", "mirror_content_only"]
 remote_option = resource_option(
     "--remote",
     default_plugin="rpm",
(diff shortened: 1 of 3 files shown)


pulpcore/cli/rpm/repository.py
```diff
@@ -165,16 +166,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
 @name_option
 @href_option
 @remote_option
-@click.option("--mirror/--no-mirror", default=None)
 @click.option(
-    "--skip-type", "skip_types", multiple=True, type=click.Choice(SKIP_TYPES, case_sensitive=False)
+    "--mirror/--no-mirror",
+    default=None,
+    help="""
+    DEPRECATED: If True, sync_policy will default to 'mirror_complete' instead
```
mdellweg, line 173: I don't know. Is "deprecated" the proper term for the cli? Should we explain, that this is only for certain old versions of the plugin? How about `needs_plugin` with an upper bound?

pulpcore/cli/rpm/repository.py
```diff
@@ -165,16 +166,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
...
+    of 'additive'.
+    """,
+)
+@pulp_option(
+    "--optimize/--no-optimize",
+    default=None,
+    help="Whether or not to optimize sync.",
+    needs_plugins=[PluginRequirement("rpm", "3.3.0")],
```
mdellweg, line 181: Tiny little bit more explicit:
```suggestion
    needs_plugins=[PluginRequirement("rpm", min="3.3.0")],
```

pulpcore/cli/rpm/repository.py
```diff
@@ -165,16 +166,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
...
+    """,
+    needs_plugins=[PluginRequirement("rpm", "3.16.0")],
 )
 @pass_repository_context
 def sync(
     repository_ctx: PulpRepositoryContext,
     remote: EntityFieldDefinition,
     mirror: Optional[bool],
+    optimize: Optional[bool],
     skip_types: Iterable[str],
```
mdellweg, line 210:
```suggestion
    skip_types: Optional[Iterable[str]],
```

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -27,12 +27,13 @@ expect_succ pulp rpm repository show --name "cli_test_rpm_repository"
...
-else
-  expect_succ pulp rpm repository sync --name "cli_test_rpm_repository" --skip-type srpm
-fi
+#if pulp debug has-plugin --name "rpm" --min-version "3.12.0" --max-version "3.12.1"
+#then
+#  expect_succ pulp rpm repository sync --name "cli_test_rpm_repository"
+#else
+#  expect_succ pulp rpm repository sync --name "cli_test_rpm_repository" --skip-type srpm
+#fi
+expect_succ pulp rpm repository sync --name "cli_test_rpm_repository"
```
mdellweg, line 36: Why this change? (The conditional was probably for a bug with srpms in 3.12.0 fixed in 3.12.1 .)
If you think we do not need to work around anymore, please remove the commented part.

pulpcore/cli/rpm/repository.py
ggainey: It's deprecated with the same wording in the pulp_rpm REST API : https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/app/serializers/repository.py#L298

It's been deprecated since pulp_rpm/3.16.0. I don't want to put a max=3.16 for RPM on it, because people are using it now, and would suddenly break. We could put a max=3.17 on it, which would make it stop working when pulp_rpm/3.18 is released Really Soon Now (and I don't know if pulp_rpm is going to be able to actually remove the option any time soon)

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
mdellweg: `Error: ["Cannot use 'retain_package_versions' in combination with a 'mirror_complete' or 'mirror_content_only' sync policy."]`

pulpcore/cli/rpm/repository.py
mdellweg: OK, as long as the option is there, we should support it. :+1:

mdellweg (approved):

PR #544, opened 2022-08-22
Add preprocess_entity to replace preprocess_body

mdellweg:


pulpcore/cli/common/generic.py
```diff
@@ -852,7 +849,7 @@ def callback(pulp_ctx: PulpContext, entity_ctx: PulpEntityContext, **kwargs: Any
         """
         Create a {entity}.
         """
-        body: EntityDefinition = entity_ctx.preprocess_body(kwargs)
+        body: EntityDefinition = entity_ctx.preprocess_entity(kwargs, partial=False)
```
mdellweg, line 852: I would like to move the call for `preprocess_entity` into the `create` and `update` calls, but it's not idempotent, and we may run into trouble with external plugins. Suggestions?

gerrod3: - Cache the results of `preprocess_entity`
- Add a field to result of `preprocess_entity` that `create` and `update` would remove and know not to run it a second time
- Change the function signature of `create` and `update` to take in an unprocessed body
- In `create` & `update` run `preprocess_entity` on a copy of `body` and then **magically** chose the correct one to use.

mdellweg: I went with a slight modification of your second suggestion: A subclass of dict.

pulpcore/cli/common/context.py
```diff
@@ -60,6 +64,21 @@ def default(self, obj: Any) -> Any:
             return super().default(obj)
 
 
+def _preprocess_value(key: str, value: Any) -> Any:
+    if isinstance(value, PulpEntityContext):
+        return value.pulp_href
+    return value
+
+
+def preprocess_body(body: EntityDefinition) -> EntityDefinition:
```
mdellweg, line 73: We now preprocess body _and_ parameters with this. So maybe it is not the best name for the function.

I'm going to call it `preprocess_payload`.

gerrod3 (approved):
Nice, I like the option you went with.

PR #546, opened 2022-09-01
Move validation steps into context

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/cli/core/context.py
+++ b/pulpcore/cli/core/context.py
@@ -23,11 +23,12 @@ class PulpAccessPolicyContext(PulpEntityContext):
     ID_PREFIX = "access_policies"
 
     def reset(self) -> Any:
+        self.pulp_ctx.needs_plugin(PluginRequirement("core", min="3.17.0"))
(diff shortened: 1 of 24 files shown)


mdellweg:
I see a lot of duplicated code here again... XD Thinking about adding the needs_plugins logic on the context class level.

pulpcore/cli/core/context.py
```diff
@@ -23,6 +23,7 @@ class PulpAccessPolicyContext(PulpEntityContext):
     ID_PREFIX = "access_policies"
 
     def reset(self) -> Any:
+        self.pulp_ctx.needs_plugin(PluginRequirement("core", min="3.17"))
```
gerrod3, line 26: We are not being consistent in writing the "z" stream parameter when it is zero.

mdellweg: I think of fetures to come with a X.Y while bugfixes and workarounds need the specific X.Y.Z but i can take a look and make them all X.Y.Z. WDYT?

gerrod3 (approved):

PR #552, opened 2022-09-12
Disable sending telemetry in CI

mdellweg:
cc @bmbouter


bmbouter (approved):
I'm not sure how we're going to CI this feature overall. I guess the best idea would be to write unit tests? I guess we should approve this. Let's get another LGTM from someone in telemetry then, @ggainey @ipanova @dkliban wdyt?

bmbouter:
I disabled TELEMETRY for all repos via [this plugin template change](https://github.com/pulp/plugin_template/pull/690). Do you think this PR is still needed?

mdellweg:
Absolutely

bmbouter:
> Absolutely

It seems they both are setting the same setting though. Is the setting not applying in the same way here or something?

PR #554, opened 2022-09-19
Add filter parameters to task list

### LINKED ISSUE #543: Support remaining 'task' query-parameters
## Summary
Currently you can list tasks by name, state, and task-group. We're missing a lot of search-options the API supports

## Examples
- created_resources [href...]
- finished_at (eq, gt, gte, lt, lte, range()) datetime
- parent href
- reserved_resources_record [href...]
- started_at (eq, gt, gte, lt, lte, range()) datetime
- worker href
- worker__in [href...]

mdellweg:
fixes #543


pulpcore/cli/core/generic.py
```diff
@@ -71,6 +94,52 @@ def _task_group_filter_callback(
         help=_("List only tasks in this task group. Provide pulp_href or UUID."),
         callback=_task_group_filter_callback,
     ),
+    click.option(
+        "--parent-task",
+        help=_("Parent task by uuid or href."),
+        callback=_task_filter_callback,
+    ),
+    resource_option(
+        "--worker",
```
ggainey, line 103: We support worker__in as well - this should support multiple=True, yeah?

pulpcore/cli/core/generic.py
```diff
@@ -71,6 +94,52 @@ def _task_group_filter_callback(
...
+        href_pattern=PulpWorkerContext.HREF_PATTERN,
+        help=_("Worker used to execute the task by name or href."),
+    ),
+    click.option(
+        "--created-resource",
+        "created_resources",
+        help=_("Href of a resource created in the task."),
+    ),
+    click.option(
+        "--reserved-resources-record",
```
ggainey, line 116: reserved_resources expects an array of resource-ids - multiple=True

tests/scripts/pulpcore/test_task.sh
```diff
@@ -41,9 +44,16 @@ task=$(echo "$ERROUTPUT" | grep -E -o "${PULP_API_ROOT}api/v3/tasks/[-[:xdigit:]
...
+reserved_resource="$(echo "$OUTPUT" | jq -r '.reserved_resources_record[0]')"
+created_resource="$(echo "$OUTPUT" | jq -r '.created_resources[0]')"
 expect_succ test "$(echo "$OUTPUT" | jq -r '.state')" = "completed"
 
 expect_succ pulp task list --name-contains file
+expect_succ pulp task list --parent-task "$task" --worker "$worker"
+expect_succ pulp task list --parent-task "$task_uuid" --worker "$worker_name"
+expect_succ pulp task list --started-before "2021-12-01" --started-after "2022-06-01"
+expect_succ pulp task list --finished-before "2021-12-01" --finished-after "2022-06-01"
+expect_fail pulp task list --reserved-resource "$reserved_resource" --created-resource "$created_resource"
```
ggainey, line 56: Why would we expect this to fail? Is it just "there's no way to specify whether this is an AND or OR combination"?

tests/scripts/pulpcore/test_task.sh
```diff
@@ -41,9 +44,16 @@ task=$(echo "$ERROUTPUT" | grep -E -o "${PULP_API_ROOT}api/v3/tasks/[-[:xdigit:]
...
 expect_succ pulp task show --wait --uuid "$task_uuid"
+reserved_resource="$(echo "$OUTPUT" | jq -r '.reserved_resources_record[0]')"
+created_resource="$(echo "$OUTPUT" | jq -r '.created_resources[0]')"
 expect_succ test "$(echo "$OUTPUT" | jq -r '.state')" = "completed"
 
 expect_succ pulp task list --name-contains file
+expect_succ pulp task list --parent-task "$task" --worker "$worker"
+expect_succ pulp task list --parent-task "$task_uuid" --worker "$worker_name"
+expect_succ pulp task list --started-before "2021-12-01" --started-after "2022-06-01"
+expect_succ pulp task list --finished-before "2021-12-01" --finished-after "2022-06-01"
```
ggainey, line 55: Can we add a test that shows the full H/M/S date format working, as well as a negative one failing when you use, say, a US-formatted date?

ggainey (changes requested):

pulpcore/cli/core/generic.py
```diff
@@ -21,6 +21,29 @@
 ##############################################################################
 # Generic reusable commands
 
+DATETIME_FORMATS = ["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S"]
+
+
+def _task_filter_callback(
```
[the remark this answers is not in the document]
mdellweg: we have `resource_option` but it didn't fit the bill.

pulpcore/cli/core/generic.py
mdellweg: I can try. Don't see much value though.

tests/scripts/pulpcore/test_task.sh
mdellweg: Oh, it should be pass...

tests/scripts/pulpcore/test_task.sh
mdellweg: We can easily support what the server accepts.

mdellweg: That's actually not true. Click interprets the date for us.

pulpcore/cli/core/generic.py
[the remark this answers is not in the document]
ggainey: I was thinking more along the lines of a utility function that would encapsulate L30-45, that could be called with a path-suffix and then do the uuid-or-href computation, so it could be used by other, similar, options. As an example, task_group does something nearly identical here : https://github.com/pulp/pulp-cli/blob/main/pulpcore/cli/core/generic.py#L25

It would be nifty if, say , that code could devolve into a call to. say, `uuid_or_href(value, "task_groups")`, and this could become a call to `uuid_or_href(value, "tasks")`. Am I being clear here?

tests/scripts/pulpcore/test_task.sh
ggainey: One benefit of the tests is an examples of "how to use this", so if a param can have two formats, I'm a fan of showing both in the tests.

pulpcore/cli/common/context.py
```diff
@@ -27,6 +27,20 @@
 
 DEFAULT_LIMIT = 25
 BATCH_SIZE = 25
+DATETIME_FORMATS = [
```
ggainey, line 30: :+1:

pulpcore/cli/common/context.py
```diff
@@ -67,6 +81,8 @@ def default(self, obj: Any) -> Any:
 def _preprocess_value(key: str, value: Any) -> Any:
     if isinstance(value, PulpEntityContext):
         return value.pulp_href
+    if isinstance(value, datetime.datetime):
```
ggainey, line 84: :+1:

pulpcore/cli/core/context.py
```diff
@@ -341,7 +342,11 @@ def scope(self) -> Dict[str, Any]:
         else:
             return {}
 
-    def purge(self, finished_before: Optional[str], states: Optional[List[str]]) -> Any:
+    def purge(
+        self,
+        finished_before: Optional[datetime.datetime],
```
ggainey, line 347: :heart:

pulpcore/cli/core/generic.py
mdellweg: The `worker__in` filter complains about "not a valid UUID" when it was supposed to accept hrefs.
I'm not going to implement that here.

pulpcore/cli/core/generic.py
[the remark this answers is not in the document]
mdellweg: Have a look again. The typical question here: Class or closure? (I made it a class this time for no particular preference...)

pulpcore/cli/core/generic.py
mdellweg: It looks complicated. Do we absolutely need it today? I'd be happy waiting for a real user request.

BTW: We have a `pulp file repository task list` command that can (almost?) do the same.

pulpcore/cli/common/openapi.py
```diff
@@ -142,11 +142,23 @@ def extract_params(
...
+                    param_type = param_schema.get("type", "string")
+                    if param_type == "array":
+                        if not param:
+                            continue
+                        assert isinstance(param, Iterable) and not isinstance(param, str)
+                        if not param_spec.get("explode", False):
+                            param = ",".join(param)
+                    elif param_type == "integer":
+                        assert isinstance(param, int)
+                result[name] = param
```
mdellweg, line 159: Proper handling of exploded or not exploded arrays kind of automatically leads to more validation of the parameters. ;)

mdellweg: And extra validation may lead to more bugs uncovered... Need to find time investigate these...

mdellweg: I put the validation and array substitution on a separate branch to investigate separately:

https://github.com/mdellweg/pulp-cli/tree/validate_parameters

pulpcore/cli/core/generic.py
```diff
@@ -22,27 +23,31 @@
...
+            href_match = re.match(
+                rf"({pulp_ctx.api_path}{self.base_href}/)?"
+                r"([0-9a-f]{8})-?([0-9a-f]{4})-?([0-9a-f]{4})-?([0-9a-f]{4})-?([0-9a-f]{12})/?",
+                value,
             )
-        else:
-            raise click.ClickException(_("Either an href or a UUID must be provided."))
+            if href_match:
+                value = "{}{}/{}-{}-{}-{}-{}/".format(
+                    pulp_ctx.api_path, self.base_href, *href_match.group(2, 3, 4, 5, 6)
```
lubosmj, line 45: Is matching groups like this even necessary? Would it make sense to have one group `[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}` which will be projected to the formatted string?

mdellweg: I think the regex accepts the uuid with and without the dashes. That's why we need to do it that way, no?

lubosmj: Ah, I overlooked those question marks. Yes. You are right.

mdellweg:
I'm currently experimenting with https://github.com/pulp/pulpcore/pull/3282 and therefore marking this as draft again.

If we go down that road, we need some version dependent fixups here.

PR #556, opened 2022-09-26
Enable sending query parameters for selecting and ordering

### LINKED ISSUE #542: Support base-object field limits and ordering
## Summary
I would like pulp-cli to expose the `ordering`, `fields`, and `exclude_fields` attributes for all entities.

## Examples
`pulp rpm remote list --field name --field url --ordering name`

lubosmj:
closes #542

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -533,6 +533,26 @@ offset_option = pulp_option(
     help=_("Skip a number of {entities} to show."),
 )
 
+ordering_option = pulp_option(
+    "--ordering",
+    required=False,
+    help=_("A field that will be used to order the results."),
+)
+field_option = pulp_option(
+    "--field",
+    "fields",
+    multiple=True,
+    required=False,
+    help=_("A field that is to be selected from a result. Can be specified multiple times."),
+)
+exclude_field_option = pulp_option(
+    "--exclude-field",
+    "exclude_fields",
+    multiple=True,
+    required=False,
+    help=_("A field that is to be excluded from a result. Can be specified multiple times."),
+)
+
 href_option = pulp_option(
     "--href",
     help=_("HREF of the {entity}"),
@@ -790,6 +810,9 @@ def list_command(**kwargs: Any) -> click.Command:
     @pulp_command(**kwargs)
     @limit_option
     @offset_option
+    @ordering_option
+    @field_option
+    @exclude_field_option
     @pass_entity_context
     @pass_pulp_context
     def callback(
--- a/CHANGES/542.feature
+++ b/CHANGES/542.feature
@@ -1,0 +1,1 @@
+Added support for query parameters: ordering, fields, exclude_fields.
--- a/tests/scripts/pulp_file/test_query_params.sh
+++ b/tests/scripts/pulp_file/test_query_params.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
+
+cleanup() {
+  pulp file repository destroy --name aaaaaaaaaa || true
+  pulp file repository destroy --name bbbbbbbbbb || true
+  pulp file repository destroy --name cccccccccc || true
+}
+trap cleanup EXIT
+
+expect_succ pulp file repository create --name aaaaaaaaaa
+expect_succ pulp file repository create --name bbbbbbbbbb
+expect_succ pulp file repository create --name cccccccccc
+
+expect_succ pulp file repository list --ordering -name
+ORDERED_RESULTS=$(echo "$OUTPUT" | jq -r .[].name | tr "\n" " " | xargs)
+EXPECTED_RESULTS="cccccccccc bbbbbbbbbb aaaaaaaaaa"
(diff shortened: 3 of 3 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -533,6 +533,26 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
...
+    "--field",
+    "fields",
+    multiple=True,
+    required=False,
+    help=_("A list of fields that are to be selected from a result."),
+)
+exclude_field_option = pulp_option(
+    "--exclude-field",
+    "exclude_fields",
+    multiple=True,
```
mdellweg, line 551: This will return a list (or tuple). It may require the transformation in #555 ...

If it does not work as expected, use `pulp -v ...` to see what actually gets sent.

lubosmj: It works; but, instead of sending all excluded fields in one query parameter, we send them repeatedly when a user specifies multiple values. I still consider this approach valid.

```
(pulp) [vagrant@pulp3-source-fedora35 scripts]$ pulp -v file repository list --ordering name
get https://pulp3-source-fedora35.localhost.example.com/pulp/api/v3/repositories/file/file/?ordering=name&offset=0&limit=25
Response: 200
[
  {
    "pulp_href": "/pulp/api/v3/repositories/file/file/89925a64-fc14-4fa6-bf69-91861eee9900/",
    "pulp_created": "2022-09-27T09:21:05.268364Z",
    "versions_href": "/pulp/api/v3/repositories/file/file/89925a64-fc14-4fa6-bf69-91861eee9900/versions/",
    "pulp_labels": {},
    "latest_version_href": "/pulp/api/v3/repositories/file/file/89925a64-fc14-4fa6-bf69-91861eee9900/versions/0/",
    "name": "aaa",
    "description": null,
    "retain_repo_versions": null,
    "remote": null,
    "autopublish": false,
    "manifest": "PULP_MANIFEST"
  }
]

(pulp) [vagrant@pulp3-source-fedora35 scripts]$ pulp -v file repository list --exclude-field name --exclude-field remote
get https://pulp3-source-fedora35.localhost.example.com/pulp/api/v3/repositories/file/file/?exclude_fields=name&exclude_fields=remote&offset=0&limit=25
Response: 200
[
  {
    "pulp_href": "/pulp/api/v3/repositories/file/file/89925a64-fc14-4fa6-bf69-91861eee9900/",
    "pulp_created": "2022-09-27T09:21:05.268364Z",
    "versions_href": "/pulp/api/v3/repositories/file/file/89925a64-fc14-4fa6-bf69-91861eee9900/versions/",
    "pulp_labels": {},
    "latest_version_href": "/pulp/api/v3/repositories/file/file/89925a64-fc14-4fa6-bf69-91861eee9900/versions/0/",
    "description": null,
    "retain_repo_versions": null,
    "autopublish": false,
    "manifest": "PULP_MANIFEST"
  }
]
```

mdellweg: It should depend on whether the api specs declare them as `explode=True`.

https://swagger.io/specification/#parameterObject

lubosmj: I am attaching an excerpt from the OpenAPI schema for `/pulp/api/v3/repositories/file/file/` (GET):
```
                    {
                        "in": "query",
                        "name": "fields",
                        "schema": {
                            "type": "string"
                        },
                        "description": "A list of fields to include in the response."
                    },
                    {
                        "in": "query",
                        "name": "exclude_fields",
                        "schema": {
                            "type": "string"
                        },
                        "description": "A list of fields to exclude from the response."
                    }
```

Based on the defaults specified at swagger.io, the endpoint expects separate parameters for each value:
> 1. **Style**: Describes how the parameter value will be serialized depending on the type of the parameter value. Default values (based on value of in): for **query - form**
> 2. **Explode**: When this is true, parameter values of type array or object generate separate parameters for each value of the array or key-value pair of the map. For other types of parameters this property has no effect. **When style is form, the default value is true.**

So, we should be fine.

mdellweg: Great analysis.

mdellweg (approved):

pulpcore/cli/common/generic.py
```diff
@@ -533,6 +533,26 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
...
+    required=False,
+    help=_("A list of fields that are to be excluded from a result."),
```
ggainey, line 553:
```suggestion
    help=_("A field that is to be excluded from a result. Can be specified multiple times."),
```

pulpcore/cli/common/generic.py
```diff
@@ -533,6 +533,26 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
...
+    "--ordering",
+    required=False,
+    help=_("A field that will be used to order the results."),
+)
+field_option = pulp_option(
+    "--field",
+    "fields",
+    multiple=True,
+    required=False,
+    help=_("A list of fields that are to be selected from a result."),
```
ggainey, line 546:
```suggestion
    help=_("A field that is to be selected from a result.  Can be specified multiple times."),
```

ggainey (changes requested):
A little bit of wordsmithing - --field and --exclude-field don't take a list, but can be repeated. But this is great, makes me very happy :)

mdellweg (approved):

PR #557, opened 2022-09-29
Adjust signing service setup to unpriviledged user

mdellweg:
[noissue]


mdellweg:
This should solve two of the three reasons the test fails. Should we merge it like that?

ggainey (approved):

PR #560, opened 2022-10-07
Add "--metadata-signing-service" option to rpm

tjmullicani:
Add "--metadata-signing-service" option to rpm repository

### DIFF
--- a/pulpcore/cli/rpm/repository.py
+++ b/pulpcore/cli/rpm/repository.py
@@ -9,6 +9,7 @@ from pulp_glue.common.context import (
     PulpRepositoryContext,
 )
 from pulp_glue.common.i18n import get_translation
+from pulp_glue.core.context import PulpSigningServiceContext
 from pulp_glue.rpm.context import (
     PulpRpmPackageContext,
     PulpRpmRemoteContext,
@@ -57,6 +58,14 @@ remote_option = resource_option(
     help=_(
         "Remote used for synching in the form '[[<plugin>:]<resource_type>:]<name>' or by href."
     ),
+)
+
+metadata_signing_service_option = resource_option(
+    "--metadata-signing-service",
+    default_plugin="core",
+    default_type="core",
+    context_table={"core:core": PulpSigningServiceContext},
+    href_pattern=PulpSigningServiceContext.HREF_PATTERN,
 )
 
 
@@ -125,6 +134,7 @@ update_options = [
     click.option("--description"),
     click.option("--retain-package-versions", type=int),
     remote_option,
+    metadata_signing_service_option,
     click.option(
         "--metadata-checksum-type", type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False)
     ),
--- a/CHANGES/605.feature
+++ b/CHANGES/605.feature
@@ -1,0 +1,1 @@
+Add '--metadata-signing-service' option to rpm.


tjmullicani:
Reference: https://github.com/pulp/pulp_rpm/blob/54ec7c3f2e0a5a2263054a685ed68213a85b70f4/pulp_rpm/app/serializers/repository.py#L52

pulpcore/cli/rpm/repository.py
```diff
@@ -126,6 +126,7 @@ def repository(ctx: click.Context, pulp_ctx: PulpContext, repo_type: str) -> Non
     click.option("--description"),
     click.option("--retain-package-versions", type=int),
     remote_option,
+    click.option("--metadata-signing-service", type=str),
```
gerrod3, line 129: Can you possibly make this a `resource_option` like the `remote_option`? This would allow users to supply either the name or the href of the signing service. You could use this context: https://github.com/pulp/pulp-cli/blob/main/pulpcore/cli/core/context.py#L314 for the `context_table`.

gerrod3 (commented):
Thanks for the contribution! In order to pass our Ready-To-Ship CI check you need to attach an issue to the commit, e.g. `fixes: #XXX`

pulpcore/cli/rpm/repository.py
mdellweg: This option probably needs a version guard (`needs_plugins=PluginRequirement("rpm", min=???)`).

cc @pulp/rpm Can you verify this?

ggainey: > This option probably needs a version guard (`needs_plugins=PluginRequirement("rpm", min=???)`).
> 
> cc @pulp/rpm Can you verify this?

metadata_signing_service has been in pulp_rpm since forever (def pre-3.9) - so we're good.

mdellweg:
ping @tjmullicani any progress on this one?

tjmullicani:
Updated code

pulpcore/cli/rpm/repository.py
```diff
@@ -61,6 +61,14 @@
     ),
 )
 
+metadata_signing_service_option = resource_option(
+    "--metadata-signing-service",
+    default_plugin="rpm",
+    default_type="rpm",
+    context_table={"rpm:rpm": PulpSigningServiceContext},
```
mdellweg, line 68: These should be "core:core".

mdellweg:
How do you think about adding a test for this?

pulpcore/cli/rpm/repository.py
```diff
@@ -43,6 +43,7 @@
     PulpRpmRepositoryContext,
     PulpUlnRemoteContext,
 )
+from pulpcore.cli.core.context import PulpSigningServiceContext
```
mdellweg, line 46: can you import from pulp_glue.core.context instead?

mdellweg (commented):
there is a signing service referenced in `pulp ansible repository sign` that can make good use of the `resource_option`, too.

pulpcore/cli/rpm/repository.py
tjmullicani: fixed

tjmullicani:
requires #652

CHANGES/607.feature
```diff
@@ -0,0 +1 @@
+Add "--metadata-signing-service" option to rpm
```
mdellweg, line 1:
```suggestion
Add '--metadata-signing-service' option to rpm.
```

mdellweg (changes requested):
Please squash the commits.

mdellweg (approved):

PR #571, opened 2022-10-14
Fixes another bug in the query params test.

dkliban:
[noissue]

### DIFF
--- a/tests/scripts/pulpcore/test_query_params.sh
+++ b/tests/scripts/pulpcore/test_query_params.sh
@@ -29,8 +29,8 @@ if ! (echo "$OUTPUT" | jq -r .[].name | sort -C -); then
 fi
 
 expect_succ pulp file repository list --field name --field remote
-SELECTED_FIELDS=$(echo "$OUTPUT" | jq -r ".[] | keys[]" | sort| tr "\n" " " | xargs)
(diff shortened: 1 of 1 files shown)


tests/scripts/pulpcore/test_query_params.sh
```diff
@@ -29,8 +29,8 @@ if ! (echo "$OUTPUT" | jq -r .[].name | sort -C -); then
 fi
 
 expect_succ pulp file repository list --field name --field remote
-SELECTED_FIELDS=$(echo "$OUTPUT" | jq -r ".[] | keys[]" | sort| tr "\n" " " | xargs)
-EXPECTED_FIELDS="name name name remote remote remote"
+SELECTED_FIELDS=$(echo "$OUTPUT" | jq -r ".[] | keys[]" | sort | sort | uniq | tr "\n" " " | xargs)
```
mdellweg, line 32:
```suggestion
SELECTED_FIELDS=$(echo "$OUTPUT" | jq -r ".[] | keys[]" | sort -u | tr "\n" " " | xargs)
```

mdellweg (changes requested):
Sort is already idempotent. No need to do it twice. ;)

mdellweg (approved):

PR #573, opened 2022-10-15
Fix missing fields KeyError

daviddavis:
Review Checklist:


daviddavis:
Here's the command I used to reproduce this (but note that this works for any plugin type):

```
pulp deb distribution create --name complex-dists
```

Obviously the problem is that base_path is missing. I could write a test for this but then the test will be void once #574 is fixed. I also tried looking for another field in the openapi schema that is required in the schema but not as an option that I could test with but gave up after a bit.

mdellweg (approved):
Nice find!

PR #577, opened 2022-10-19
Refactor the labels code

### LINKED ISSUE #274: As a user, I can specify labels when creating/updating resources
Currently, we have a label subcommand to manage labels:

```
pulp file remote label set ...
pulp file remote label unset ...
pulp file remote label show ...
```

It would be nice to also have an option when creating/updating resources.

mdellweg:
fixes #274

Review Checklist:

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -332,6 +332,21 @@ def load_json_callback(ctx: click.Context, param: click.Parameter, value: Option
         return json_object
 
 
+def labels_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[str]
+) -> Optional[Dict[str, str]]:
+    # None is legal - shortcircuit here
+    if value is None:
+        return value
+
+    value = load_json_callback(ctx, param, value)
+    if isinstance(value, dict) and all(
+        (isinstance(key, str) and isinstance(val, str) for key, val in value.items())
+    ):
+        return value
+    raise click.ClickException(_("Labels must be provided as a dictionary of strings."))
+
+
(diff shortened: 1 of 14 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -712,6 +712,13 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
     help=_("Number of repository versions to keep."),
 )
 
+pulp_labels_option = pulp_option(
+    "--labels",
+    "pulp_labels",
+    help=_("JSON dictionary of labels to set on {entity}"),
+    callback=load_json_callback,
+)
```
mdellweg, line 720: This is the easiest version to allow for this, but maybe we can do better.
Should we for example make a `multiple=True` option that takes `<key> <value>` args and assemble the dict ourselves?
The nice thing about this one is that we can pass the content of a json-file. But we should probably add a little schema validation around it.

ggainey: the multiple=True path is better for the user that's doing a small number of labels as one-offs. If I'm setting up a process where I want more than a couple of labels, or I want a standard set of labels applied to a standard set of things every time, accepting a JSON file is way easier.

I suppose we could have --labels=(JSON or @JSON), **and** a --label=(key value) multiple=True? That covers both usecases. Our code would be responsible for Doing The Right Thing, of course...

ggainey (approved):
I think there are more use-cases that are well-served by JSON/JSON-file, than by multiple=True. I'm good with where this is now (and if there's desire to support both, we can work out the resulting set-theoretic decisions then)

PR #590, opened 2022-11-07
Add domains support

gerrod3:
Still working on this, current behavior is reliant on docs endpoint being still exposed at original endpoint `/pulp/api/v3/docs/api.json`. 

Review Checklist:

### DIFF
--- a/pulp-glue/pulp_glue/common/context.py
+++ b/pulp-glue/pulp_glue/common/context.py
@@ -3,7 +3,20 @@ import os
 import re
 import sys
 import time
-from typing import IO, Any, ClassVar, Dict, List, Mapping, NamedTuple, Optional, Set, Type, Union
+from typing import (
+    IO,
+    Any,
+    ClassVar,
+    Dict,
+    List,
+    Mapping,
+    NamedTuple,
+    Optional,
+    Set,
+    Type,
+    Union,
+    cast,
+)
 
 from packaging.version import parse as parse_version
 from requests import HTTPError
@@ -95,12 +108,14 @@ class PulpContext:
         format: str,
         background_tasks: bool,
         timeout: int,
+        domain: str = "default",
     ) -> None:
         self._api: Optional[OpenAPI] = None
-        self.api_path: str = api_root + "api/v3/"
+        self._api_root: str = api_root
         self._api_kwargs = api_kwargs
         self._needed_plugins: List[PluginRequirement] = [PluginRequirement("core", min="3.11.0")]
         self.isatty: bool = sys.stdout.isatty()
+        self.pulp_domain: str = domain
 
         self.format: str = format
         self.background_tasks: bool = background_tasks
@@ -156,12 +171,24 @@ class PulpContext:
                 patched_python_remote_serializer["properties"][prop]["items"] = {"type": "string"}
 
     @property
+    def domain_enabled(self) -> bool:
+        return cast(bool, self.api.api_spec.get("info", {}).get("x-pulp-domain-enabled", False))
+
+    @property
+    def api_path(self) -> str:
+        if self.domain_enabled:
+            return self._api_root + self.pulp_domain + "/api/v3/"
+        return self._api_root + "api/v3/"
+
+    @property
     def api(self) -> OpenAPI:
         if self._api is None:
             if self._api_kwargs.get("username") and not self._api_kwargs.get("password"):
                 self._api_kwargs["password"] = self.prompt("password", hide_input=True)
             try:
-                self._api = OpenAPI(doc_path=f"{self.api_path}docs/api.json", **self._api_kwargs)
+                self._api = OpenAPI(
+                    doc_path=f"{self._api_root}api/v3/docs/api.json", **self._api_kwargs
+                )
             except OpenAPIError as e:
                 raise PulpException(str(e))
             # Rerun scheduled version checks
@@ -189,8 +216,13 @@ class PulpContext:
         Returns the operation result, or the finished task.
         If non_blocking, returns unfinished tasks.
         """
-        if parameters is not None:
-            parameters = preprocess_payload(parameters)
+        if parameters is None:
+            parameters = {}
+        if self.domain_enabled:
+            # Validation will fail if path doesn't need domain parameter
+            if "pulp_domain" in self.api.param_spec(operation_id, "path", required=True):
+                parameters["pulp_domain"] = self.pulp_domain
+        parameters = preprocess_payload(parameters)
         if body is not None:
             body = preprocess_payload(body)
         try:
--- a/pulp-glue/pulp_glue/common/openapi.py
+++ b/pulp-glue/pulp_glue/common/openapi.py
@@ -150,6 +150,29 @@ class OpenAPI:
         else:
             self._session.headers["Correlation-ID"] = correlation_id
 
+    def param_spec(
+        self, operation_id: str, param_type: str, required: bool = False
+    ) -> Dict[str, Any]:
+        method, path = self.operations[operation_id]
+        path_spec = self.api_spec["paths"][path]
+        method_spec = path_spec[method]
+
+        param_spec = {
+            entry["name"]: entry
+            for entry in path_spec.get("parameters", [])
+            if entry["in"] == param_type
+        }
+        param_spec.update(
+            {
+                entry["name"]: entry
+                for entry in method_spec.get("parameters", [])
+                if entry["in"] == param_type
+            }
+        )
+        if required:
+            param_spec = {k: v for k, v in param_spec.items() if v.get("required", False)}
+        return param_spec
+
     def extract_params(
         self,
         param_in: str,
--- a/pulp-glue/pulp_glue/core/context.py
+++ b/pulp-glue/pulp_glue/core/context.py
@@ -86,6 +86,14 @@ class PulpArtifactContext(PulpEntityContext):
         return task["created_resources"][0]
 
 
+class PulpDomainContext(PulpEntityContext):
+    ENTITY = _("Pulp domain")
+    ENTITIES = _("Pulp domains")
+    HREF = "domain_href"
+    ID_PREFIX = "domains"
+    NEEDS_PLUGINS = [PluginRequirement("core", "3.23.0.dev")]
+
+
 class PulpExporterContext(PulpEntityContext):
     ENTITY = _("Pulp exporter")
     ENTITIES = _("Pulp exporters")
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -470,6 +470,9 @@ def pulp_option(*args: Any, **kwargs: Any) -> Callable[[FC], FC]:
     return click.option(*args, **kwargs)
 
 
+domain_pattern = r"(?P<pulp_domain>[-a-zA-Z0-9_]+)"
+
+
 def resource_lookup_option(*args: Any, **kwargs: Any) -> Callable[[FC], FC]:
     lookup_key: str = kwargs.pop("lookup_key", "name")
     context_class: Type[PulpEntityContext] = kwargs.pop("context_class")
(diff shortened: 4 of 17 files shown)


pulpcore/cli/common/openapi.py
```diff
@@ -291,6 +293,11 @@ def call(
             parameters = {}
         else:
             parameters = parameters.copy()
+        if self.api_spec["info"].get("x-pulp-domain-enabled", False):
```
mdellweg, line 296: I know, this is an early draft, but i'd like to keep any thing Pulp-specific out of the `openapi.py` file. Can this be handled in `context.py`?

gerrod3: I put the domains check into `context.py`, but I had to add a new method to `OpenApi`, `param_spec(operation_id, param_type)`, so I could know when to add `domains_path` to the parameters.

tests/scripts/pulp_file/test_acs.sh
```diff
@@ -38,7 +38,7 @@ test "$(echo "$OUTPUT" | jq ".paths | length")" -eq 2
 
 # test refresh
 expect_succ pulp --background file acs refresh --acs $acs
-task_group=$(echo "$ERROUTPUT" | grep -E -o "${PULP_API_ROOT}api/v3/task-groups/[-[:xdigit:]]*/")
+task_group=$(echo "$ERROUTPUT" | grep -E -o "/[/a-zA-Z0-9_]*/api/v3/task-groups/[-[:xdigit:]]*/")
```
ggainey, line 41: This regex will allow some really odd URLs. 

PULP_API_ROOT starts and ends with /, domain-name is "just" the name. Can we put an if-domain switch here, with something like 

```suggestion
task_group=$(echo "$ERROUTPUT" | grep -E -o "${PULP_API_ROOT}{$domain_name}/api/v3/task-groups/[-[:xdigit:]]*/")
```

if-domain, and the existing check if-not-domain?

ggainey (changes requested):

pulp-glue/pulp_glue/core/context.py
```diff
@@ -87,6 +87,13 @@ def upload(
         return task["created_resources"][0]
 
 
+class PulpDomainContext(PulpEntityContext):
+    ENTITY = _("Pulp domain")
+    ENTITIES = _("Pulp domains")
+    HREF = "domain_href"
+    ID_PREFIX = "domains"
```
mdellweg, line 94: This should have a NEED_PLUGINS.

tests/scripts/pulp_file/test_acs.sh
mdellweg: It is also de facto of a subtle check that the href has the expected format. I think the grep would fail our test. I do not feel stong either way. If you simplify, however we can just use `.*`.

pulpcore/cli/common/generic.py
```diff
@@ -490,7 +490,11 @@ def _option_callback(
 
         if value.startswith("/"):
             # The HREF of a resource was passed
-            pattern = rf"^{pulp_ctx.api_path}{entity_ctx.HREF_PATTERN}"
+            href_pattern = entity_ctx.HREF_PATTERN
+            if pulp_ctx.domain_enabled:
+                pattern = rf"^{pulp_ctx._api_root}(?P<pulp_domain>\w+)/api/v3/{href_pattern}"
```
mdellweg, line 495: Why is this needed? We only accept entities of the proper domain anyway. If the domain in hrefs is optional (I don't think it is), we can still do `rf"^{pulp_ctx._api_root}(?{pulp_ctx.domain_name}/)api/v3/{href_pattern}"` right?

pulpcore/cli/common/generic.py
```diff
@@ -547,7 +551,10 @@ def _option_callback(
                         option_name=param.name
                     )
                 )
-            pattern = rf"^{pulp_ctx.api_path}{href_pattern}"
+            if pulp_ctx.domain_enabled:
+                pattern = rf"^{pulp_ctx._api_root}(?P<pulp_domain>\w+)/api/v3/{href_pattern}"
+            else:
+                pattern = rf"^{pulp_ctx.api_path}{href_pattern}"
```
mdellweg, line 557: Same here.

pulpcore/cli/core/group.py
```diff
@@ -193,13 +207,13 @@ def role(ctx: click.Context, pulp_ctx: PulpCLIContext, group_ctx: PulpGroupConte
         decorators=[
             group_option,
             click.option("--role", required=True),
-            click.option("--object", "content_object", required=True),
```
mdellweg, line 196: This should not have been there in the first place, right?

pulpcore/cli/core/group.py
```diff
@@ -185,6 +198,7 @@ def role(ctx: click.Context, pulp_ctx: PulpCLIContext, group_ctx: PulpGroupConte
                 callback=null_callback,
                 help=_('Filter roles by the associated object. Use "" to list global assignments.'),
             ),
+            domain_option,
```
mdellweg, line 201: Should we add a guard that certain commands can only be used in the default domain?

pulpcore/cli/core/group.py
```diff
@@ -193,13 +207,13 @@ def role(ctx: click.Context, pulp_ctx: PulpCLIContext, group_ctx: PulpGroupConte
         decorators=[
             group_option,
             click.option("--role", required=True),
-            click.option("--object", "content_object", required=True),
             click.option(
                 "--object",
                 "content_object",
-                required=True,
+                default="",
```
mdellweg, line 213: It is required on purpose. Assigning global permissions should not be done by accident.

pulpcore/cli/core/group.py
```diff
@@ -217,11 +231,19 @@ def role(ctx: click.Context, pulp_ctx: PulpCLIContext, group_ctx: PulpGroupConte
...
                 callback=lookup_callback("content_object", PulpGroupRoleContext),
                 expose_value=False,
                 help=_('Associated object; use "" for global assignments.'),
             ),
+            pulp_option(
+                "--domain",
+                # This only accepts HREFs
+                callback=lookup_callback("domain", PulpGroupRoleContext),
```
mdellweg, line 242:
```suggestion
                callback=lookup_callback("domain", PulpDomainContext),
```
does this help to revisit you comment about HREFs?

pulp-glue/pulp_glue/common/context.py
```diff
@@ -90,15 +90,18 @@ def __init__(
...
         api_kwargs: Dict[str, Any],
+        domain: Optional[str],
         format: str,
         background_tasks: bool,
         timeout: int,
     ) -> None:
         self._api: Optional[OpenAPI] = None
-        self.api_path: str = api_root + "api/v3/"
+        self._api_root: str = api_root
+        self._normal_api_path: str = api_root + "api/v3/"
```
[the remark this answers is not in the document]
gerrod3: Yes, the status and docs endpoints are both still available at the normal paths.

pulp-glue/pulp_glue/common/context.py
```diff
@@ -153,13 +156,29 @@ def _patch_api_spec(self) -> None:
...
+
     @property
     def api(self) -> OpenAPI:
         if self._api is None:
             if self._api_kwargs.get("username") and not self._api_kwargs.get("password"):
                 self._api_kwargs["password"] = self.prompt("password", hide_input=True)
             try:
-                self._api = OpenAPI(doc_path=f"{self.api_path}docs/api.json", **self._api_kwargs)
+                self._api = OpenAPI(
+                    doc_path=f"{self._normal_api_path}docs/api.json", **self._api_kwargs
```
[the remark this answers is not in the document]
gerrod3: Yep, api-docs will also be available in each domain.

pulp-glue/pulp_glue/common/context.py
```diff
@@ -188,8 +207,12 @@ def call(
         Returns the operation result, or the finished task.
         If non_blocking, returns unfinished tasks.
         """
-        if parameters is not None:
-            parameters = preprocess_payload(parameters)
+        if parameters is None:
+            parameters = {}
+        if self.domain_enabled:
+            if "pulp_domain" in self.api.param_spec(operation_id, "path", required=True):
+                parameters["pulp_domain"] = self.pulp_domain
```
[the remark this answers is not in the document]
gerrod3: This is correct, detail endpoints don't require the pulp_domain path parameter (since they already have it filled in) so we can't blindly add it to each call.

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -150,6 +150,29 @@ def _set_correlation_id(self, correlation_id: str) -> None:
         else:
             self._session.headers["Correlation-ID"] = correlation_id
 
+    def param_spec(
```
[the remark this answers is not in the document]
gerrod3: Probably, I didn't want to touch the openapi code more than I had to. Maybe in a future refactor.

pulpcore/cli/common/generic.py
gerrod3: This is for user/group role assignment. Assigning an object-level role can have the object be in a different domain than the user/group which are always in the default domain.

pulpcore/cli/core/group.py
gerrod3: Possibly, currently Pulp should disallow this, but having a nice error message from the CLI would be helpful.

pulpcore/cli/core/group.py
gerrod3: Correct, it was  duplicate option.

pulpcore/cli/core/group.py
gerrod3: Added a callback to keep current behavior if domain is not specified.

pulpcore/cli/core/group.py
gerrod3: No, the `lookup_callback` only adds the field to the entity context's lookup and requires that the context exists in the command, which `PulpDomainContext` does not.

 I added more functionality to `resource_option` to allow for specifying the domain name for this command.

pulp-glue/pulp_glue/common/context.py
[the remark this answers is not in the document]
mdellweg: But still we need to assemble the string manually, because the `api_root` property would call to the server giving us an infinite loop. Also when using the non-domained api docs, we only save a local copy once (pretty much preferred).

pulpcore/cli/core/group.py
```diff
@@ -217,11 +247,12 @@ def role(ctx: click.Context, pulp_ctx: PulpCLIContext, group_ctx: PulpGroupConte
             click.option(
                 "--object",
                 "content_object",
-                required=True,
+                default="",
```
mdellweg, line 250:
```suggestion
                callback=_object_required_callback,
```

pulpcore/cli/core/user.py
```diff
@@ -124,11 +154,12 @@ def role(ctx: click.Context, pulp_ctx: PulpCLIContext, user_ctx: PulpUserContext
             click.option(
                 "--object",
                 "content_object",
-                required=True,
+                default="",
```
mdellweg, line 157:
```suggestion
                callback=_object_required_callback,
```

tests/scripts/pulp_file/test_acs.sh
mdellweg:
```suggestion
task_group=$(echo "$ERROUTPUT" | grep -E -o "${PULP_API_ROOT}\(/[_a-zA-Z0-9]*\)?/api/v3/task-groups/[-[:xdigit:]]*/")
```

mdellweg (commented):
I think you missed two places.

tests/scripts/pulpcore/test_pulpimporter.sh
```diff
@@ -5,6 +5,8 @@
 
 pulp debug has-plugin --name "file" || exit 23
 
+eval "$(pulp status | jq -r '.domain_enabled // false | not')" || exit 23
```
mdellweg, line 8: We could have a `pulp debug` command like `pulp debug has-plugin`.

mdellweg (approved):

tests/scripts/pulpcore/test_pulpimporter.sh
ggainey: > We could have a `pulp debug` command like `pulp debug has-plugin`.

@mdellweg  Open an issue pliz!

ggainey (approved):
OK, I think we are good to go - thanks for all the work gerrod!

PR #593, opened 2022-11-08
Add resource_lookup_option to accept identities and HREFs

### LINKED ISSUE #475: Provide option aliases consistent across different subcommands
## Summary
A sub-command about some entity (for example "repositories"), tends to refer to it's name using the option `--name`, while other sub-commands tend to refer to the same kind of entity via the entity name (e.g.: `--repository`)

This is inconvenient (particularly during ad hoc CLI usage), since I want to reuse the option I just typed when changing the sub-command. The idea, is to have option aliases (for backwards compatibility) that are descriptive irrespective of the subcommand (e.g.: `--repository-name`).

## Examples
Sticking with the above example see the difference between having to amend between the following two command pairs:
```
pulp deb repository show --name="test"
pulp deb repository version show --repository="test"
```
and
```
pulp deb repository show --repository-name="test"
pulp deb repository version show --repository-name="test"
```
In the latter case, I can simply add the subcommand, without needing to change the option.
Note that the example is `pulp deb`, but these particular subcommands are from common CLI and presumably used by every plugin.

## Things to consider/next steps
Before starting to code, it is necessary to identify a list of subcommands and options that could likewise benefit from a change like the one I have proposed for just one example. I am sure there are more than just the one.

## Extra Information
I originally formulated this issue as a discussion on discourse: https://discourse.pulpproject.org/t/pulp-cli-usability-discussion/370

lubosmj:
closes #475

- [x] Modify all the necessary commands (besides `show` commands)
- [x] Update tests considering the backward compatibility

### DIFF
--- a/CHANGES/475.feature
+++ b/CHANGES/475.feature
@@ -1,0 +1,2 @@
+Made all commands referencing entities accept both the HREF and name of the resource via the same command option.
+For example, users can additionally use the `--repository` option in ``repository show`` commands.
--- a/pulpcore/cli/ansible/context.py
+++ b/pulpcore/cli/ansible/context.py
@@ -4,7 +4,7 @@ from pulpcore.cli.common.context import (
     EntityDefinition,
     PluginRequirement,
     PulpContentContext,
-    PulpEntityContext,
+    PulpDistributionContext,
     PulpRemoteContext,
     PulpRepositoryContext,
     PulpRepositoryVersionContext,
@@ -41,7 +41,7 @@ class PulpAnsibleCollectionVersionSignatureContext(PulpContentContext):
     ID_PREFIX = "content_ansible_collection_signatures"
 
 
-class PulpAnsibleDistributionContext(PulpEntityContext):
+class PulpAnsibleDistributionContext(PulpDistributionContext):
     ENTITY = _("ansible distribution")
     ENTITIES = _("ansible distributions")
     HREF = "ansible_ansible_distribution_href"
@@ -61,6 +61,7 @@ class PulpAnsibleRoleRemoteContext(PulpRemoteContext):
     ENTITIES = _("role remotes")
     HREF = "ansible_role_remote_href"
     ID_PREFIX = "remotes_ansible_role"
+    HREF_PATTERN = r"remotes/(?P<plugin>ansible)/(?P<resource_type>role)/"
 
 
 class PulpAnsibleCollectionRemoteContext(PulpRemoteContext):
@@ -68,7 +69,7 @@ class PulpAnsibleCollectionRemoteContext(PulpRemoteContext):
     ENTITIES = _("collection remotes")
     HREF = "ansible_collection_remote_href"
     ID_PREFIX = "remotes_ansible_collection"
-    collection_nullable = ["auth_url", "requirements_file", "token"]
+    HREF_PATTERN = r"remotes/(?P<plugin>ansible)/(?P<resource_type>collection)/"
 
     def preprocess_entity(self, body: EntityDefinition, partial: bool = False) -> EntityDefinition:
         body = super().preprocess_entity(body, partial=partial)
--- a/pulpcore/cli/ansible/distribution.py
+++ b/pulpcore/cli/ansible/distribution.py
@@ -16,6 +16,7 @@ from pulpcore.cli.common.generic import (
     create_command,
     destroy_command,
     distribution_filter_options,
+    distribution_lookup_option,
     href_option,
     label_command,
     list_command,
@@ -38,6 +39,7 @@ repository_option = resource_option(
     default_plugin="ansible",
     default_type="ansible",
     context_table={"ansible:ansible": PulpAnsibleRepositoryContext},
+    href_pattern=PulpAnsibleRepositoryContext.HREF_PATTERN,
 )
 
 
@@ -58,7 +60,8 @@ def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: s
         raise NotImplementedError()
 
 
-lookup_options = [href_option, name_option]
+lookup_options = [href_option, name_option, distribution_lookup_option]
+nested_lookup_options = [distribution_lookup_option]
 create_options = [
     click.option("--name", required=True),
     click.option(
@@ -78,9 +81,10 @@ distribution.add_command(destroy_command(decorators=lookup_options))
 distribution.add_command(create_command(decorators=create_options))
 distribution.add_command(
     label_command(
+        decorators=nested_lookup_options,
         need_plugins=[
             PluginRequirement("ansible", "0.8.0"),
-        ]
+        ],
     )
 )
 
@@ -89,8 +93,9 @@ distribution.add_command(
 @distribution.command()
 @name_option
 @href_option
+@distribution_lookup_option
 @click.option("--base-path", help=_("new base_path"))
-@click.option("--repository", type=str, default=None, help=_("new repository to be served"))
+@repository_option
 @click.option(
     "--version",
     type=int,
--- a/pulpcore/cli/ansible/repository.py
+++ b/pulpcore/cli/ansible/repository.py
@@ -36,7 +36,7 @@ from pulpcore.cli.common.generic import (
     pulp_option,
     repository_content_command,
     repository_href_option,
-    repository_option,
+    repository_lookup_option,
     resource_option,
     retained_versions_option,
     show_command,
@@ -102,8 +102,8 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) -> 
         raise NotImplementedError()
 
 
-lookup_options = [href_option, name_option]
-nested_lookup_options = [repository_href_option, repository_option]
+lookup_options = [href_option, name_option, repository_lookup_option]
+nested_lookup_options = [repository_href_option, repository_lookup_option]
 update_options = [
     click.option("--description"),
     pulp_option(
@@ -174,8 +174,8 @@ repository.add_command(create_command(decorators=create_options))
 repository.add_command(update_command(decorators=lookup_options + update_options))
 repository.add_command(destroy_command(decorators=lookup_options))
 repository.add_command(task_command(decorators=nested_lookup_options))
-repository.add_command(version_command())
-repository.add_command(label_command())
+repository.add_command(version_command(decorators=nested_lookup_options))
+repository.add_command(label_command(decorators=nested_lookup_options))
 repository.add_command(
     repository_content_command(
         contexts={
@@ -192,6 +192,7 @@ repository.add_command(
 @repository.command()
 @name_option
 @href_option
+@repository_lookup_option
 @remote_option
 @pass_repository_context
 def sync(
@@ -223,6 +224,7 @@ def sync(
 @repository.command()
 @name_option
 @href_option
+@repository_lookup_option
 @click.option("--signing-service", required=True, callback=_signing_service_callback)
 @click.option("--content-units", callback=load_json_callback)
 @pass_repository_context
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -412,6 +412,7 @@ class PulpEntityContext:
     #           ]
     #       }
     CAPABILITIES: ClassVar[Dict[str, List[PluginRequirement]]] = {}
+    HREF_PATTERN: str
 
     # Hidden values for the lazy entity lookup
     _entity: Optional[EntityDefinition]
@@ -704,7 +705,7 @@ class PulpRemoteContext(PulpEntityContext):
     ENTITY = _("remote")
     ENTITIES = _("remotes")
     ID_PREFIX = "remotes"
-    HREF_PATTERN = r"remotes/(?P<plugin>\w+)/(?P<resource_type>\w+)/"
+    HREF_PATTERN = r"remotes/(?P<plugin>[\w\-_]+)/(?P<resource_type>[\w\-_]+)/"
     NULLABLES = {
         "ca_cert",
         "client_cert",
@@ -722,6 +723,13 @@ class PulpRemoteContext(PulpEntityContext):
         "sock_read_timeout",
         "rate_limit",
     }
+
+
+class PulpDistributionContext(PulpEntityContext):
+    ENTITY = _("distribution")
+    ENTITIES = _("distributions")
+    ID_PREFIX = "distributions"
+    HREF_PATTERN = r"distributions/(?P<plugin>[\w\-_]+)/(?P<resource_type>[\w\-_]+)/"
 
 
 class PulpRepositoryVersionContext(PulpEntityContext):
@@ -756,7 +764,7 @@ class PulpRepositoryContext(PulpEntityContext):
 
     ENTITY = _("repository")
     ENTITIES = _("repositories")
-    HREF_PATTERN = r"repositories/(?P<plugin>\w+)/(?P<resource_type>\w+)/"
+    HREF_PATTERN = r"repositories/(?P<plugin>[\w\-_]+)/(?P<resource_type>[\w\-_]+)/"
     ID_PREFIX = "repositories"
     VERSION_CONTEXT: ClassVar[Type[PulpRepositoryVersionContext]]
     NULLABLES = {"description", "retain_repo_versions"}
@@ -806,6 +814,8 @@ class PulpContentContext(PulpEntityContext):
 class PulpACSContext(PulpEntityContext):
     ENTITY = _("ACS")
     ENTITIES = _("ACSes")
+    HREF_PATTERN = r"acs/(?P<plugin>[\w\-_]+)/(?P<resource_type>[\w\-_]+)/"
+    ID_PREFIX = "acs"
 
     def refresh(self, href: Optional[str] = None) -> Any:
         return self.call("refresh", parameters={self.HREF: href or self.pulp_href})
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -15,11 +15,14 @@ from pulpcore.cli.common.context import (
     EntityDefinition,
     EntityFieldDefinition,
     PluginRequirement,
+    PulpACSContext,
     PulpContentContext,
     PulpContext,
+    PulpDistributionContext,
     PulpEntityContext,
     PulpException,
     PulpNoWait,
+    PulpRemoteContext,
     PulpRepositoryContext,
     PulpRepositoryVersionContext,
 )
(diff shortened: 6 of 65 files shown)


pulpcore/cli/common/generic.py
```diff
@@ -414,6 +414,7 @@ def resource_option(*args: Any, **kwargs: Any) -> Callable[[FC], FC]:
     context_table: Dict[str, Type[PulpEntityContext]] = kwargs.pop("context_table")
     capabilities: Optional[List[str]] = kwargs.pop("capabilities", None)
     href_pattern: Optional[str] = kwargs.pop("href_pattern", None)
+    is_command_entity: Optional[bool] = kwargs.pop("is_command_entity", False)
```
mdellweg, line 417: I don't think, a new argument is needed here. Can we just take `expose_value` for it?
If `expose_value=True` we return a new context, else we find and polulate an existing context.

pulpcore/cli/common/generic.py
```diff
@@ -509,23 +518,41 @@ def _multi_option_callback(
...
-            plugin_form=_("[<plugin>:]") if default_plugin else _("<plugin>:"),
-            type_form=_("[<resource_type>:]") if default_type else _("<resource_type>:"),
-            plugin_default=_("'<plugin>' defaults to {plugin}. ").format(plugin=default_plugin)
-            if default_plugin
-            else "",
-            type_default=_("'<resource_type>' defaults to {type}. ").format(type=default_type)
-            if default_type
-            else "",
-        )
+        kwargs["help"] = init_help_message(default_plugin, default_type)
```
mdellweg, line 521: I don't see the need here.

pulpcore/cli/file/repository.py
```diff
@@ -55,9 +55,19 @@
     default_type="file",
     context_table={"file:file": PulpFileRemoteContext},
     href_pattern=PulpRemoteContext.HREF_PATTERN,
+    help=_("Remote used for syncing in the form '[[<plugin>:]<resource_type>:]<name>' or by href."),
+)
+ref_repository_option = resource_option(
```
mdellweg, line 60: Maybe call it repository_lookup_option?

pulpcore/cli/common/generic.py
```diff
@@ -481,7 +482,15 @@ def _option_callback(
                     "is not valid for the {option_name} option."
                 ).format(plugin=plugin, resource_type=resource_type, option_name=param.name)
             )
-        entity_ctx: PulpEntityContext = context_class(pulp_ctx, pulp_href=pulp_href, entity=entity)
+
+        if is_command_entity:
+            entity_ctx: PulpEntityContext = init_command_entity_context(
```
mdellweg, line 487: I think we should not make this a function. It'd be easier to see what happens if it just happened here.

lubosmj: I wanted to make the function more readable. That is why I extracted "init_help_message" from it. I see you do not like it. ⛔

mdellweg: I see the intent, but not the desired result. Sorry.
As for the help message, yes that has grown into an almost unreadable state. But I don't see you changing it. So I would urge you to not touch it here. It's easier to discuss these refactoring steps on separate PR's.

mdellweg:
BTW. I think, technically this is exactly the solution I was hoping for.

CHANGES/475.feature
```diff
@@ -0,0 +1 @@
+Made all commands referencing entities accept both their HREFs and names via relevant command options. For example, users can additionaly start using option ``--repository`` in the ``repository-show`` command.
```
mdellweg, line 1:
```suggestion
Made all commands referencing entities accept both their HREFs and names via relevant command options.
For example, users can additionally start using ``--repository`` in the ``repository show`` command.
```

pulpcore/cli/ansible/repository.py
```diff
@@ -65,6 +64,17 @@
         "Remote used for synching in the form '[[<plugin>:]<resource_type>:]<name>' or by href."
     ),
 )
+repository_option = resource_option(
```
mdellweg, line 67: I'm wondering if calling these `repository_lookup_option` would prevent confusion with the `repository_option` we use e.g. with distribution.

pulpcore/cli/common/generic.py
```diff
@@ -414,6 +414,7 @@ def resource_option(*args: Any, **kwargs: Any) -> Callable[[FC], FC]:
     context_table: Dict[str, Type[PulpEntityContext]] = kwargs.pop("context_table")
     capabilities: Optional[List[str]] = kwargs.pop("capabilities", None)
     href_pattern: Optional[str] = kwargs.pop("href_pattern", None)
+    expose_value: Optional[bool] = kwargs.get("expose_value", True)
```
mdellweg, line 417: By specifying the default (`True`) we can be sure it _is_ a boolean.
```suggestion
    expose_value: bool = kwargs.get("expose_value", True)
```

mdellweg: Also do we need to extract it here? Can we maybe just query `param.expose_value` in the callback?

pulpcore/cli/ansible/repository.py
lubosmj: I would leave it as it is. I consider it to be the same option like we use for distributions. Were you referring to something else?

```
pulp file repository create --name bbb
pulp file repository show --repository bbb
pulp file distribution create --name bbb --base-path bbb --repository bbb
```

mdellweg: It should look the same on the outside, yes. But to the cli code inside it is something quite different. Maybe to make it more obvious, my suggestion is:
```suggestion
repository_lookup_option = resource_option(
```

mdellweg (commented):
Not sure if i saw too much similarities in the beginning, but this PR is getting a little bit out of control.
What do you think about a better targeted `resource_lookup_option`.
It does not need the context-table, and instead gets the base context type (like `PulpRepositoryContext` for generic repository commands). Also the `_multi_option_callback` is worthless in that context.

CHANGES/475.feature
```diff
@@ -0,0 +1,2 @@
+Made all commands referencing entities accept both their HREFs and names via relevant command options.
+For example, users can additionaly start using option ``--repository`` in the ``repository-show`` command.
```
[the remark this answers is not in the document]
lubosmj: It makes more sense to me to stick to the standard man pages, where common commands are separated by a dash (e.g., podman-manifest-create, https://docs.podman.io/en/latest/markdown/podman-manifest-create.1.html). What do you think?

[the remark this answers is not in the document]
mdellweg: Never thought of this. We should have it consistent in the changelogs at least...

pulpcore/cli/ansible/distribution.py
```diff
@@ -38,6 +39,15 @@
...
     default_type="ansible",
     context_table={"ansible:ansible": PulpAnsibleRepositoryContext},
+    href_pattern=PulpAnsibleRepositoryContext.HREF_PATTERN,
+)
+distribution_option = resource_option(
+    "--distribution",
+    default_plugin="ansible",
+    default_type="ansible",
+    context_table={"ansible:ansible": PulpAnsibleDistributionContext},
+    href_pattern=PulpDistributionContext.HREF_PATTERN,
```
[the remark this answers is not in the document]
lubosmj: `PulpAnsibleDistributionContext` does not inherit from `PulpDistributionContext`. `HREF_PATTERN` is not available there. That is similar for all distribution context classes. Not sure if it has any meaning. May I change the parent classes for the distribution context classes across the plugins?

Still, I think we are good with a generic variant: `HREF_PATTERN = r"distributions/(?P<plugin>\w+)/(?P<resource_type>\w+)/".`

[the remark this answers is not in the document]
mdellweg: Interesting! Yes, they should be safe to inherit from the common base.

pulpcore/cli/common/generic.py
```diff
@@ -448,6 +448,65 @@ def pulp_option(*args: Any, **kwargs: Any) -> Callable[[FC], FC]:
     return click.option(*args, **kwargs)
 
 
+def resource_lookup_option(*args: Any, **kwargs: Any) -> Callable[[FC], FC]:
+    lookup_key: str = kwargs.pop("lookup_key", "name")
+    href_pattern: Optional[str] = kwargs.pop("href_pattern", None)
```
mdellweg, line 453: I would pass the Context class instead of the href_pattern to this function ...

pulpcore/cli/common/generic.py
```diff
@@ -448,6 +448,65 @@ def pulp_option(*args: Any, **kwargs: Any) -> Callable[[FC], FC]:
...
+                )
+
+            pulp_href = value
+        else:
+            # A natural key identifier was passed
+            identifier = value.split(":", maxsplit=2)[-1]
+            entity = {lookup_key: identifier}
+
+        if entity:
+            ctx.obj.entity = entity
```
mdellweg, line 492: and use `ctx.find_object(context_class)` here instead. This is the way we find the _right_ entity context and not only the _last_ one. The requirement being that the context class has or inherits an `href_pattern` obviously.

pulpcore/cli/common/generic.py
lubosmj: I decided to pass there a list of context classes. Now, I am iterating over the list to find the best matching entity context. It was required to handle commands like:
```
pulp container repository show --repository foo
pulp container repository -t push show --repository foo
pulp container repository -t push show --repository pushed-foo
```

mdellweg: I'm not getting that. All the contexts there should be subclasses of `PulpRepositoryContext` and therefore be the one we find. So `ctx.get_object(PulpRepositoryContext).HREF_PATTERN` is also the only pattern we need to match against, no?

lubosmj: You were totally right. I updated the code a bit.

pulpcore/cli/common/context.py
```diff
@@ -701,7 +716,7 @@ class PulpRepositoryContext(PulpEntityContext):
 
     ENTITY = _("repository")
     ENTITIES = _("repositories")
-    HREF_PATTERN = r"repositories/(?P<plugin>\w+)/(?P<resource_type>\w+)/"
+    HREF_PATTERN = r"repositories/(?P<plugin>\w+)/(?P<resource_type>[\w\-_]+)/"
```
lubosmj, line 719: This was required for pulp_container where we have the `/repositories/container/container-push/` endpoint.

mdellweg: Can plugins have dashes too? I'd update the in all cases.

mdellweg (commented):
Given all of the above comments, we are really closing in on the final design.

As for the tests, can i ask you to try not to create any extra resources? I don't want the runtime of the tests to increase significantly. I think otoh it would be ok to say that the new parameter is the standard way to access resources. So you could replace almost all appearances of `--name` and `--href` and keep only a couple of the old ones to ensure they still work.

pulpcore/cli/core/context.py
```diff
@@ -254,13 +254,6 @@ class PulpContentRedirectContentGuardContext(PulpEntityContext):
     NEEDS_PLUGINS = [PluginRequirement("core", "3.18.0")]
 
 
-class PulpDistributionContext(PulpEntityContext):
-    ENTITY = _("distribution")
-    ENTITIES = _("distributions")
-    ID_PREFIX = "distributions"
-    HREF_PATTERN = r"distributions/(?P<plugin>\w+)/(?P<resource_type>\w+)/"
```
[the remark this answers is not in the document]
lubosmj: I was receiving the following errors before moving it there:
```
ImportError: cannot import name 'PulpCLIContext' from partially initialized module 'pulpcore.cli.common.generic' (most likely due to a circular import) (/home/vagrant/devel/pulp-cli/pulpcore/cli/common/generic.py)
```

pulpcore/cli/ansible/context.py
```diff
@@ -32,6 +32,7 @@ class PulpAnsibleRoleContext(PulpContentContext):
     ENTITIES = _("ansible roles")
     HREF = "ansible_role_href"
     ID_PREFIX = "content_ansible_roles"
+    HREF_PATTERN = r"remotes/(?P<plugin>ansible)/(?P<resource_type>role)/"
```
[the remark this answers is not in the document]
lubosmj: Oh, nice find. I wanted to define it for `PulpAnsibleRoleRemoteContext`.

pulpcore/cli/ansible/context.py
```diff
@@ -69,6 +70,7 @@ class PulpAnsibleCollectionRemoteContext(PulpRemoteContext):
     HREF = "ansible_collection_remote_href"
     ID_PREFIX = "remotes_ansible_collection"
     collection_nullable = ["auth_url", "requirements_file", "token"]
+    HREF_PATTERN = r"remotes/(?P<plugin>ansible)/(?P<resource_type>collection)/"
```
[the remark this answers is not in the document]
lubosmj: Yes, exactly. Similarly to pulp_container, we want to prevent users from using HREF that does not refer to a currently used context.
```
pulp container repository -t push show --repository /pulp/api/v3/repositories/container/container/7864a76f-5d5b-4b7b-9587-2fe076abf4f0/
Error: '/pulp/api/v3/repositories/container/container/7864a76f-5d5b-4b7b-9587-2fe076abf4f0/' is not a valid href for repository.
```
 ^ the CLI expects repositories/container/container-push/

pulpcore/cli/container/context.py
```diff
@@ -48,9 +49,10 @@ class PulpContainerNamespaceContext(PulpEntityContext):
     HREF = "container_container_namespace_href"
     ID_PREFIX = "pulp_container_namespaces"
     CAPABILITIES = {"roles": [PluginRequirement("container", "2.11.0")]}
+    HREF_PATTERN = r"(?P<plugin>pulp_container)/(?P<resource_type>namespaces)/"
```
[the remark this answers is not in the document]
lubosmj: This is more or less an artefact from https://github.com/pulp/pulp-cli/blob/70b6979c8a9e5bd89b3a3cb296b1580b5e9cf5f4/pulpcore/cli/common/context.py#L759. We still check for the resource type in `resource_option`. It is not currently used anywhere, but, I left those matching named groups untouched, similarly to pulp_ansible remotes.

pulpcore/cli/python/remote.py
```diff
@@ -29,6 +30,13 @@
 _ = translation.gettext
 
 
+remote_lookup_option = resource_lookup_option(
+    "--remote",
+    needs_plugins=[PluginRequirement("python", "3.6.0")],
```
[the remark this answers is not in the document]
lubosmj: From https://github.com/pulp/pulp-cli/blob/70b6979c8a9e5bd89b3a3cb296b1580b5e9cf5f4/pulpcore/cli/python/distribution.py#L50. I am removing it, sorry.

pulpcore/cli/core/context.py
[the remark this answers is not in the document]
mdellweg: Yes, and that is fine. I was just adding this comment to not be confused again.

pulpcore/cli/common/context.py
lubosmj: I added both the dash and underscore to all default regexes.

pulpcore/cli/container/context.py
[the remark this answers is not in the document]
mdellweg: What happens if we do this?
```suggestion
    HREF_PATTERN = r"pulp_container/namespaces/"
```

pulpcore/cli/ansible/repository.py
```diff
@@ -176,7 +176,7 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
 repository.add_command(destroy_command(decorators=lookup_options))
 repository.add_command(task_command(decorators=nested_lookup_options))
 repository.add_command(version_command())
-repository.add_command(label_command())
+repository.add_command(label_command(decorators=lookup_options))
```
mdellweg, line 179: should be:
```suggestion
repository.add_command(label_command(decorators=nested_lookup_options))
```
...until we removed the _old_ lookup options completely.

pulpcore/cli/container/context.py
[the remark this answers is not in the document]
lubosmj: Nothing will change now. It is there for the sake of possible future use because we have got the `resource_option` option that reads the matched groups and initializes a corresponding resource (entity_ctx) class further on by relying on the data in those groups.

https://github.com/pulp/pulp-cli/blob/70b6979c8a9e5bd89b3a3cb296b1580b5e9cf5f4/pulpcore/cli/common/generic.py#L489-L490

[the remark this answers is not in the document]
mdellweg: Well if anything, in any case, the plugin should be "container", and the resource_type is unknown or also "container".
But i see that the extraction there is at least resiliant to missing matches.

pulpcore/cli/ansible/repository.py
```diff
@@ -175,8 +175,8 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
 repository.add_command(update_command(decorators=lookup_options + update_options))
 repository.add_command(destroy_command(decorators=lookup_options))
 repository.add_command(task_command(decorators=nested_lookup_options))
-repository.add_command(version_command())
-repository.add_command(label_command())
+repository.add_command(version_command(decorators=nested_lookup_options))
+repository.add_command(label_command(decorators=nested_lookup_options))
```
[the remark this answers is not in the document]
lubosmj: Seems like the change is already present there:
```
pulpcore/cli/ansible/repository.py:repository.add_command(task_command(decorators=nested_lookup_options))
pulpcore/cli/ansible/repository.py:repository.add_command(version_command(decorators=nested_lookup_options))
pulpcore/cli/ansible/repository.py:repository.add_command(label_command(decorators=nested_lookup_options))
pulpcore/cli/container/repository.py:repository.add_command(task_command(decorators=nested_lookup_options))
pulpcore/cli/container/repository.py:repository.add_command(version_command(decorators=nested_lookup_options))
pulpcore/cli/container/repository.py:repository.add_command(label_command(decorators=nested_lookup_options))
pulpcore/cli/file/repository.py:repository.add_command(task_command(decorators=nested_lookup_options))
pulpcore/cli/file/repository.py:repository.add_command(version_command(decorators=nested_lookup_options))
pulpcore/cli/file/repository.py:repository.add_command(label_command(decorators=nested_lookup_options))
pulpcore/cli/python/repository.py:repository.add_command(task_command(decorators=nested_lookup_options))
pulpcore/cli/python/repository.py:repository.add_command(version_command(decorators=nested_lookup_options))
pulpcore/cli/python/repository.py:repository.add_command(label_command(decorators=nested_lookup_options))
pulpcore/cli/rpm/repository.py:repository.add_command(task_command(decorators=nested_lookup_options))
pulpcore/cli/rpm/repository.py:repository.add_command(version_command(decorators=nested_lookup_options))
pulpcore/cli/rpm/repository.py:repository.add_command(label_command(decorators=nested_lookup_options))
```

pulpcore/cli/ansible/distribution.py
```diff
@@ -78,9 +80,10 @@ def distribution(ctx: click.Context, pulp_ctx: PulpContext, distribution_type: s
 distribution.add_command(create_command(decorators=create_options))
 distribution.add_command(
     label_command(
+        decorators=lookup_options,
```
mdellweg, line 83: This should be `nested_lookup_options`.

pulpcore/cli/common/context.py
```diff
@@ -724,6 +725,20 @@ class PulpRemoteContext(PulpEntityContext):
     }
 
 
+class PulpAcsContext(PulpEntityContext):
+    ENTITY = _("ACS")
+    ENTITIES = _("ACSes")
+    HREF_PATTERN = r"acs/(?P<plugin>[\w\-_]+)/(?P<resource_type>[\w\-_]+)/"
+    ID_PREFIX = "acs"
```
mdellweg, line 732: Sorry for the silent merge conflict here. This class is (by this time) already defined at the end of the file.

mdellweg (changes requested):
Almost there. Just the duplicate ACS context left.

pulpcore/cli/common/context.py
```diff
@@ -806,6 +814,8 @@ class PulpContentContext(PulpEntityContext):
 class PulpACSContext(PulpEntityContext):
     ENTITY = _("ACS")
     ENTITIES = _("ACSes")
+    HREF_PATTERN = r"acs/(?P<plugin>[\w\-_]+)/(?P<resource_type>[\w\-_]+)/"
+    ID_PREFIX = "acs"
```
mdellweg, line 818: :+1:

mdellweg (approved):

PR #598, opened 2022-11-11
Click separation

mdellweg:
Review Checklist:

### DIFF
--- a/CHANGES/597.feature
+++ b/CHANGES/597.feature
@@ -1,0 +1,1 @@
+Made the context layer independent of click to allow it being used like a library.
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -1,27 +1,14 @@
 import datetime
-import json
 import re
 import sys
 import time
 from typing import Any, ClassVar, Dict, Iterable, List, NamedTuple, Optional, Set, Type, Union
 
-import click
-import yaml
 from packaging.version import parse as parse_version
 from requests import HTTPError
 
 from pulpcore.cli.common.i18n import get_translation
 from pulpcore.cli.common.openapi import OpenAPI, OpenAPIError
-
-try:
-    from pygments import highlight
-    from pygments.formatters import Terminal256Formatter
-    from pygments.lexers import JsonLexer, YamlLexer
-except ImportError:
-    PYGMENTS = False
-else:
-    PYGMENTS = True
-    PYGMENTS_STYLE = "solarized-dark"
 
 translation = get_translation(__name__)
 _ = translation.gettext
@@ -61,24 +48,12 @@ class PluginRequirement(NamedTuple):
     inverted: bool = False
 
 
-# This is deprecated we keep it to not break plugins immediately
-new_component_names_to_pre_3_11_names: Dict[str, str] = {}
-
-
 class PulpException(Exception):
     pass
 
 
 class PulpNoWait(Exception):
     pass
-
-
-class PulpJSONEncoder(json.JSONEncoder):
-    def default(self, obj: Any) -> Any:
-        if isinstance(obj, datetime.datetime):
-            return obj.isoformat()
-        else:
-            return super().default(obj)
 
 
 def _preprocess_value(value: Any) -> Any:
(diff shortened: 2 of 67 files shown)


pytest_pulp_cli/__init__.py
```diff
@@ -17,7 +17,7 @@ class ScriptFile(pytest.File):
     obj = None
 
     def collect(self):
-        name = self.path.name[5:]
+        name = self.path.name[5:][:-3]
```
ggainey, line 20: I don't think I understand what this change is doing? Can you explain? (and maybe add a comment so I won't ask you the same question in a few months :) )

tests/scripts/test_config.sh
```diff
@@ -14,29 +14,29 @@ export XDG_CONFIG_HOME=/nowhere
 
 sed -e '/base_url/c\base_url = "http://badurl"' "$good_settings" > "$bad_settings"
 
-expect_succ pulp --config "$good_settings" file repository list
-expect_succ pulp --config "$bad_settings" --base-url "$PULP_BASE_URL" file repository list
+expect_succ pulp --config "$good_settings" task list
```
ggainey, line 17: I assume this section is to clean up the test so it doesn't need a plugin to test the configs, yeah?

ggainey (changes requested):
Looks like a scary PR, but 90%+ is just the tediousness of changing import-statements. I have one question/change, but it looks really good, thanks!

pulpcore/cli/common/generic.py
```diff
@@ -66,31 +132,6 @@ def float_or_empty(value: str) -> Union[str, float]:
...
-class PulpCLIContext(PulpContext):
-    """
-    Subclass of the Context that overwrites the CLI specifics.
-    """
-
-    def echo(self, message: str, nl: bool = True, err: bool = False) -> None:
-        click.echo(message, nl=nl, err=err)
-
-    def prompt(self, text: str, hide_input: bool = False) -> Any:
-        return click.prompt(text, hide_input=hide_input)
```
gerrod3, line 91: How was this working before? `pass_pulp_context` was making a decorator from `PulpContext`, how was any command able to output to the terminal?

pulpcore/cli/common/context.py
```diff
@@ -797,18 +751,6 @@ class PulpContentContext(PulpEntityContext):
...
-##############################################################################
-# Decorator to access certain contexts
-# DEPRECATED These are moved to generic.py and will be removed here.
-
-
-pass_pulp_context = click.make_pass_decorator(PulpContext)
-pass_entity_context = click.make_pass_decorator(PulpEntityContext)
-pass_repository_context = click.make_pass_decorator(PulpRepositoryContext)
-pass_repository_version_context = click.make_pass_decorator(PulpRepositoryVersionContext)
-pass_content_context = click.make_pass_decorator(PulpContentContext)
```
gerrod3, line 809: Have we checked the other CLI plugins to see if they have switched over to the correct import spot?

pulpcore/cli/common/context.py
```diff
@@ -61,10 +48,6 @@ class PluginRequirement(NamedTuple):
     inverted: bool = False
 
 
-# This is deprecated we keep it to not break plugins immediately
-new_component_names_to_pre_3_11_names: Dict[str, str] = {}
```
gerrod3, line 65: Same here, have all CLI plugins updated to not use this anymore?

CHANGES/597.feature
```diff
@@ -0,0 +1 @@
+Made the context layer independent of click to allow it being used like a library.
```
gerrod3, line 1: What's the next step in the plan? Move into separate repository or have other projects inherit from the CLI?

gerrod3 (commented):
Some questions I have for this large change, but everything looks good.

pytest_pulp_cli/__init__.py
mdellweg: Technically this is not even related. But it was bugging me.

```suggestion
        # Extract the name between "test_" and ".sh".
        name = self.path.name[5:][:-3]
```

tests/scripts/test_config.sh
mdellweg: Yes, i found this after running against a pulp without pulp_file. Uhg...
It was either this or adding a test requirement on pulp_file. But why skip a test for a bad reason.

pulpcore/cli/common/generic.py
mdellweg: `pass_pulp_context` would return the first found **subclass of `PulpContext`** context object in the context hierarchy. This subclass happened to be the cli specific type.

pulpcore/cli/common/context.py
mdellweg: I switched them last week.

pulpcore/cli/common/context.py
mdellweg: I'll check.

CHANGES/597.feature
mdellweg: That **is** the right question to ask. As for this change alone, it is a step towards better architectural separation of concerns. The next step i plan is to see if i can remove `echo` and `prompt` from `PulpContext`.
After that, it is either of:
1. Keep everthing as is, and install e.g. click when all you want is the library part. (unnecessary potential for version conflicts)
2. Make click and others an optional dependency of the cli. (kind of weird to install `pulp-cli[cli]`...)
3. Merge the cli-plugins and split into two repos/pypi packages.
4. Split the cli into 6+ repos/packages.

I'm really not ready to decide.

gerrod3 (approved):

PR #599, opened 2022-11-14
Validators

mdellweg:
Review Checklist:

### DIFF
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -2,19 +2,7 @@ import datetime
 import re
 import sys
 import time
-from typing import (
-    Any,
-    ClassVar,
-    Dict,
-    Iterable,
-    List,
-    Mapping,
-    NamedTuple,
-    Optional,
-    Set,
-    Type,
-    Union,
-)
+from typing import Any, ClassVar, Dict, List, Mapping, NamedTuple, Optional, Set, Type, Union
 
 from packaging.version import parse as parse_version
 from requests import HTTPError
@@ -69,13 +57,11 @@ class PulpNoWait(Exception):
 
 
 def _preprocess_value(value: Any) -> Any:
-    if isinstance(value, str):
-        return value
     if isinstance(value, PulpEntityContext):
         return value.pulp_href
     if isinstance(value, Mapping):
         return {k: _preprocess_value(v) for k, v in value.items()}
-    if isinstance(value, Iterable):
+    if isinstance(value, (list, tuple)):
         return [_preprocess_value(item) for item in value]
     return value
 
@@ -150,6 +136,11 @@ class PulpContext:
                             and parameter["schema"]["type"] == "string"
                         ):
                             parameter["schema"] = {"type": "array", "items": {"type": "string"}}
+        if self.has_plugin(PluginRequirement("file", max="1.11.0")):
+            operation = api_spec["paths"]["{file_file_alternate_content_source_href}refresh/"][
+                "post"
+            ]
+            operation.pop("requestBody")
         if self.has_plugin(
             PluginRequirement("python", max="99.99.0.dev")
         ):  # TODO Add version bounds
@@ -202,12 +193,22 @@ class PulpContext:
             parameters = preprocess_payload(parameters)
         if body is not None:
             body = preprocess_payload(body)
+        # ---- SNIP ----
+        # In the future we want everything as part of the body.
+        if uploads:
+            self.echo(
+                _("Deprecated use of 'uploads' for operation '{operation_id}'").format(
+                    operation_id=operation_id
+                )
+            )
+            body = body or {}
+            body.update(uploads)
+        # ---- SNIP ----
         try:
             result = self.api.call(
                 operation_id,
                 parameters=parameters,
                 body=body,
-                uploads=uploads,
                 validate_body=validate_body,
             )
         except OpenAPIError as e:
@@ -743,7 +744,7 @@ class PulpRepositoryVersionContext(PulpEntityContext):
         return {self.repository_ctx.HREF: self.repository_ctx.pulp_href}
 
     def repair(self, href: Optional[str]) -> Any:
-        return self.call("repair", parameters={self.HREF: href or self.pulp_href})
+        return self.call("repair", parameters={self.HREF: href or self.pulp_href}, body={})
 
 
 class PulpRepositoryContext(PulpEntityContext):
--- a/pulpcore/cli/common/context.py
+++ b/pulpcore/cli/common/context.py
@@ -2,13 +2,25 @@ import datetime
 import re
 import sys
 import time
-from typing import Any, ClassVar, Dict, Iterable, List, NamedTuple, Optional, Set, Type, Union
+from typing import (
+    Any,
+    ClassVar,
+    Dict,
+    Iterable,
+    List,
+    Mapping,
+    NamedTuple,
+    Optional,
+    Set,
+    Type,
+    Union,
+)
 
 from packaging.version import parse as parse_version
 from requests import HTTPError
 
 from pulpcore.cli.common.i18n import get_translation
-from pulpcore.cli.common.openapi import OpenAPI, OpenAPIError
+from pulpcore.cli.common.openapi import OpenAPI, OpenAPIError, UploadsMap
 
 translation = get_translation(__name__)
 _ = translation.gettext
@@ -61,8 +73,8 @@ def _preprocess_value(value: Any) -> Any:
         return value
     if isinstance(value, PulpEntityContext):
         return value.pulp_href
-    if isinstance(value, datetime.datetime):
-        return value.strftime(DATETIME_FORMATS[0])
+    if isinstance(value, Mapping):
+        return {k: _preprocess_value(v) for k, v in value.items()}
     if isinstance(value, Iterable):
         return [_preprocess_value(item) for item in value]
     return value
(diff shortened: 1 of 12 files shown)


pulpcore/cli/core/context.py
```diff
@@ -117,7 +118,7 @@ def HREF(self) -> str:  # type:ignore
 
 class PulpGroupPermissionContext(PulpEntityContext):
     ENTITY = _("group permission")
-    ENTITIES = _("group permissions")
+    ENTITIES = _("group permissiopulpcore/cli/common/debug.py:118:ns")
```
ggainey, line 121: Um...paste-typo?

pulpcore/cli/common/openapi.py
```diff
@@ -14,7 +16,11 @@
 translation = get_translation(__name__)
 _ = translation.gettext
 
+UploadType = Union[bytes, IO[bytes]]
```
ggainey, line 19: Is there a reason not to get this from common/context?

ggainey (changes requested):
A couple of questions, and I'm still working on grokking the full changes. Def needds more eyes than me.

pulpcore/cli/common/openapi.py
```diff
@@ -164,54 +170,117 @@ def extract_params(
...
+            return value
+        else:
+            if not isinstance(value, str):
+                raise OpenAPIError(_("'{name}' is expected to be a string.").format(name=name))
+            return value
+
+    def validate_integer(self, schema: Any, name: str, value: Any) -> int:
+        if not isinstance(value, int):
+            raise OpenAPIError(_("'{name}' is expected to be an integer.").format(name=name))
+        return value
```
gerrod3, line 283: Can we try converting to `int` before raising an error? Also, can we have min/max value checks here?

pulpcore/cli/common/openapi.py
```diff
@@ -164,54 +170,117 @@ def extract_params(
...
+            value = self.validate_object(schema, name, value)
+        elif schema_type == "array":
+            value = self.validate_array(schema, name, value)
+        elif schema_type == "string":
+            value = self.validate_string(schema, name, value)
+        elif schema_type == "integer":
+            value = self.validate_integer(schema, name, value)
+        elif schema_type == "boolean":
+            value = self.validate_integer(schema, name, value)
+            if not isinstance(value, bool):
```
gerrod3, line 222: `validate_integer` checks if the value is an instance of an `int`? Won't this fail for a `bool`?

pulpcore/cli/common/openapi.py
```diff
@@ -164,54 +170,117 @@ def extract_params(
...
+                raise OpenAPIError(_("'{name}' is expected to be a boolean.").format(name=name))
+        # TODO: Add more types here.
+        return value
```
gerrod3, line 225: Can we add a check for floats/doubles? https://swagger.io/specification/#data-types

pulpcore/cli/common/openapi.py
```diff
@@ -164,54 +170,117 @@ def extract_params(
...
+            return value.strftime(ISO_DATETIME_FORMAT)
+        elif schema_format == "binary":
+            if not isinstance(value, (bytes, BufferedReader)):
+                print(repr(value))
+                raise OpenAPIError(_("'{name}' is expected to be binary.").format(name=name))
+            return value
+        else:
+            if not isinstance(value, str):
+                raise OpenAPIError(_("'{name}' is expected to be a string.").format(name=name))
+            return value
```
gerrod3, line 278: Should we be concern about the `byte` format? https://swagger.io/specification/#data-types. The comment says base64 encoded characters so maybe nothing else needs to be added, but I'm not sure.

pulpcore/cli/common/openapi.py
```diff
@@ -279,11 +348,21 @@ def render_request(
                 else:
                     raise OpenAPIError(_("No suitable content type for request specified."))
         if content_type and validate_body:
-            self.validate_body(
+            body = body or {}
+            if uploads:
+                # Add uploads for validation (CRAZY HACK)
+                body.update(uploads)
+            # This only works, because data or json will be the same object as body...
```
gerrod3, line 355: Not sure what the second comment means, is it referring to the fact that `validate_schema` returns the same dict object passed in with each key-value pair validated?

As for the hack, not sure what else we can do besides preprocess out the upload fields into a separate schema to check against, but that requires all upload fields to be of the same type. (string: format-binary?)

pulpcore/cli/common/context.py
```diff
@@ -133,6 +147,49 @@ def __init__(
         self.timeout: int = timeout
         self.start_time: Optional[datetime.datetime] = None
 
+    def _patch_api_spec(self) -> None:
+        # A place for last minute fixes to the api_spec.
+        # WARNING: Operations are already indexed at this point.
+        api_spec = self.api.api_spec
+        if self.has_plugin(PluginRequirement("core", max="3.20.0")):
```
gerrod3, line 154: Wonder if there is a way to put this in a separate file(s). It's possible we could end up with a bunch of small little spec changes we need to prevent bugs, and that would make this function become quite bloated.

gerrod3 (changes requested):
Yay for validation!

pulpcore/cli/common/openapi.py
mdellweg: That would be a circular import.

And importing the other way around lead to "this was not explicitly exported" errors.

pulpcore/cli/common/openapi.py
mdellweg: At this point, we _want_ the next layer to provide a "hydrated" type (in this case an int).
Min-max, sure possible. We also have other types and formats left for further validation. Can we collect what's missing in an issue?

pulpcore/cli/common/openapi.py
mdellweg: copy-past leftover.

pulpcore/cli/common/openapi.py
mdellweg: base64 could be checked for sure. On the other hand pulp does not use all the formats, so we could add a lot of untested code here.

pulpcore/cli/common/openapi.py
mdellweg: Even more so it processes the objects in the dict in place. Yes.
I think in the long run we may change the outer signature to accept all parts as one body, and separate the files after validation for the request call.

mdellweg:
@gerrod3 @ggainey I think i addressed all comments. Please have a second look.

pulpcore/cli/common/openapi.py
ggainey: OK, makes sense - can we get a comment that says "duplicated from common/context because of circular imports" so that Future Us will not forget and end up changing this in only one place.

mdellweg: Maybe importing from two levels down is not _that_ bad. I changed the definitions a bit...

pulpcore/cli/common/context.py
```diff
@@ -108,6 +106,54 @@ def __init__(
...
+                        ):
+                            parameter["schema"] = {"type": "array", "items": {"type": "string"}}
+        if self.has_plugin(PluginRequirement("file", max="1.11.0")):
+            operation = api_spec["paths"]["{file_file_alternate_content_source_href}refresh/"][
+                "post"
+            ]
+            operation.pop("requestBody")
+        if self.has_plugin(
+            PluginRequirement("python", max="99.99.0.dev")
+        ):  # TODO Add version bounds
```
gerrod3, line 146: We should fix this in pulp_python. I see the issue you have filed for it.

pulpcore/cli/common/context.py
```diff
@@ -689,7 +744,7 @@ def scope(self) -> Dict[str, Any]:
         return {self.repository_ctx.HREF: self.repository_ctx.pulp_href}
 
     def repair(self, href: Optional[str]) -> Any:
-        return self.call("repair", parameters={self.HREF: href or self.pulp_href})
+        return self.call("repair", parameters={self.HREF: href or self.pulp_href}, body={})
```
gerrod3, line 747: Why does this require an empty dict for `body`? Can it not be left out?

pulpcore/cli/common/openapi.py
```diff
@@ -164,54 +176,160 @@ def extract_params(
...
+        minimum = schema.get("minimum")
+        if minimum is not None and value < minimum:
+            raise OpenAPIValidationError(
+                _("'{name}' is violating the minimum constraint").format(name=name)
+            )
+        maximum = schema.get("maximum")
+        if maximum is not None and value > maximum:
+            raise OpenAPIValidationError(
+                _("'{name}' is violating the maximum constraint").format(name=name)
+            )
```
gerrod3, line 322: Nice! :100:

pulpcore/cli/common/openapi.py
```diff
@@ -164,54 +176,160 @@ def extract_params(
...
+        return value
+
+    def validate_number(self, schema: Any, name: str, value: Any) -> float:
+        # https://swagger.io/specification/#data-types describes float and double.
+        # Python does not distinguish them.
+        if not isinstance(value, float):
+            raise OpenAPIValidationError(
+                _("'{name}' is expected to be an integer.").format(name=name)
```
gerrod3, line 330:
```suggestion
                _("'{name}' is expected to be an float.").format(name=name)
```

gerrod3 (changes requested):

pulpcore/cli/common/context.py
mdellweg: It requires a body with only an optional property.

`pulp debug openapi operation --id repositories_rpm_rpm_versions_repair | jq '.operation.requestBody'`
```json
{
  "content": {
    "application/json": {
      "schema": {
        "$ref": "#/components/schemas/Repair"
      }
    },
    "application/x-www-form-urlencoded": {
      "schema": {
        "$ref": "#/components/schemas/Repair"
      }
    },
    "multipart/form-data": {
      "schema": {
        "$ref": "#/components/schemas/Repair"
      }
    }
  },
  "required": true
}
```
`pulp debug openapi schema --name Repair`
```
{
  "type": "object",
  "properties": {
    "verify_checksums": {
      "type": "boolean",
      "default": true,
      "description": "Will verify that the checksum of all stored files matches what saved in the database. Otherwise only the existence of the files will be checked. Enabled by default"
    }
  }
}
```

But I could add a check for the `required` attribute in the `responseBody`.

pulpcore/cli/common/openapi.py
mdellweg: Actually in the OpenAPI world they call it "number".

pulpcore/cli/common/context.py
mdellweg: Accounted for the `responseBody.required` now. This lead to a new subroutine `render_request_body`.

gerrod3 (approved):

ggainey (approved):

PR #608, opened 2022-11-30
Make the --requirements param to accept file and text

### LINKED ISSUE #230: Change the `--requirements` option on ansible remote commands to accept a string or file
We should probably follow the pattern we use elsewhere and have `--requirements` accept a string or a file with `@` instead of having `--requirements` and `--requirements-file`.

lubosmj:
closes #230


pulpcore/cli/common/generic.py
```diff
@@ -328,74 +328,59 @@ def _version_callback(
     return value
 
 
-# TODO: would be great to have enable this to take a validator, rather than having
-# to build "on top of" it like I'm doing now w/ json_callback
 def load_file_or_string_callback(
```
mdellweg, line 331: I'm not grocking the whole change yet, but this thing now looks deceivingly like a decorator. Do you think we can use it as one?

pulpcore/cli/ansible/remote.py
```diff
@@ -69,7 +68,7 @@ def remote(ctx: click.Context, pulp_ctx: PulpCLIContext, remote_type: str) -> No
 collection_options = [
     pulp_option(
         "--requirements-file",
-        callback=_requirements_callback,
+        callback=lambda c, p, x: f"{yaml.safe_load(x)}" if x else x,
```
mdellweg, line 71: Guess this is not what we wanted...

pulpcore/cli/ansible/repository.py
```diff
@@ -146,11 +146,11 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
     ),
     href_option,
 ]
-content_json_callback = create_content_json_callback(schema=CONTENT_LIST_SCHEMA)
+
```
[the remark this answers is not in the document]
lubosmj: I do not understand. I am using literally the same callback as before a couple of lines below.

pulpcore/cli/common/generic.py
mdellweg: So this one can be used as a callback or chained when used as a decorator...
Maybe that should be in the docstring.

pulpcore/cli/common/generic.py
```diff
@@ -328,59 +328,52 @@ def _version_callback(
...
 
-    return the_content
+    return _load_file_or_string_callback
 
 
-def load_json_callback(ctx: click.Context, param: click.Parameter, value: Optional[str]) -> Any:
-    """Load JSON from input string or from file if string starts with @."""
+@load_file_or_string_callback
+def string_callback(ctx: click.Context, param: click.Parameter, value: str) -> str:
+    return value
```
mdellweg, line 361: This is just like using the load_file_or_string_callback directly.

pulpcore/cli/common/generic.py
```diff
@@ -328,59 +328,52 @@ def _version_callback(
...
-    ctx: click.Context, param: click.Parameter, value: Optional[str]
-) -> Optional[Dict[str, str]]:
-    # None is legal - shortcircuit here
-    if value is None:
-        return value
-
-    value = load_json_callback(ctx, param, value)
+@load_file_or_string_callback
+def labels_callback(ctx: click.Context, param: click.Parameter, value: str) -> Any:
+    value = json_callback(ctx, param, value)
```
mdellweg, line 376: Maybe `json_callback` should follow the chaining pattern. Also this version of `json_callback` is already decorated. So we are attemting to lookup a file twice here.

pulpcore/cli/common/generic.py
```diff
@@ -724,7 +718,7 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
         "Search for {entities} with these content hrefs in them (JSON list or "
         "@file containing a JSON list)"
     ),
-    callback=load_json_callback,
+    callback=json_callback,
```
mdellweg, line 721: What do you think about the following naming convention?
`load_string_callback` searches for "@" and attempts to load from file else value.
`json_callback` transforms `value` to json.
`load_json_callback` is `json_callback` decorated (chained) with `load_string_callback`.

mdellweg: `string_callback` would be the same as specifying nothing...

pulpcore/cli/common/generic.py
mdellweg: I think, i'd be ok to use the load_json_callback here. We never want a version of the label callback that cannot read from file, right?

pulpcore/cli/common/generic.py
```diff
@@ -791,7 +789,7 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
         "Search for {entities} with these content hrefs in them (JSON list or "
         "@file containing a JSON list)"
     ),
-    callback=load_json_callback,
+    callback=load_file_or_string_callback,
```
mdellweg, line 792: The help text states we _want_ JSON...

mdellweg (changes requested):

pulpcore/cli/ansible/remote.py
```diff
@@ -71,26 +72,26 @@ def remote(ctx: click.Context, pulp_ctx: PulpCLIContext, remote_type: str) -> No
 collection_options = [
     pulp_option(
         "--requirements-file",
-        callback=_requirements_callback,
+        callback=requirements_file_callback,
         type=click.File(),
         help=_("Collections only: a Collection requirements yaml"),
         allowed_with_contexts=collection_context,
     ),
```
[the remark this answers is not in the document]
lubosmj: Should I define any deprecation warning or remove the option completely?

pulpcore/cli/common/generic.py
lubosmj: Makes perfect sense. Thanks!

pulpcore/cli/ansible/remote.py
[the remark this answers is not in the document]
mdellweg: Hmm, can we deprecate single parameters?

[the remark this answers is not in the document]
lubosmj: Seems like there is no simple way to deprecate the option. I am in favour of removing it.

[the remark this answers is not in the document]
mdellweg: It's probably a matter of adding a note to the help text then. Removing may also be ok.
WDYT @gerrod3 ?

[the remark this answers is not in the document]
gerrod3: Add a note in the help option and remove it in a future release.

pulpcore/cli/common/generic.py
```diff
@@ -331,74 +331,60 @@ def _version_callback(
     return value
 
 
-# TODO: would be great to have enable this to take a validator, rather than having
-# to build "on top of" it like I'm doing now w/ json_callback
-def load_file_or_string_callback(
-    ctx: click.Context, param: click.Parameter, value: Optional[str]
+def load_file_wrapper(
+    handler: Callable[[click.Context, click.Parameter, str], Any] = lambda c, p, x: x
```
mdellweg, line 335: I think, i'd be better to not have the default value here, but pass the lambda in line 364.
`load_string_callback = load_file_wrapper(lambda c, p, x: x)`

pulpcore/cli/common/generic.py
```diff
@@ -842,7 +828,7 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: Optional[s
     help=_(
         "JSON dictionary of labels to set on {entity} (or " "@file containing a JSON dictionary)"
     ),
-    callback=labels_callback,
+    callback=load_json_callback,
```
mdellweg, line 831: The existing `labels_callback` provides valuable validation. Please keep it.

pulpcore/cli/file/repository.py
```diff
@@ -69,10 +70,9 @@ def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) ->
 CONTENT_LIST_SCHEMA = s.Schema([{"sha256": str, "relative_path": s.And(str, len)}])
 
 
-def _content_list_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Any:
-    result = load_json_callback(ctx, param, value)
-    if result is None:
-        return None
+@load_file_wrapper
+def _content_list_callback(ctx: click.Context, param: click.Parameter, value: str) -> Any:
+    result = json_callback(ctx, param, value)
```
mdellweg, line 75: Don't we still need the
```python
    if result is None:
        return None
```
part?

lubosmj: Yes, we do! But, this is the right thing:

```python
    if not value:
        return value
```

pulpcore/cli/common/generic.py
lubosmj: I thought you wanted to remove it by saying this: https://github.com/pulp/pulp-cli/pull/608#discussion_r1049673939. My bad. I did this now:
```

def load_labels_callback(ctx: click.Context, param: click.Parameter, value: str) -> Any:
    value = load_json_callback(ctx, param, value)

```

mdellweg: So labels are optionally read from file and always validated? Sounds right to me.

pulpcore/cli/common/generic.py
```diff
@@ -331,55 +331,53 @@ def _version_callback(
...
 
 
-def labels_callback(
-    ctx: click.Context, param: click.Parameter, value: Optional[str]
-) -> Optional[Dict[str, str]]:
-    # None is legal - shortcircuit here
+load_json_callback = load_file_wrapper(json_callback)
+
+
+def load_labels_callback(ctx: click.Context, param: click.Parameter, value: str) -> Any:
```
mdellweg, line 380: We can be more specific here.
```suggestion
def load_labels_callback(ctx: click.Context, param: click.Parameter, value: str) -> Optional[Dict[str, str]]:
```
(Don't forget to run black...)
Also when you change the name, can you verify that this is not used anywhere in the known plugins?

This should be my last request. We are ready to go then...

lubosmj: I run `git grep` against pulp-cli, pulp-cli-deb, and pulp-cli-ostree. Only these occurrences were found:
```
pulpcore/cli/common/generic.py:def load_labels_callback(
pulpcore/cli/common/generic.py:    callback=load_labels_callback,
```

I think we are good to go.

lubosmj:
I have made a couple of `value` parameters to receive `Optional[str]` values. I hope it is alright.

PR #617, opened 2023-01-05
Raise the IncompatibleContext exception in case of bad context

### LINKED ISSUE #580: test_help_pages doesn't properly test polymorphic commands
## Summary
`test_help_pages` iterates through all command groups and tests each commands help page, but it does not take into account that some commands are only callable from certain contexts when using the type option.

## Additonal context

Ideas:

Add an excludes list for certain commands to prevent them from being called without the proper context.

Add comments in the commands doc's string that the test checks to understand how or if the help page should be checked.

lubosmj:
closes #580

### DIFF
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -42,6 +42,10 @@ translation = get_translation(__name__)
 _ = translation.gettext
 
 
+class IncompatibleContext(click.UsageError):
+    pass
+
+
 class ClickNoWait(click.ClickException):
     exit_code = 0
 
@@ -195,8 +199,24 @@ class PulpGroup(PulpCommand, click.Group):
         if isinstance(cmd, (PulpCommand, PulpGroup)):
             if cmd.allowed_with_contexts is not None:
                 if not isinstance(ctx.obj, cmd.allowed_with_contexts):
-                    return None
+                    raise IncompatibleContext(
+                        _("The subcommand '{name}' is not available in this context.").format(
+                            name=cmd.name
+                        )
+                    )
         return cmd
+
+    def list_commands(self, ctx: click.Context) -> List[str]:
+        commands_filtered_by_context = []
+
+        for name, cmd in self.commands.items():
+            if isinstance(cmd, (PulpCommand, PulpGroup)):
+                if cmd.allowed_with_contexts is None or isinstance(
+                    ctx.obj, cmd.allowed_with_contexts
+                ):
+                    commands_filtered_by_context.append(name)
+
+        return sorted(commands_filtered_by_context)
 
 
 def pulp_command(name: Optional[str] = None, **kwargs: Any) -> Callable[[F], click.Command]:
--- a/tests/test_help_pages.py
+++ b/tests/test_help_pages.py
@@ -11,9 +11,21 @@ main = load_plugins()
 
 def traverse_commands(command, args):
     yield args
+
     if isinstance(command, click.Group):
         for name, sub in command.commands.items():
             yield from traverse_commands(sub, args + [name])
+
+        params = command.params
+        if params:
+            if "--type" in params[0].opts:
+
+                # iterate over commands with specific context types
+                for context_type in params[0].type.choices:
+                    yield args + ["--type", context_type]
+
+                    for name, sub in command.commands.items():
+                        yield from traverse_commands(sub, args + ["--type", context_type, name])
 
 
 @pytest.fixture
@@ -31,5 +43,9 @@ def test_access_help(no_api, args):
     """Test, that all help screens are accessible without touching the api property."""
     runner = CliRunner()
     result = runner.invoke(main, args + ["--help"], catch_exceptions=False)
-    assert result.exit_code == 0
-    assert result.stdout.startswith("Usage:") or result.stdout.startswith("DeprecationWarning:")
+
+    if result.exit_code == 2:
+        assert "not available in this context" in result.stdout
+    else:
+        assert result.exit_code == 0
+        assert result.stdout.startswith("Usage:") or result.stdout.startswith("DeprecationWarning:")
--- a/CHANGES/566.feature
+++ b/CHANGES/566.feature
@@ -1,0 +1,1 @@
+Added remove-image command to pulp_container.
--- a/CHANGES/580.misc
+++ b/CHANGES/580.misc
@@ -1,0 +1,1 @@
+Updated the error message shown to users when using commands in the wrong context.
--- a/pulpcore/cli/container/context.py
+++ b/pulpcore/cli/container/context.py
@@ -168,7 +168,13 @@ class PulpContainerPushRepositoryContext(PulpContainerBaseRepositoryContext):
     CAPABILITIES = {
         "tag": [PluginRequirement("container", "2.3.0")],
         "roles": [PluginRequirement("container", "2.11.0")],
+        "remove": [PluginRequirement("container", "2.4.0")],
     }
+
+    def remove_image(self, digest: str) -> Any:
+        self.needs_capability("remove")
+        body = {"digest": digest}
+        return self.call("remove_image", parameters={self.HREF: self.pulp_href}, body=body)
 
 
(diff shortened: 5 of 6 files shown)


tests/test_help_pages.py
```diff
@@ -26,10 +27,30 @@ def getter(self):
...
     """Test, that all help screens are accessible without touching the api property."""
     runner = CliRunner()
+
+    command, args = command_args
     result = runner.invoke(main, args + ["--help"], catch_exceptions=False)
+
+    if result.exit_code != 0:
+        # verify whether a non-default repository context is necessary for executing the command
+        for context_name, registered_context_class in registered_repository_contexts.items():
+            for context_class in command.allowed_with_contexts:
```
gerrod3, line 41: This works well for repository commands, but doesn't help for other multi-context commands like content. Maybe we need a giant table to register all the different types possible for each command.

lubosmj: Yes, as you are saying. I focused only on repository commands. I did not want to create the table. However, it seems like it would be the best option.

mdellweg: Can we get away with monkeypatching (maybe the `allowed_with_context` machinery) somehow?
All we need is the to see the help pages. I don't think, we need a fully working context. And i'd prefer not to add functionality to the actual code for the purpose of this test.

lubosmj: > I don not quite remember the details of the issue we are tying to solve. But it is that some subcommands do not work with certain contexts, right?

Yes. In particular, the `pulp container repository remove-image` command is implemented only for push repositories (i.e., we have to always use the option `-t push` to change the default context and make the help pages available).

> If while enumerating subcommands, we clear their allowed_with_contexts attribute (with monkeypatch) they should again be available to the test, right?

Well, yes and no. We want to tell the test that this specific command can be run only within this specific context. In the PR, we are iterating through allowed context types until we reach the point where the help page is eventually available under one of the contexts.

> The question to follow up on: Is this then still testing the thing we want to test? Namely that all help pages are accessible without calling into the API.

I would say *yes*. Polymorphic commands could have different types of options per context. And, we want to show that to a user. When reading `allowed_with_context` classes, do we initiate any API calls?

mdellweg: From reading that attribute, no we do not call the api. I guess it's hard to impossible to derive the proper value for `--type` from the context class in `allowed_with_context`, no?
Also by that logic, we should exercise all combinations of `--type` options.

How about the following: We raise a more specific exception (a new subclass of `ClickException` maybe called `IncompatibleContext`) on this condition (a small to unnoticeable codechange), iterate over all `--type` choices and ignore that new exception.

lubosmj: I cannot think of any easy way of getting the all `--type` choices apart from doing introspection. Is there any Click facility that will allow me to do that? I suppose we have the same thing in our minds (i.e., the `--type` choices available under a corresponding decorator):
```
@pulp_group()
@type_option(
    choices={
        "container": PulpContainerRepositoryContext,
        "push": PulpContainerPushRepositoryContext,
    },
    default="container",
)
def repository() -> None:
    pass
```

pulpcore/cli/common/generic.py
```diff
@@ -42,6 +42,21 @@
...
+    def show(self, file: Optional[IO[str]] = None) -> None:
+        """
+        Format the message into file or STDERR.
+        Overwritten from base class to print "IncompatibleContext: ".
+        """
+        if file is None:
+            file = click.get_text_stream("stderr")
+        click.echo(
+            _("IncompatibleContext: {message}").format(message=self.format_message()), file=file
+        )
```
mdellweg, line 57: I think less is more:
```suggestion
    pass
```

First this exception is signifying a user error so "exit 1" is appropriate. Second let's not copy code from click if we do not really need to.

pulpcore/cli/common/generic.py
```diff
@@ -195,9 +210,46 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
         if isinstance(cmd, (PulpCommand, PulpGroup)):
             if cmd.allowed_with_contexts is not None:
                 if not isinstance(ctx.obj, cmd.allowed_with_contexts):
-                    return None
+                    raise IncompatibleContext(
+                        "The command is not available under the current context."
+                    )
         return cmd
 
+    def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
```
mdellweg, line 218: Now this is really unfortunate. And right now it's hard to see which part was copied from click and what you changed to make it work. I would really want to avoid doing this. The more we monkeypatch click, the more likely it will break us in the future. We may need to think about another angle here.
What if instead we overwrite `list_commands` too to be ctx sensitive?

tests/test_help_pages.py
```diff
@@ -11,10 +11,26 @@
...
     yield args
+
     if isinstance(command, click.Group):
         for name, sub in command.commands.items():
             yield from traverse_commands(sub, args + [name])
 
+        params = command.params
+        if params:
+            try:
+                context_choices = params[0].type.choices
```
mdellweg, line 22: This is probably not the best indicator. We should "find" the param with the name "--type" in the list.

tests/test_help_pages.py
```diff
@@ -11,10 +11,26 @@
...
+            except AttributeError:
+                return
+
+            # iterate over commands with specific context types
+            for context_type in context_choices:
+                yield args + ["--type", context_type]
+
+                if isinstance(command, click.Group):
```
mdellweg, line 30: We know that already.

mdellweg (changes requested):

pulpcore/cli/common/generic.py
lubosmj: I decided to make this change to better inform the user about the possibly wrong use of context. I can update `exit_code`. But, I am in favour of leaving the error message formatting untouched.

pulpcore/cli/common/generic.py
lubosmj: I can take a look at `list_commands`.

lubosmj:
@mdellweg, I changed the exit code to 1 and overridden the `list_commands` method.

pulpcore/cli/common/generic.py
```diff
@@ -195,9 +210,24 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
...
+        for subcommand in super().list_commands(ctx):
+            try:
+                cmd = self.get_command(ctx, subcommand)
+            except IncompatibleContext:
+                continue
+            else:
+                if cmd is not None:
+                    commands_filtered_by_context.append(subcommand)
+
+        return commands_filtered_by_context
```
mdellweg, line 229: The original looks like this: https://github.com/pallets/click/blob/cba52fa76135af2edf46c154203b47106f898eb3/src/click/core.py#L1938
What if we say
```python
return sorted([name for name, cmd in self.commands.items() if not hasattr(cmd, "allowed_with_contexts") or ctx.obj in cmd.allowed_with_contexts])
```

I think it is less back and forth between the command object and it's name.

lubosmj: In the end, it looks similar to the `get_command` method. Do you find it better?

tests/test_help_pages.py
```diff
@@ -31,5 +43,9 @@ def test_access_help(no_api, args):
     """Test, that all help screens are accessible without touching the api property."""
     runner = CliRunner()
     result = runner.invoke(main, args + ["--help"], catch_exceptions=False)
-    assert result.exit_code == 0
-    assert result.stdout.startswith("Usage:") or result.stdout.startswith("DeprecationWarning:")
+
+    if result.exit_code == 2:
+        assert "not available in this context" in result.stdout
```
mdellweg, line 48: Ha, usage errors have exit 2. That's nice.

pulpcore/cli/common/generic.py
mdellweg: Well they are similar in nature. But the fact that `list_commands` returns the names only makes it a bit harder.

mdellweg (approved):

PR #623, opened 2023-01-30
Update Supported Workflows

gerrod3:
Review Checklist:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,9 +4,9 @@ This is the documentation for Pulp 3's command line interface.
 
 Here you can find information about how to install the Pulp 3 CLI, and general usage and syntax information.
 
-Currently `pulp-cli` supports 5 versions of `pulpcore` from 3.11-3.16 and 5 of Pulp's plugins:
-`pulp_ansible`, `pulp_container`, `pulp_file`, `pulp_python` and `pulp_rpm`. See section below for information
-on how to extend the CLI to support other Pulp plugins.
+Currently `pulp-cli` supports at a minimum the 5 latest versions of pulpcore and the corresponding compatible 
+versions of 5 Pulp's plugins: `pulp_ansible`, `pulp_container`, `pulp_file`, `pulp_python` and `pulp_rpm`.
+See section below for information on how to extend the CLI to support other Pulp plugins.
 
 Check out the [supported workflows section](supported_workflows) to learn about the capabilities of the CLI. 
 You can find more workflow examples for the Pulp 3 CLI throughout the [plugin documentation](https://docs.pulpproject.org/pulpcore/plugins/index.html).
--- a/docs/supported_workflows.md
+++ b/docs/supported_workflows.md
@@ -1,9 +1,15 @@
 # Supported Workflows
 
 `pulp-cli` is still in beta, so the features and workflows listed here are subject to change.
-Currently `pulp-cli` supports 5 versions of `pulpcore` from 3.11-3.16 and 5 of Pulp's plugins:
-`pulp_ansible`, `pulp_container`, `pulp_file`, `pulp_python` and `pulp_rpm`. Some of Pulp's other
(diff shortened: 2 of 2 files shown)


docs/index.md
```diff
@@ -4,7 +4,7 @@ This is the documentation for Pulp 3's command line interface.
 
 Here you can find information about how to install the Pulp 3 CLI, and general usage and syntax information.
 
-Currently `pulp-cli` supports 5 versions of `pulpcore` from 3.11-3.16 and 5 of Pulp's plugins:
+Currently `pulp-cli` supports at a minimum the 5 latest versions of pulpcore and 5 of Pulp's plugins:
```
gerrod3, line 7: Not sure what the best way to word this is (or if this is even still our official stance).

docs/supported_workflows.md
```diff
@@ -152,9 +153,10 @@ The CLI currently supports the following operations on these `pulp_rpm` objects
...
 
 * RPM Alternative Content Sources - **CRUD, Add, Remove, Refresh**
-* RPM Packages Content - **CR**
+* RPM Package/Advisory/ModuleMD Defaults/ModuleMD Content - **CR**
+* RPM Distribution Tree/Repo Metadata Content - **R**
 * RPM Distributions - **CRUD**
 * RPM Publications - **CRD**
-* RPM Remote - **CRUD**
+* RPM/ULN Remote - **CRUD**
 * RPM Repository - **CRUD, Modify, Sync**
```
mdellweg, line 161: So we are missing RBAC on RPM... I see a prio-list here.

docs/index.md
mdellweg: Wow, it's even more complicated. Let's collect information:
- We test with 5 consecutive versions of pulpcore.
- We test the plugins as they are bundled with these 5 pulpcore versions in the OCI containers. This can divert from 5 versions and may not be continuous.
- We decided not to remove all the workarounds we collected for older versions right away.
- By "support", we do not mean all available features.

So is this like two levels of support?
Is it time to archive the migration plugin parts?

docs/supported_workflows.md
ggainey: To be fair, RBAC only got released for RPM in rpm/3.19 on 2023-02-06

docs/index.md
ggainey: (2to3 is supported through core/3.19. As long as there are major Pulp users on <core/3.20, we should prob leave the 2to3 support as-is.)

Regarding "support" - we have "support" meaning "tested against", "enables features of", and "what we'll accept bug-reports for". That's a lot for one word to carry. In that context, maybe we can say something like this:

"""
pulp-cli is tested against the five most-recent pulpcore releases as of the date of the pulp-cli release. It enables support for (currently) five Pulp plugins: `pulp_ansible`, `pulp_container`, `pulp_file`, `pulp_python` and `pulp_rpm`.

We test against each of those pulpcore releases, and the plugin-releases associated with them in Pulp's oci-images (q.v.)

While pulp-cli currently continues to work against older versions of pulpcore, we're unlikely to take bug-reports for support of such older versions.

(NOTE: pulp-cli does not (yet) expose all functionality provided by the REST API of pulpcore and enabled plugins. RFEs and pull-requests for missing features gratefully and cheerfully accepted!)
"""

wdyt?

gerrod3: Took some of your suggestions and added it in. Take a look.

ggainey (approved):
Cool -  I'm good w/ the new wording, thanks!

docs/index.md
mdellweg: The thing is the 5 supported plugin versions is plainly wrong. If the plugin releases not often we may only support 2 versions that come in the timeline of 5 pulpcore versions.

gerrod3: Woops, forgot to update this file as well. Take a look now.

mdellweg (approved):
Thank you. That is much clearer and more accurate.

PR #629, opened 2023-02-10
Split out glue library as pulp-glue package

### LINKED ISSUE #628: [TASK] Split glue layer as a library into separate python package
https://discourse.pulpproject.org/t/rfc-reunite-the-library-that-drives-pulp-cli-and-squeezer/262

mdellweg:
fixes #628

TODO Checklist:
- [x] Move the contexts of all plugins to glue.
- [x] Figure out the release process for the split packages.

Review Checklist:


pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -74,7 +74,7 @@ def __init__(
             raise OpenAPIError(_("Cert is required if key is set."))
         self._session.headers.update(
             {
-                "User-Agent": user_agent or "Pulp-CLI openapi parser",
+                "User-Agent": user_agent or f"Pulp-glue openapi parser ({__version__})",
```
ggainey, line 77: :+1:

pulpcore/cli/core/user.py
```diff
@@ -1,8 +1,8 @@
 import click
+from pulp_glue.common.context import PluginRequirement  # PulpEntityContext,; pass_entity_context,
```
ggainey, line 2: Is there a reason to leave the commented-out imports here?

ggainey (commented):
This looks good! I have one question below. Also - should we give the Build Gang some advance warning on this? (assuming you haven't already)

pulpcore/cli/core/user.py
mdellweg: Not at all. TBH they were there before and i used sed to update _all_ the files.

setup.py
```diff
@@ -39,12 +39,11 @@
     package_data={"": ["py.typed", "locale/*/LC_MESSAGES/*.mo"]},
     python_requires=">=3.6",
     install_requires=[
+        "pulp-glue>=0.17.0.dev,<0.18",
         "click>=8.0.0,<9.0.0",
-        "packaging",
```
ggainey, line 43: So contexts were/are the only things that rely on 'packaging' and 'requests', and that's now all in pulp_glue, correct?

setup.py
```diff
@@ -39,12 +39,11 @@
     package_data={"": ["py.typed", "locale/*/LC_MESSAGES/*.mo"]},
     python_requires=">=3.6",
     install_requires=[
+        "pulp-glue>=0.17.0.dev,<0.18",
```
ggainey, line 42: The concern I have here is that we'll need a pulp-cli release when we release a new pulp-glue. That may be OK - I'm just trying to envision the release-process for the tooling as we go forward. Thoughts?

ggainey (commented):
I left a couple of comments, but I think this all makes sense - I may need to see it "in action" to really have a feeling for how releasing is going to look.

setup.py
mdellweg: And bump2version will take care of it, i hope...

pulpcore/cli/common/context.py
```diff
@@ -1,833 +1,23 @@
...
-    "%Y-%m-%dT%H:%M:%S",  # ... with time and T as a separator
-    "%y/%m/%d",  # US format
-    "%y/%m/%d %h:%M:%S %p",  # ... with time
-    "%y/%m/%dT%h:%M:%S %p",  # ... with time and T as a separator
-    "%y/%m/%d %H:%M:%S",  # ... with time 24h
-    "%y/%m/%dT%H:%M:%S",  # ... with time 24h and T as a separator
-    "%x",  # local format
-    "%x %X",  # ... with time
-    "%xT%X",  # ... with time and T as a separator
+from pulp_glue.common.context import (  # noqa: F401
```
ggainey, line 1: This is so that any cli-plugins we don't know about, will continue working post-this-change, correct?

ggainey (commented):
Left one question, but I think we're about ready to merge.

pulpcore/cli/common/context.py
mdellweg: Absolutey!

ggainey (approved):

PR #633, opened 2023-02-16
Use subtests to group the help_page tests

mdellweg:
[noissue]

Review Checklist:


ggainey (approved):
Cool, I had not known about pytest-subtest before

PR #661, opened 2023-04-02
Refactor task cancel mechanics

mdellweg:
This is not a user visible change.

[noissue]

Review Checklist:

### DIFF
--- a/pulp-glue/pulp_glue/common/context.py
+++ b/pulp-glue/pulp_glue/common/context.py
@@ -280,10 +280,14 @@ class PulpContext:
         return result
 
     @classmethod
-    def _check_task_finished(cls, task: EntityDefinition) -> bool:
+    def _check_task_finished(cls, task: EntityDefinition, expect_cancel: bool = False) -> bool:
         task_href = task["pulp_href"]
 
         if task["state"] == "completed":
+            if expect_cancel:
+                raise PulpException(
+                    _("The task {task_href} meant to be canceled, completed instead.")
+                )
             return True
         elif task["state"] == "failed":
             raise PulpException(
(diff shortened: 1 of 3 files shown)


lubosmj (approved):

pulp-glue/pulp_glue/common/context.py
```diff
@@ -280,10 +280,12 @@ def call(
         return result
 
     @classmethod
-    def _check_task_finished(cls, task: EntityDefinition) -> bool:
+    def _check_task_finished(cls, task: EntityDefinition, expect_cancel: bool = False) -> bool:
         task_href = task["pulp_href"]
 
         if task["state"] == "completed":
+            if expect_cancel:
+                raise PulpException(_("The task {task_href} meant to be canceled, completed instead."))
```
ipanova, line 288: How about : The task completed before it got canceled?

pulp-glue/pulp_glue/common/context.py
```diff
@@ -280,10 +280,12 @@ def call(
...
             return True
         elif task["state"] == "failed":
```
ipanova, line 290: not sure, but do you want to check here for the `expect_cancel` too?

mdellweg: I think no. If i cancel and it fails instead, i want to know that.

pulp-glue/pulp_glue/common/context.py
mdellweg: Discussed it with Lubos in the channel. This is what we agreed upon. Also i don't even see how this could happen on a recent pulpcore...

ipanova (approved):

PR #666, opened 2023-04-12
Fixed generic-publications (it's not 'pulications' any more).

### LINKED ISSUE #665: "publication list"" is broken
## Summary

Running `pulp publication list` fails with `KeyError: 'pulications_list'`

"pulications" is missing a "b" in `pulp-glue/pulp_glue/common/context.py` - make sure to include a test for this call in the fix!

ggainey:
Also fixed a test that would have caught this if it weren't failing silently.

fixes #665


pulp-glue/pulp_glue/common/context.py
```diff
@@ -780,8 +780,8 @@ class PulpRemoteContext(PulpEntityContext):
 
 class PulpPublicationContext(PulpEntityContext):
     ENTITY = _("publication")
-    ENTITIES = _("pulications")
-    ID_PREFIX = "pulications"
+    ENTITIES = _("publications")
+    ID_PREFIX = "publications"
```
ggainey, line 784: This is literally the only fix I wanted to make in this PR...

pulpcore/cli/container/distribution.py
```diff
@@ -132,7 +132,8 @@ def update(
             distribution_ctx.update(body={"repository": ""}, non_blocking=True)
             body["repository_version"] = f'{distribution["repository"]}versions/{version}/'
         elif distribution["repository_version"]:
-            repository_href, _, _ = distribution["repository_version"].partition("versions")
+            # unused x1 and x2 are to get us around a mypy/1.2 complaint about '_'
+            repository_href, x1, x2 = distribution["repository_version"].partition("versions")
```
mdellweg, line 136:
```suggestion
            repository_href, dummy1, dummy2 = distribution["repository_version"].partition("versions")
```

"_" is the python suggested way to do things like this...
Anyway, i always suspected this to collide with `_ = gettext.gettext`...
Your comment kind of drops out of its context when not viewd in the diff, but the resulting code....

ggainey: > "_" is the python suggested way to do things like this...

Oh, it's definitely "pythonic" - but it breaks ion mypy1.2. 

I think the comment is appropriate even w/out the context - "we don't use the 2nd and 3rd values from the tuple, but we can't use '_' because mypy 1.2 complains" feels like a long way to say the same thing. What did you have in mind?

mdellweg: Just calling them "dummy", to make it obvious they are not used.
Next time someone tries to use `_` again, mypy will do the yelling for us.

tests/scripts/pulp_file/test_publication.sh
```diff
@@ -16,11 +16,12 @@ pulp file repository create --name "cli_test_file_repository" --remote "cli_test
 pulp file repository sync --name "cli_test_file_repository"
 
 expect_succ pulp file publication create --repository "cli_test_file_repository"
+
 PUBLICATION_HREF="$(echo "$OUTPUT" | jq -r .pulp_href)"
 expect_succ pulp file publication destroy --href "$PUBLICATION_HREF"
 expect_succ pulp file publication create --repository "cli_test_file_repository" --version 0
 PUBLICATION_HREF="$(echo "$OUTPUT" | jq -r .pulp_href)"
-if pulp debug has-plugin --name "core" --min "3.20.0"
+if pulp debug has-plugin --name "core" --min-version "3.20.0"
```
mdellweg, line 24: subtle...

pulpcore/cli/container/distribution.py
```diff
@@ -132,7 +132,8 @@ def update(
             distribution_ctx.update(body={"repository": ""}, non_blocking=True)
             body["repository_version"] = f'{distribution["repository"]}versions/{version}/'
         elif distribution["repository_version"]:
-            repository_href, _, _ = distribution["repository_version"].partition("versions")
+            # 'dummy' vars are to get us around a mypy/1.2 complaint about '_'
+            repository_href, dummy, dummy = distribution["repository_version"].partition("versions")
```
mdellweg, line 136: :+1:

mdellweg (approved):

PR #688, opened 2023-05-05
Add TYPE_REGISTRY to master-detail contexts

mdellweg:
[noissue]

Review Checklist:


pulp-glue/pulp_glue/ansible/context.py
```diff
@@ -17,6 +16,8 @@
 
 
 class PulpAnsibleCollectionVersionContext(PulpContentContext):
+    PLUGIN = "ansible"
+    RESOURCE_TYPE = "collection_version"
```
mdellweg, line 20: In the end this is a user facing string (still an identifier, so no, not translatable). Do we want to use `-` instead of `_` here?

ggainey: I don't have a strong opinion either way (amazing, I know...) Obviously, if we decide '-', then we do it for all the RESOURCE_TYPEs - but I am OK with this kind-of identifier looking more like an identifier, than a human concept.

ggainey (approved):
This looks good to me. It's a little more magic, but it's consistent magic, which is the best kind. Assuming we're good with the _/- discussion, I am good w/ approving!

mdellweg:
I'm going to merge this. If we decide to change the "- _" thing, we will only annoy all users and break existing scripts. ;)

PR #691, opened 2023-05-11
Update test matrix

mdellweg:
Start testing against pulpcore 3.27.

[noissue]

Review Checklist:


ggainey (commented):
What are the unstarted/waiting checks? They look like the "old" ones?

mdellweg:
Seems like i need to make a very bad hack to depend on all test targets.

PR #693, opened 2023-05-18
Fixed tests that break in domain-enabled environments.

### LINKED ISSUE #692: RPM tests failing in the presence of domains
Domains change URLs to have the domain-name embedded. test_acs and test_rpm_sync_publish.sh make asssumptions that are invalid. There are better ways to extract the information, that work with and without domains.

ggainey:
fixed #692.


ggainey:
This need to be released before the rpm-domains-PR can run clean.

tests/scripts/pulp_file/test_remote.sh
```diff
@@ -17,7 +17,8 @@ expect_succ pulp file remote show --remote "cli_test_file_remote"
 HREF="$(echo "$OUTPUT" | jq -r '.pulp_href')"
 test "$(echo "$OUTPUT" | jq -r '.proxy_url')" = "http://proxy.org"
 test "$(echo "$OUTPUT" | jq -r '.max_retries')" = "5"
-test "$(echo "$OUTPUT" | jq -r '.total_timeout')" = "32"
+TIMEOUT="$(echo "$OUTPUT" | jq -r '.total_timeout')"
+test 1 -eq "$(echo "${TIMEOUT}==32" | bc)"  # total_timeout is a float that can return as either 32 or 32.0
```
gerrod3, line 21: I still don't understand the reason this can be returned as a float on different machines.

gerrod3 (approved):

tests/scripts/pulp_file/test_remote.sh
ggainey: Me either - but I decided "protection from Random Evironmental Weirdness" was the way to go. It's ugly - but, programming in bash just kind of *is*, so here we are.

mdellweg: This may actually be jq's fault (json does not really distinguish floats from ints. It only knows "number"):
```
echo "32.0" | jq '.'
32
```

mdellweg (approved):

PR #712, opened 2023-07-03
Simplify task waiting procedure

mdellweg:
Calculates task deadline upfront.

[noissue]

Note: This is to understand a nightly failure on the 3.18 better.

Review Checklist:

### DIFF
--- a/pulp-glue/pulp_glue/common/context.py
+++ b/pulp-glue/pulp_glue/common/context.py
@@ -120,7 +120,7 @@ class PulpContext:
         api_root: str,
         api_kwargs: Dict[str, Any],
         background_tasks: bool,
-        timeout: int,
+        timeout: Union[int, datetime.timedelta],
         domain: str = "default",
         format: Optional[str] = None,  # Deprecated
     ) -> None:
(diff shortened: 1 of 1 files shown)


pulp-glue/pulp_glue/common/context.py
```diff
@@ -120,7 +120,7 @@ def __init__(
         api_root: str,
         api_kwargs: Dict[str, Any],
         background_tasks: bool,
-        timeout: int,
+        timeout: Union[int, datetime.timedelta],
```
lubosmj, line 123: I like the change overall. But, can you provide an example where `datetime.timedelta` will be used?

mdellweg: Well it's an extension to the current interface, currently used nowhere. It felt like a low hanging fruit to me.

lubosmj (approved):
This PR makes the code cleaner. Still, I would leave the fruit hanging low without the need for `datetime.timedelta`. 🙈

mdellweg:
> This PR makes the code cleaner. Still, I would leave the fruit hanging low without the need for `datetime.timedelta`. 🙈

I would like to keep it, because having the same variable name being internally represented as timedelta, but not accepting a timedelta from an external call seems not nice. OTOH, we need to keep the interface accepting int, because that is what we can get from the config (and also the interface of the glue library we need to maintain).

ggainey (approved):

PR #716, opened 2023-07-07
Narrow down click version given a break change on 8.1.4.

decko:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     python_requires=">=3.6",
     install_requires=[
         "pulp-glue==0.21.0.dev",
-        "click>=8.0.0,<9.0.0",
+        "click>=8.0.0,<8.1.4",
         "PyYAML~=5.3",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -38,7 +38,7 @@
     python_requires=">=3.6",
     install_requires=[
         "pulp-glue==0.21.0.dev",
-        "click>=8.0.0,<9.0.0",
+        "click>=8.1.0,<8.1.3",
```
ggainey, line 41:
```suggestion
        "click>=8.1.0,<8.1.4",
```

ggainey (changes requested):

setup.py
ggainey: also, is there a reason to up the LB from 8.0.0? If not, let's make smallest-required-change here

decko: Ok, lowering LB to 8.0.0

ggainey (approved):

PR #721, opened 2023-07-14
Collect changelogs from all branches in main

mdellweg:
[noissue]

Review Checklist:

### DIFF
--- a/.github/workflows/changes.yml
+++ b/.github/workflows/changes.yml
@@ -1,0 +1,30 @@
+name: Collect changes
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  collect-changes:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: "main"
+          fetch-depth: 0
+      - uses: actions/setup-python@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/changes.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Collect changes
+        run: |
+          pip install GitPython
+          python3 .ci/scripts/collect_changes.py
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          title: "Update Changelog"
+          branch: "update_changes"
+          delete-branch: true
```
gerrod3, line 30: I think you need to specify the commit message to have [noissue]. Also, do you need to pass in the token for the pull request to be made?

mdellweg: Hard to say without trying...

mdellweg: [noissue] is in the commit created by the collect script. This workflow is not supposed to add another commit.

mdellweg: I was tempted to just push to main from the workflow...

gerrod3 (approved):

Issue #620, opened 2023-01-19
Add support for repositories/reclaim_space 


ggainey:
## Summary
Add support for this REST call :+1: 
https://docs.pulpproject.org/pulpcore/workflows/reclaim-disk-space.html

MichalPysik:
I'm currently working on this issue, but there is a little thing that could use a quick discussion. The call needs a list of repositories which should be used, and also list of hrefs of versions to ignore. I'm currently passing both parameters as lists, but they can also be individual duplicate command line parameters (```./program --repo "repo1" --repo "repo2" --repo repo3```). What do you think?

MichalPysik:
I've discussed with @lubosmj quite a lot, and so far settled on the idea (which can be changed if someone disagrees) of using ```{--repo one --repo two | --all} --ignored_versions [href1_, ..., href_n], meaning that one can list individual repositories (or use --all), but list is still used for the optional repository versions to ignore.

lubosmj:
@evgeni, some commands allow the following:
```
# a list of hrefs
tests/scripts/pulpcore/test_orphans.sh:    expect_succ pulp orphan cleanup --content-hrefs "[\"$content_href\"]"
```
or
```
# a list of key-value pairs
tests/scripts/pulp_file/test_content_bulk.sh:expect_succ pulp file repository modify --repository "cli_test_file_repository" --remove-content "[{\"sha256\":\"$sha256_1\",\"relative_path\":\"upload_test/test_1.txt\"},{\"sha256\":\"$sha256_2\",\"relative_path\":\"upload_test/test_2.txt\"},{\"sha256\":\"$sha256_3\",\"relative_path\":\"upload_test/test_3.txt\"}]"
```

So, another option was to list repositories in this fashion:
```
pulp repository reclaim --repos "[\"aa\", \"bb\"]"
```

I do not like it though. Being consistent does not always mean being proper. We can later focus on improving and unifying the interface of commands that accept a list of string values in the `"[\"$\"]"` format.

I am leaning more towards what @MichalPysik has proposed in the comment above,

lubosmj:
Another option for `--ignored_versions` is to allow passing pairs of `--repo-name/version-number`. Users will not need to manually extract the hrefs of versions themselves. This can be studied later.

PR #749, opened 2023-07-24
Added support for repositories/reclaim_space

MichalPysik:
The repositories/reclaim_space endpoint can now be called via the CLI.

closes: #620

### DIFF
--- a/CHANGES/620.feature
+++ b/CHANGES/620.feature
@@ -1,0 +1,1 @@
+Added the ability to reclaim disk space (cmd: 'pulp repository reclaim').
--- a/CHANGES/pulp-glue/620.feature
+++ b/CHANGES/pulp-glue/620.feature
@@ -1,0 +1,1 @@
+Added 'PulpGenericRepositoryContext' class to handle repository commands not available with subtypes.
--- a/pulp-glue/pulp_glue/common/context.py
+++ b/pulp-glue/pulp_glue/common/context.py
@@ -1259,6 +1259,32 @@ class PulpRepositoryContext(PulpEntityContext):
         return self.call("modify", parameters={self.HREF: href or self.pulp_href}, body=body)
 
 
+class PulpGenericRepositoryContext(PulpRepositoryContext):
+    """Generic repository context class to separate specific general functionality."""
+
+    def reclaim(
+        self,
+        repo_hrefs: List[t.Union[str, PulpRepositoryContext]],
+        repo_versions_keeplist: Optional[List[t.Union[str, PulpRepositoryVersionContext]]] = None,
+    ) -> Any:
+        """
+        Reclaim disk space for a list of repositories.
+
+        Parameters:
+            repo_hrefs: List of repository hrefs to reclaim.
+            repo_versions_keeplist: List of repository version hrefs to keep unaffected.
+
+        Returns:
+            Record of the reclaim space task.
+        """
+        self.pulp_ctx.needs_plugin(PluginRequirement("core", specifier=">=3.19.0"))
+        body: Dict[str, Any] = {}
+        body["repo_hrefs"] = repo_hrefs
+        if repo_versions_keeplist:
+            body["repo_versions_keeplist"] = repo_versions_keeplist
+        return self.call("reclaim_space_reclaim", body=body)
+
+
 class PulpContentContext(PulpEntityContext):
     """Base class for content contexts."""
 
--- a/pulpcore/cli/core/repository.py
+++ b/pulpcore/cli/core/repository.py
@@ -1,5 +1,13 @@
+import re
+import typing as t
+
 import click
-from pulp_glue.common.context import PulpRepositoryContext
+from pulp_glue.common.context import (
+    EntityDefinition,
+    PulpGenericRepositoryContext,
+    PulpRepositoryContext,
+    PulpRepositoryVersionContext,
+)
 from pulp_glue.common.i18n import get_translation
 
 from pulpcore.cli.common.generic import (
@@ -8,11 +16,59 @@ from pulpcore.cli.common.generic import (
     name_filter_options,
     pass_pulp_context,
     pulp_group,
+    pulp_option,
+    resource_option,
     version_command,
 )
 
 translation = get_translation(__name__)
 _ = translation.gettext
+
+
+def _version_list_callback(
+    ctx: click.Context, param: click.Parameter, value: t.Iterable[t.Tuple[str, int]]
+) -> t.Iterable[PulpRepositoryVersionContext]:
+    result = []
+    pulp_ctx = ctx.find_object(PulpCLIContext)
+    assert pulp_ctx is not None
+    for item in value:
+        pulp_href: t.Optional[str] = None
+        entity: t.Optional[EntityDefinition] = None
+
+        if item[0].startswith("/"):
+            pattern = rf"^{pulp_ctx.api_path}{PulpRepositoryContext.HREF_PATTERN}"
+            match = re.match(pattern, item[0])
+            if match is None:
+                raise click.ClickException(
+                    _("'{value}' is not a valid href for {option_name}.").format(
+                        value=value, option_name=param.name
+                    )
+                )
+            match_groups = match.groupdict(default="")
+            plugin = match_groups.get("plugin", "")
+            resource_type = match_groups.get("resource_type", "")
+            pulp_href = item[0]
+        else:
+            plugin, resource_type, identifier = item[0].split(":", maxsplit=2)
+            if not identifier:
+                raise click.ClickException(_("Repositories must be specified with plugin and type"))
+            entity = {"name": identifier}
+        context_class = PulpRepositoryContext.TYPE_REGISTRY.get(plugin + ":" + resource_type)
+        if context_class is None:
+            raise click.ClickException(
+                _(
+                    "The type '{plugin}:{resource_type}' "
+                    "is not valid for the {option_name} option."
+                ).format(plugin=plugin, resource_type=resource_type, option_name=param.name)
+            )
+        repository_ctx: PulpRepositoryContext = context_class(
+            pulp_ctx, pulp_href=pulp_href, entity=entity
+        )
+
+        entity_ctx = repository_ctx.get_version_context(number=item[1])
+        result.append(entity_ctx)
+
+    return result
 
 
 @pulp_group()
@@ -32,3 +88,55 @@ filter_options = name_filter_options
 
 repository.add_command(list_command(decorators=filter_options))
 repository.add_command(version_command(decorators=[], list_only=True))
+
+
+@repository.command()
+@resource_option(
+    "--repository",
+    "repositories",
+    context_table=PulpRepositoryContext.TYPE_REGISTRY,
+    multiple=True,
+    href_pattern=PulpRepositoryContext.HREF_PATTERN,
+    help=_(
+        "Repository in the form '[[<plugin>:]<resource_type>:]<name>' or by href."
+        " Can be called multiple times."
+    ),
+)
+@pulp_option(
+    "--all",
+    "all_repositories",
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help=_("Select all repositories."),
+)
+@pulp_option(
+    "--keep-version",
+    "keep_versions",
+    type=tuple([str, int]),
+    help=_(
+        "Repository version specified as '<repo_name_or_href> <version_number>' to keep unaffected."
+        " Can be specified multiple times."
+    ),
+    callback=_version_list_callback,
+    multiple=True,
+    required=False,
+)
+@pass_pulp_context
+def reclaim(
+    pulp_ctx: PulpCLIContext,
+    repositories: t.Iterable[str],
+    all_repositories: bool,
+    keep_versions: t.Iterable[str],
+) -> None:
+    if not (repositories or all_repositories):
+        raise click.UsageError(_("Either --repository or --all must be specified."))
+    elif repositories and all_repositories:
+        raise click.UsageError(_("Only one of --repository or --all can be specified."))
+
+    if all_repositories:
+        repositories = ["*"]
+
+    PulpGenericRepositoryContext(pulp_ctx=pulp_ctx).reclaim(
+        repo_hrefs=list(repositories), repo_versions_keeplist=list(keep_versions)
+    )
--- a/tests/scripts/pulpcore/test_repository_reclaim_space.sh
+++ b/tests/scripts/pulpcore/test_repository_reclaim_space.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
+
+pulp debug has-plugin --name "core" --min-version "3.19.0" || exit 23
+pulp debug has-plugin --name "file" || exit 23
+
+cleanup() {
+  pulp file repository destroy --name "cli_test_reclaim_repo" || true
+  pulp file remote destroy --name "cli_test_reclaim_remote" || true
+  pulp file distribution destroy --name "cli_test_reclaim_distro" || true
+}
+trap cleanup EXIT
+
+# Setup needed prerequisites
+expect_succ pulp file remote create --name "cli_test_reclaim_remote" --url "$FILE_REMOTE_URL" --policy immediate
+expect_succ pulp file repository create --name "cli_test_reclaim_repo" --remote "cli_test_reclaim_remote" --autopublish
+expect_succ pulp file repository sync --name "cli_test_reclaim_repo"
+expect_succ pulp file distribution create --name "cli_test_reclaim_distro" --base-path "cli_test_reclaim_distro" --repository "cli_test_reclaim_repo"
+
+# Verify that the artifact was created
+expect_succ pulp file repository content list --repository "cli_test_reclaim_repo" --version 1
+ARTIFACT="$(echo "$OUTPUT" | jq -r '.[] | select(.relative_path | endswith("1.iso")) | .artifact')"
+test "$ARTIFACT" != "null"
+
+# Reclaim disk space with repo version 1 on keeplist and verify that the artifact wasn't affected
+expect_succ pulp repository reclaim --repository "file:file:cli_test_reclaim_repo" --keep-version "file:file:cli_test_reclaim_repo" 1
+expect_succ pulp file repository content list --repository "cli_test_reclaim_repo" --version 1
+test "$(echo "$OUTPUT" | jq -r '.[] | select(.relative_path | endswith("1.iso")) | .artifact')" = "$ARTIFACT"
+
+# Reclaim disk space and verify that the artifact is no longer available
+expect_succ pulp repository reclaim --repository "file:file:cli_test_reclaim_repo"
+expect_succ pulp file repository content list --repository "cli_test_reclaim_repo" --version 1
+test "$(echo "$OUTPUT" | jq -r '.[] | select(.relative_path | endswith("1.iso")) | .artifact')" = "null" 


gerrod3 (commented):
Can you add a test for this?

MichalPysik:
> Can you add a test for this?

Haha, that's actually the reason why I'm keeping it as draft for now.

pulpcore/cli/common/generic.py
```diff
@@ -805,6 +805,13 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: t.Optional
     expose_value=False,
 )
 
+repositories_option = click.option(
```
lubosmj, line 808: It might be worth adding the callback (e.g., `get_list_of_repos` returning a list of contexts) here so it can be used by other commands in the future.

mdellweg: It should be a resource option in the first place.
Also i am beginning to think that there is no nice way to interface this feature.

MichalPysik: @mdellweg I also think so. I initially though that working with multiple repositories is quite a generic thing, but later found out it actually isn't, so it's quite hard to think of a nice interface that doesn't break the style of the other commands.

mdellweg: Multiple repositories is not the issue. See https://github.com/pulp/pulp-cli/blob/main/pulpcore/cli/core/exporter.py#L26
Multiple versions is a bit harder: https://github.com/pulp/pulp-cli/blob/main/pulpcore/cli/core/export.py#L131

The issue is an api endpoint that is supposed to "Work on these repositories, but don't touch these versions.".

MichalPysik: I made selecting repositories a resource option now, the only difference now is that you have to type (example) "file:file:my_repo" instead of "my_repo", which may slow down the experience, but I actually like it for differentiating between different plugins' repositories. 

When it comes to the keeplisted versions, I let it be as a simple list of hrefs. I can make it more elegant if it's really needed (e.g. entering the version as an integer), but otherwise I'm for keeping it this simple as I believe the option is not used that much often (I may be wrong tho).

mdellweg: Have you seen this? https://github.com/pulp/pulp-cli/blob/main/tests/scripts/pulpcore/test_pulpexporter.sh#L58
Consistency is key.

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +42,44 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
 
 repository.add_command(list_command(decorators=filter_options))
 repository.add_command(version_command(decorators=[], list_only=True))
+
+
+# needs pulpcore 3.19.0
```
lubosmj, line 47:
```suggestion
```

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +42,44 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
...
+@repositories_option
+@pulp_option(
+    "--all",
+    "all_repos",
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help=_("Select all repositories."),
+)
+@click.option(
```
lubosmj, line 60:
```suggestion
@pulp_option(
```

MichalPysik:
If anyone knows why Lint wants me to replace an empty line with an empty line, please let me know. Otherwise I think the PR is ready to merge.

lubosmj:
Selecting the text tells me there are some whitespace characters:
image

MichalPysik:
> Selecting the text tells me there are some whitespace characters: image

All checks have passed now 😄

pulpcore/cli/common/generic.py
```diff
@@ -805,6 +805,18 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: t.Optional
...
 
+repositories_option = resource_option(
+    "--repository",
+    "repositories",
+    context_table=PulpRepositoryContext.TYPE_REGISTRY,
+    multiple=True,
+    href_pattern=PulpRepositoryContext.HREF_PATTERN,
+    help=_(
+        "Repository to export from in the form '[[<plugin>:]<resource_type>:]<name>' or by href."
+        " Can be called multiple times."
```
mdellweg, line 816: This help text is not valid. Maybe it's not worth making the option generic.

CHANGES/620.feature
```diff
@@ -0,0 +1 @@
+Added support for the 'repositories/reclaim_space' request to be called via the CLI.
```
mdellweg, line 1: This is user facing, so don't talk about the api, but about features.
Also a changelog for glue is missing.

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +47,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
 
 repository.add_command(list_command(decorators=filter_options))
 repository.add_command(version_command(decorators=[], list_only=True))
+
+
+@repository.command()
+@pass_repository_context
+@pass_pulp_context
```
mdellweg, line 54: The pass decorators should go last.

pulpcore/cli/core/repository.py
mdellweg: ^

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +47,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
...
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help=_("Select all repositories."),
+)
+@click.option(
+    "--keep-version",
+    "keep_versions",
+    type=tuple([str, int]),
+    help=_("List of (pulp hrefs) repository versions to keep."),
```
mdellweg, line 68: This is not true.

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +47,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
...
+        raise click.UsageError(_("Either --repository or --all must be specified."))
+    if all_repos:
+        repository_hrefs = ["*"]
+    else:
+        repository_hrefs = [
+            repo.pulp_href for repo in repositories if isinstance(repo, PulpRepositoryContext)
+        ]
+
+    versions_keeplist = [version.pulp_href for version in keep_versions]
+    print(versions_keeplist)
```
mdellweg, line 91: This is not needed.

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +47,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
...
+    required=False,
+)
+def reclaim_space(
+    pulp_ctx: PulpCLIContext,
+    repository_ctx: PulpRepositoryContext,
+    repositories: Iterable[EntityFieldDefinition],
+    all_repos: bool,
+    keep_versions: Iterable[PulpRepositoryVersionContext],
+) -> None:
+    pulp_ctx.needs_plugin(PluginRequirement("core", specifier=">=3.19.0"))
```
mdellweg, line 80: This goes in the glue layer.

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +47,49 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
...
 repository.add_command(list_command(decorators=filter_options))
 repository.add_command(version_command(decorators=[], list_only=True))
+
+
+@repository.command()
+@pass_repository_context
+@pass_pulp_context
+@repositories_option
+@pulp_option(
+    "--all",
```
mdellweg, line 57: Is this a good name, or too ambiguous?

@gerrod3 @ggainey

pulpcore/cli/common/generic.py
```diff
@@ -1448,3 +1460,58 @@ def content_group() -> None:
             content_group.add_command(command)
 
     return content_group
+
+
+def _version_list_callback(
```
mdellweg, line 1465: Why is this function hiding away at the end of the file far away from the other fellow callbacks? Also don't make an export start with `_`.

pulpcore/cli/core/repository.py
```diff
@@ -1,20 +1,35 @@
+from typing import Iterable
```
mdellweg, line 1:
```suggestion
import typing as t
```
Seems to be more convenient in the long run.

tests/scripts/pulpcore/test_repository_reclaim_space.sh
```diff
@@ -0,0 +1,46 @@
...
+DIST_BASE_URL="$(echo "$OUTPUT" | jq -r '.base_url')"
+
+# Download the artifact and verify that it exists
+expect_succ curl "${DIST_BASE_URL}1.iso"
+expect_succ pulp file repository content list --repository "cli_test_reclaim_repo" --version 1
+ARTIFACT="$(echo "$OUTPUT" | jq -r '.[] | select(.relative_path | endswith("1.iso")) | .artifact')"
+if [ "$ARTIFACT" = "null" ]
+then
+  exit 1
+fi
```
mdellweg, line 30:
```suggestion
test "$ARTIFACT" != "null"
```

tests/scripts/pulpcore/test_repository_reclaim_space.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# On demand policy ensures that the artifacts are not downloaded until it is requested
+expect_succ pulp file remote create --name "cli_test_reclaim_remote" --url "$FILE_REMOTE_URL" --policy on_demand
+expect_succ pulp file repository create --name "cli_test_reclaim_repo" --remote "cli_test_reclaim_remote" --autopublish
+expect_succ pulp file repository sync --name "cli_test_reclaim_repo"
+expect_succ pulp file distribution create --name "cli_test_reclaim_distro" --base-path "cli_test_reclaim_distro" --repository "cli_test_reclaim_repo"
+DIST_BASE_URL="$(echo "$OUTPUT" | jq -r '.base_url')"
+
+# Download the artifact and verify that it exists
+expect_succ curl "${DIST_BASE_URL}1.iso"
```
mdellweg, line 24: make the sync immediate instead?

pulp-glue/pulp_glue/common/context.py
```diff
@@ -883,6 +883,18 @@ def modify(
             body["base_version"] = base_version
         return self.call("modify", parameters={self.HREF: href or self.pulp_href}, body=body)
 
+    def reclaim_space(
```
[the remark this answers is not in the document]
mdellweg: Still unsure about this. But the command is not added to all the plugin sections.
It's just available on all the contexts where it is doomed to fail.

mdellweg (changes requested):

pulpcore/cli/common/generic.py
MichalPysik: @mdellweg thanks, I fixed the text, I just copy-pasted the original and then forgot to change it. I really think it is a generic thing, as its purpose is to be able to obtain a list of repositories in the most practical way, I see nothing too specialized about it.

tests/scripts/pulpcore/test_repository_reclaim_space.sh
MichalPysik: I though the reclaim space command is mostly useful for cases where the content is downloaded on demand. The content gets stored when user requests it, then it can be deleted via reclaim-space, and then the user can download it again with the on_demand policy.

pulpcore/cli/core/repository.py
MichalPysik: @mdellweg if you only mean the print, that's left there by accident. Or are you referring to the variables as well?

pulpcore/cli/core/repository.py
MichalPysik: Another leftover from previous state of the PR, thanks for the heads up

pulpcore/cli/core/repository.py
mdellweg: No i mean lines 86 to 91

pulp-glue/pulp_glue/common/context.py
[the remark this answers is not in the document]
MichalPysik: We talked about this with @lubosmj futher and I think the current state is the best compromise.

You have to enter ```pulp repository reclaim-space ...``` , currently there is no way to call it in a way like ```pulp file repository reclaim-space ...``` but I believe it could be added easily if needed, since the method is in a class from which the plugin specific repository-sub-classes inherit (which wouldnt be possible if we created something like PulpCoreRepositoryContext to stop the inheritance). The individual plugins are also fairly differentiated between in the current state, since the user has to specify "file:file:repo" instead of "repo", which helps in avoiding mistakes.

pulpcore/cli/core/repository.py
MichalPysik: you dont like storing the lists into variables and would prefer passing then straight there, or is there something else I'm missing?

tests/scripts/pulpcore/test_repository_reclaim_space.sh
gerrod3: It is useful in any case where the content has been downloaded onto the system. You can perform an immediate sync and then turn the repo into an on-demand repo by using reclaim. When Pulp syncs content, immediate or on-demand we remember where we got the content from so we can download it again in the future.

pulpcore/cli/core/repository.py
gerrod3: I think it is fine, the command would be `pulp repository reclaim-space --all`. An alternative would be `all-repos`, but it should be obvious what this field does looking at the help message.

pulpcore/cli/core/repository.py
MichalPysik: @mdellweg can you elaborate please?

pulpcore/cli/core/repository.py
mdellweg: No you ought to pass the contexts to the glue layer. No conversion at this level.

pulpcore/cli/core/repository.py
mdellweg: The information that this feature needs a certain pulpcore version is not limited to the cli and should be asserted on all users of pulp-glue.

pulp-glue/pulp_glue/common/context.py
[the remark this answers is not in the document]
mdellweg: No you can't reclaim is only provided on the generic api endpoint. Any attempt to call this method on the PulpFileRepositoryContext will result in a key error looking up the corresponding operation in the api-spec.

Also this question must be answered with the glue-librarian-hat on. There is no arguing from the cli perspective.

pulp-glue/pulp_glue/common/context.py
```diff
@@ -1256,6 +1256,22 @@ def modify(
...
+        self,
+        repo_hrefs: List[str],
+        repo_versions_keeplist: Optional[List[str]] = None,
+    ) -> Any:
+        self.pulp_ctx.needs_plugin(PluginRequirement("core", specifier=">=3.19.0"))
+        body: Dict[str, Any] = {}
+        body["repo_hrefs"] = repo_hrefs
+        if repo_versions_keeplist:
+            body["repo_versions_keeplist"] = repo_versions_keeplist
+        return self.call("reclaim_space_reclaim", body=body)
```
mdellweg, line 1272: Is the id `repositories_reclaim_space_reclaim`? Tztztz.

pulp-glue/pulp_glue/common/context.py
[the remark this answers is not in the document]
mdellweg: Can we just call it reclaim? Would there be any ambiguity?

pulpcore/cli/core/repository.py
```diff
@@ -1,20 +1,35 @@
...
 )
 
 translation = get_translation(__name__)
 _ = translation.gettext
 
 
+def repo_href_from_name(name: str, pulp_ctx: PulpCLIContext) -> str:
+    return PulpRepositoryContext(pulp_ctx, entity={"name": name}).pulp_href
+
+
```
mdellweg, line 32: Not needed.

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +47,44 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
 
 repository.add_command(list_command(decorators=filter_options))
 repository.add_command(version_command(decorators=[], list_only=True))
+
+
+@repository.command()
+@repositories_option
+@pulp_option(
+    "--all",
+    "all_repos",
```
mdellweg, line 56:
```suggestion
    "all_repositories",
```

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +47,44 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
...
+        "Repository version specified as '<repo_name_or_href> <version_number>' to keep unaffected."
+        " Can be specified multiple times."
+    ),
+    callback=version_list_callback,
+    multiple=True,
+    required=False,
+)
+@pass_repository_context
+@pass_pulp_context
+def reclaim_space(
```
mdellweg, line 76: Also would "reclaim" be sufficient?

pulpcore/cli/common/generic.py
```diff
@@ -807,6 +862,18 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: t.Optional
...
+    "repositories",
+    context_table=PulpRepositoryContext.TYPE_REGISTRY,
+    multiple=True,
+    href_pattern=PulpRepositoryContext.HREF_PATTERN,
+    help=_(
+        "Repository in the form '[[<plugin>:]<resource_type>:]<name>' or by href."
+        " Can be called multiple times."
+    ),
+)
+
```
mdellweg, line 876: I don't see this being reused anywhere. Why should we make it generic? (Most other places use an additional capability.)

pulpcore/cli/common/generic.py
```diff
@@ -489,6 +489,61 @@ def null_callback(
     return value
 
 
+def version_list_callback(
```
mdellweg, line 492: What makes this generic? How is it different from the one used in export?

mdellweg (changes requested):

pulpcore/cli/common/generic.py
MichalPysik: @mdellweg I just think it is generic by nature, choosing multiple repositories just doesn't seem too specific in the context of Pulp imo

mdellweg: The resource option is the generic thing here.
When selecting arbitrary repositories, usually we need to inject an additional capabilities. This makes this version just not so generic.

pulpcore/cli/common/generic.py
mdellweg: Is it different now?

MichalPysik: @mdellweg except for one line, which was removed by a different PR while my PR was active - works the same for this use case

mdellweg: So one line is different, but we are still using the same callback?

CHANGES/620.feature
```diff
@@ -0,0 +1 @@
+Added the ability to reclaim disk space (cmd: 'pulp repository reclaim-space').
```
mdellweg, line 1:
```suggestion
Added the ability to reclaim disk space (cmd: 'pulp repository reclaim').
```

pulpcore/cli/core/repository.py
```diff
@@ -8,8 +10,11 @@
     name_filter_options,
     pass_pulp_context,
     pulp_group,
+    pulp_option,
+    resource_option,
     version_command,
 )
+from pulpcore.cli.core.export import _version_list_callback
```
mdellweg, line 17: That is not looking like a proper import.
I think the callback belongs in core.generics.

mdellweg (changes requested):

pulpcore/cli/common/generic.py
MichalPysik: @mdellweg no, the line has nothing to do with my PR

let me put it this way (arbitary numbers of days):
- on day 1, I copy the function to generic.py
- on day 15, a PR (not mine) which adds a line to the original function gets merged to main (and the new line has no effect on my use case)
- on day 30, I delete the copy of the (older version) function and decide to use the main one, which is updated now, but it doesn't matter

CHANGES/620.feature
MichalPysik: @mdellweg thanks, I missed that one

pulpcore/cli/core/repository.py
MichalPysik: @mdellweg Fixing it right now, just want to say that we went export -> common.generic -> export -> core.generic so far

CHANGES/pulp-glue/620.feature
```diff
@@ -0,0 +1,2 @@
+Added 'PulpGenericRepositoryContext' class to handle some repository commands
+not needing a plugin-specific implementation.
```
mdellweg, line 2:
```suggestion
Added 'PulpGenericRepositoryContext' class to handle repository commands not available with subtypes.
```
Also don't do the crazy line breaking here (specific to this repository).

pulp-glue/pulp_glue/common/context.py
```diff
@@ -1259,6 +1259,22 @@ def modify(
...
 
 
+class PulpGenericRepositoryContext(PulpRepositoryContext):
+    """Generic repository context class to separate specific general functionality."""
+
+    def reclaim(
+        self,
+        repo_hrefs: List[str],
+        repo_versions_keeplist: Optional[List[str]] = None,
+    ) -> Any:
```
mdellweg, line 1269: Please add a docstring.
Use `make servedocs` to see how it is rendered.

pulpcore/cli/core/repository.py
```diff
@@ -32,3 +88,52 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
...
+)
+@pass_pulp_context
+def reclaim(
+    pulp_ctx: PulpCLIContext,
+    repositories: t.Iterable[str],
+    all_repositories: bool,
+    keep_versions: t.Iterable[str],
+) -> None:
+    if not (repositories or all_repositories):
+        raise click.UsageError(_("Either --repository or --all must be specified."))
```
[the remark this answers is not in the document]
MichalPysik: @mdellweg in my opinion, all_repositories should have a higher priority and therefore then ignore any specified repository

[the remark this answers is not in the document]
mdellweg: It's more likely the user didn't know what they were doing when specifying both. So by the principle of least surprise, failing immediately is better.

CHANGES/pulp-glue/620.feature
```diff
@@ -0,0 +1,2 @@
+Added 'PulpGenericRepositoryContext' class to handle repository commands not available
+with subtypes.
```
mdellweg, line 2:
```suggestion
Added 'PulpGenericRepositoryContext' class to handle repository commands not available with subtypes.
```

pulp-glue/pulp_glue/common/context.py
```diff
@@ -1259,6 +1259,32 @@ def modify(
         return self.call("modify", parameters={self.HREF: href or self.pulp_href}, body=body)
 
 
+class PulpGenericRepositoryContext(PulpRepositoryContext):
+    """Generic repository context class to separate specific general functionality."""
+
+    def reclaim(
+        self,
+        repo_hrefs: List[str],
```
mdellweg, line 1267: I thought we could pass strings or RepositoryContexts here.

pulp-glue/pulp_glue/common/context.py
```diff
@@ -1259,6 +1259,32 @@ def modify(
...
+        repo_versions_keeplist: Optional[List[str]] = None,
```
mdellweg, line 1268: Same here.

pulp-glue/pulp_glue/common/context.py
```diff
@@ -1259,6 +1259,32 @@ def modify(
...
+        """
+        Reclaim disk space for a list of repositories.
+
+        Parameters:
+            repo_hrefs: List of repository hrefs to reclaim.
+            repo_versions_keeplist: List of repository version hrefs to keep unaffected.
+
+        Returns:
+            Record of the reclaim space task.
+        """
```
mdellweg, line 1279: Great docstring!

mdellweg (approved):

PR #751, opened 2023-07-28
Fix schema patching for pulpcore>=3.23,<3.30

gerrod3:
As a part of my quest to fix the oci-images CI, I have discovered that recent schema fixes has broken the CLI when performing patching. This should make it more resilient.


mdellweg:
Too bad, i looked for an issue and didn't see this.

In my pr there is another one identified:
https://github.com/pulp/pulp-cli/pull/753

gerrod3:
Thanks for fixing this, closing since it is a dup.

PR #765, opened 2023-08-05
Improves the replicate command test

dkliban:
When the tests are run in pulpcore or plugin GitHub repos, there is a .netrc file present so the replicate task succeeds. Providing a bad API root ensures that the task always fails.

### DIFF
--- a/tests/scripts/pulpcore/test_upstream_pulp.sh
+++ b/tests/scripts/pulpcore/test_upstream_pulp.sh
@@ -10,13 +10,13 @@ cleanup() {
 }
 trap cleanup EXIT
 
-expect_succ pulp upstream-pulp create --name "cli_test_upstream_pulp" --base-url "$PULP_BASE_URL" --api-root "/pulp/"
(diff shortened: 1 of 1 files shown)


tests/scripts/pulpcore/test_upstream_pulp.sh
```diff
@@ -10,13 +10,13 @@ cleanup() {
 }
 trap cleanup EXIT
 
-expect_succ pulp upstream-pulp create --name "cli_test_upstream_pulp" --base-url "$PULP_BASE_URL" --api-root "/pulp/"
+expect_succ pulp upstream-pulp create --name "cli_test_upstream_pulp" --base-url "$PULP_BASE_URL" --api-root "/123/"
```
mdellweg, line 13: Nice idea.
I was wondering if the server on the upstream side should validate that there are credentials provided. But now that I think of it, you seem to be able to provide them via `.netrc`. So it would be right to not fail. Is this a feature or a bug?

mdellweg (approved):

PR #767, opened 2023-08-08
Use github caches for pip files

mdellweg:
[noissue]

Review Checklist:


gerrod3:
Interesting, will this speed up the installs for all the CI runners now that they won't have to pull from PyPI every single time? This could be very useful in the plugin CIs as well.

mdellweg:
> Interesting, will this speed up the installs for all the CI runners now that they won't have to pull from PyPI every single time? This could be very useful in the plugin CIs as well.

If they didn't run pip inside the container, yes. (Or we can maybe mount it in the container?)

PR #770, opened 2023-08-11
Fixed crashing pulp domain

### LINKED ISSUE #769: 'pulp domain' leads to a stack trace when a default value for --domain is provided in the config file.
~~When writing 'pulp' in the cli, 'domain' shows as one of the available commands, but~~ entering 'pulp domain (...)' always leads to a Python traceback. ~~Right way to access pulp domain commands is typing 'pulp domains (...)'.~~

~~Typing 'pulp domains' shows a valid domain menu, so the problem applies to the top-level help list.~~

edit:
It's a conflict between --domain and domain and the way click handles default_maps for child contexts.
https://click.palletsprojects.com/en/8.1.x/commands/#overriding-defaults: "The default map can be nested arbitrarily for each subcommand:"
Long story short, if you remove the "domain" entry from the config file (that is where the string comes from that it later tries to use as a new default_map), we are fine again.

mdellweg:
Providing a default value for --domain lead to the domain subcommand group getting an invalid default_map.

fixes #769

Review Checklist:

### DIFF
--- a/CHANGES/769.bugfix
+++ b/CHANGES/769.bugfix
@@ -1,0 +1,1 @@
+Fixed a crash in `pulp domain` when a default value for `--domain` was provided in the config file.
--- a/pulpcore/cli/core/domain.py
+++ b/pulpcore/cli/core/domain.py
@@ -21,7 +21,9 @@ translation = get_translation(__name__)
 _ = translation.gettext
 
 
-@pulp_group()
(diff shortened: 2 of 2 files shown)


ggainey (approved):
I was going to ask for an extension to our domain-tests to catch this - and discovered we don't *have* domain-tests. Opened #771 to fix that, let's leave this PR to fix the immediate problem.

PR #792, opened 2023-09-25
Add version selection widget to docs

mdellweg:
[noissue]

Review Checklist:


docs_theme/main.html
```diff
@@ -0,0 +1,57 @@
+{% extends "base.html" %}
+
+{% block scripts %}
+{{ super() }}
+<script>
+  fetch('{{ 'versions.txt'|url }}')
+    .then(function(res) {if (res.ok) return res.text();})
```
gerrod3, line 7: Funny to see you using the function syntax, most JS I see uses anonymous functions for their callbacks.

I think you need to also supply the onRejected argument for these promises incase there is a network error.

docs_theme/main.html
```diff
@@ -0,0 +1,57 @@
...
+    .then(function(text) {return text.trim().split("\n");})
```
gerrod3, line 8: I'm pretty sure this function could be called with undefined and throw an error. The above `then` only returns the text if the res is OK, a 404 will not throw an error and thus it will pass undefined to this function.

docs_theme/main.html
```diff
@@ -0,0 +1,57 @@
...
+        option.value = '/en/' + version;
+        version_select.add(option);
+      });
+      {% if config.theme.version.endswith("dev") %}
+      version_select.value = "";
+      {% else %}
+      version_select.value = "{{ config.theme.version }}";
+      {% endif %}
+      version_select.onchange = function(event) { window.location = '{{ '.'|url }}' + event.target.value };
+      console.log(version_select)
```
gerrod3, line 33: Do you want to leave this log statement in?

gerrod3 (changes requested):

docs_theme/main.html
mdellweg: This is an anonymous function. It does not have a name.
What are we supposed to do in the rejected code path? I think, we just don't add the widget to the dom tree.

docs_theme/main.html
mdellweg: Nope...

docs_theme/main.html
gerrod3: Woops, meant the arrow syntax is what I always see. 

Yeah probably do nothing on error. I guess we probably don't have to add a handler.

docs_theme/main.html
mdellweg: I added a codepath for res not OK.

PR #797, opened 2023-09-27
Generate command reference in docs

mdellweg:
[noissue]

Review Checklist:

### DIFF
--- a/pulp_cli/__init__.py
+++ b/pulp_cli/__init__.py
@@ -1,33 +1,22 @@
+import typing as t
 from types import ModuleType
-from typing import Any, Dict, Optional
 
 import click
 import pkg_resources
 
 __version__ = "0.22.0.dev"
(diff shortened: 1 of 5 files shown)


pulp_cli/__init__.py
gerrod3, line 1: Did you need to change the way the main command is loaded for the docs building?

mdellweg: Yes, unfortunately.

gerrod3: Is there any downsides to this new approach?

mdellweg: If i remember correctly, i wanted to have a way to defer plugin loading until one could select the plugins to load in a config file. But that never worked. So probably no.

mdellweg:
I am still not proud, how the commands are enumerated (without the PulpContexts in particular) here. Let's not merge it this way.

PR #809, opened 2023-10-04
Add translations to pulp-glue

### LINKED ISSUE #634: [task] make the i18n features work with the glue lib

mdellweg:
fixes #634

Review Checklist:


gerrod3:
What is the difference between `.pot` and `.po` files? Also, do you plan to add more translations to the message strings?

mdellweg:
If `po` stands for "i don't know what", `pot` is "template for i don't know what".

I think, you extract the strings into the template, multtiply the template into all the languages as po files, and add the translations there. Then you compile them into mo files and ship those with the binaries.

No I don't intend to do any more translations now.

gerrod3 (approved):

PR #822, opened 2023-11-06
Remove the option to choose a checksum weaker than sha256 when publishing rpm repos

dralley:
publishing RPM repos

[noissue]


pulpcore/cli/rpm/common.py
```diff
@@ -1 +1 @@
-CHECKSUM_CHOICES = ("unknown", "md5", "sha1", "sha224", "sha256", "sha384", "sha512")
+CHECKSUM_CHOICES = ("sha256", "sha384", "sha512")
```
mdellweg, line 1: Please find a way to support the old versions of pulp_rpm still.

dralley: I'm about 99.9% certain that nobody has ever used this option in the first place.  Every RHEL since RHEL 6 has supported SHA256.  "unknown" actually shouldn't have been here in the first place.

But if I do - is there a pre-existing framework for tweaking the choices based on available versions of pulp/plugins?  I see `PluginRequirement` but it's not immediately clear if it would work in this context, as it has no "context" (object) available to operate off of.

https://github.com/pulp/pulp-cli/blob/589f9f4fdd0d7222141df6c53861fdfd990e4850/pulpcore/cli/rpm/repository.py#L139

dralley:
Commit message fixed.

pulpcore/cli/rpm/common.py
mdellweg: Let me see. First we should talk about pulp_rpm versions here. So let's assume these choices were removed in pulp_rpm version x.y.z.
Remove the `"unknown"` because that was never useful. But keep the reset, maybe add a note to the helpstring of the option. Then in the context object in the glue part there should be a `preprocess_entity` where based on the choice made, you can issue `self.pulp_ctx.needs_plugin(PluginRequirement("rpm", specifier=">=x.y.z", inverted=True, feature="deprecated checksums")`.

dralley:
Updated.

pulpcore/cli/rpm/repository.py
```diff
@@ -144,17 +144,32 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
...
+        "--package-checksum-type",
+        type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False),
+        help=_(
+            "DEPRECATED: Option specifying the checksum type to use for packages in "
+            "repository metadata."
+        ),
+    ),
+    pulp_option(
+        "--checksum-type",
+        type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False),
```
mdellweg, line 161: I have the impression (without digging too deep into pulp_rpm) that this should be limited to the strong checksum algorithms.
Can you add a test to show this option in action?

dralley: Is the test going to work, if the feature is merged but not yet released?

pulpcore/cli/rpm/repository.py
```diff
@@ -144,17 +144,31 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
...
-        "--package-checksum-type", type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False)
+        "--package-checksum-type",
+        type=click.Choice(CHECKSUM_CHOICES, case_sensitive=False),
+        help=_(
+            "DEPRECATED: Option specifying the checksum type to use for packages in "
+            "repository metadata."
+        ),
+    ),
+    click.option(
+        "--checksum-type",
```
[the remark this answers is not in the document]
dralley: Updated - not 100% sure this is correct but I used `pulp_option` with `needs_plugins`

pulpcore/cli/rpm/common.py
```diff
@@ -1 +1 @@
-CHECKSUM_CHOICES = ("unknown", "md5", "sha1", "sha224", "sha256", "sha384", "sha512")
+CHECKSUM_CHOICES = ("md5", "sha1", "sha224", "sha256", "sha384", "sha512")
```
mdellweg, line 1: So i think we need this now:
```suggestion
LEGACY_CHECKSUM_CHOICES = ("md5", "sha1", "sha224", "sha256", "sha384", "sha512")
CHECKSUM_CHOICES = ("sha256", "sha384", "sha512")
```

pulpcore/cli/rpm/repository.py
mdellweg: I'd expect one test to "use" the option. Before it's released but merged, it may be guarded with `pulp_rpm>=3.25.0.dev`. The "nighty" test is running with an all sources container.

pulpcore/cli/rpm/repository.py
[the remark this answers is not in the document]
mdellweg: Until pulp_rpm 3.25.0 is released, you need to make the guard use 3.25.0.dev i think.

pulp-glue/pulp_glue/rpm/context.py
```diff
@@ -299,6 +316,23 @@ def preprocess_entity(self, body: EntityDefinition, partial: bool = False) -> En
...
+                        "rpm", specifier=">=3.25.0", inverted=True, feature=_("weak checksums")
+                    )
+                )
+            if package_checksum_type and package_checksum_type in disallowed_checksums:
+                self.pulp_ctx.needs_plugin(
+                    PluginRequirement(
+                        "rpm", specifier=">=3.25.0", inverted=True, feature=_("weak checksums")
+                    )
+                )
         return body
```
mdellweg, line 336: I thought, we could add a needs_plugin here for "checksum_type" too.

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -127,6 +127,17 @@ then
   expect_fail pulp rpm repository sync --repository "cli_test_rpm_repository" --sync-policy foobar
 fi
 
+if pulp debug has-plugin --name "rpm" --min-version "3.25.0.dev"
+then
+  expect_succ pulp rpm publication create --repository "cli_test_rpm_repository" --checksum-type sha512
```
mdellweg, line 132: You added the option for repository. I think the tests are failing, because this is the publication command.

.ci/settings/settings.py
```diff
@@ -2,3 +2,4 @@
 ALLOWED_EXPORT_PATHS = ["/tmp"]
 ORPHAN_PROTECTION_TIME = 0
 TELEMETRY = False
```
mdellweg, line 4: This should be ANALYTICS, right?

CHANGES/850.feature
```diff
@@ -0,0 +1 @@
+Added support for --checksum-type option (combination of --package-checksum-type and --metadata-checksum-type) when creating publications and configuring repository publish settings.
```
mdellweg, line 1:
```suggestion
Added support for `--checksum-type` option (combination of `--package-checksum-type` and `--metadata-checksum-type`) when creating rpm publications and configuring rpm repository publish settings.
```

mdellweg (approved):

Issue #831, opened 2023-11-23
Remove sqlite-metadata option from "pulp-rpm rpm update"


pedro-psb:
## Summary
As we are removing sqlite metadata support in pulp-rpm (pulp/pulp_rpm#2457), we should remove it as an option in the CLI aswell.

The only place it appears is in `pulp rpm repository update` as a boolean option.

PR #832, opened 2023-11-23
Remove sqlite-metadata option from `rpm update` cmd

pedro-psb:
Removed from "pulp rpm repository update" command. The reason is that it was phased out from pulp_rpm plugin. https://github.com/pulp/pulp_rpm/issues/2457

fixes: #831

Review Checklist:

### DIFF
--- a/CHANGES/831.removal
+++ b/CHANGES/831.removal
@@ -1,0 +1,1 @@
+Marked option `--sqlite-metadata` on `pulp rpm repository update/create` unavailable for `pulp_rpm>=3.25.0`, as it was removed there.
--- a/pulp-glue/pulp_glue/rpm/context.py
+++ b/pulp-glue/pulp_glue/rpm/context.py
@@ -211,6 +211,15 @@ class PulpRpmPublicationContext(PulpPublicationContext):
                     feature=_("customization of the config.repo file"),
                 )
             )
+        if "sqlite_metadata" in body:
+            self.pulp_ctx.needs_plugin(
+                PluginRequirement(
+                    "rpm",
+                    specifier=">=3.25.0",
+                    inverted=True,
+                    feature=_("sqlite_metadata generation"),
+                )
+            )
         return body
 
 
@@ -279,6 +288,15 @@ class PulpRpmRepositoryContext(PulpRepositoryContext):
                     "rpm",
                     specifier=">=3.24.0",
                     feature=_("customization of the config.repo file"),
+                )
+            )
+        if "sqlite_metadata" in body:
+            self.pulp_ctx.needs_plugin(
+                PluginRequirement(
+                    "rpm",
+                    specifier=">=3.25.0",
+                    inverted=True,
+                    feature=_("sqlite_metadata generation"),
                 )
             )
         return body
--- a/pulpcore/cli/rpm/repository.py
+++ b/pulpcore/cli/rpm/repository.py
@@ -167,7 +167,14 @@ update_options = [
             on the repodata."""
         ),
     ),
-    click.option("--sqlite-metadata/--no-sqlite-metadata", default=None),
+    click.option(
+        "--sqlite-metadata/--no-sqlite-metadata",
+        default=None,
+        help=_(
+            """DEPRECATED: An option specifying whether Pulp should generate SQLite metadata.
+            Unavailable for pulp_rpm>=3.25.0"""
+        ),
+    ),
     pulp_option(
         "--autopublish/--no-autopublish",
         needs_plugins=[PluginRequirement("rpm", specifier=">=3.12.0")],
--- a/tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
+++ b/tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
@@ -42,6 +42,14 @@ expect_succ pulp rpm repository update --repository "cli_test_rpm_repository" --
 expect_succ pulp rpm repository show --repository "cli_test_rpm_repository"
 test "$(echo "$OUTPUT" | jq -r '.description')" = "null"
 
+# sqlite metadata removal from pulp_rpm (pulp/pulp_rpm#3328)
+if pulp debug has-plugin --name "rpm" --specifier "<3.25.0"
+then
+  expect_succ pulp rpm repository update --name "cli_test_rpm_repository" --no-sqlite-metadata
+else
+  expect_fail pulp rpm repository update --name "cli_test_rpm_repository" --no-sqlite-metadata
+fi
+
 if pulp debug has-plugin --name "rpm" --specifier ">=3.24.0"
 then
 expect_succ pulp rpm repository create --name "cli_test_rpm_repository2" --repo-config "@repo_config.json"
--- a/CHANGES/pulp-glue/831.removal
+++ b/CHANGES/pulp-glue/831.removal
@@ -1,0 +1,1 @@
+Added version restriction to prevent the use of `sqlite_metadata` attribute on Repository and Publication contexts for `pulp_rpm>=3.25.0`.


mdellweg (changes requested):
Please read the docs first. Specifically these parts:

https://docs.pulpproject.org/pulp_cli/contributing/
https://docs.pulpproject.org/pulp_cli/architecture/

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -42,6 +42,12 @@ expect_succ pulp rpm repository update --repository "cli_test_rpm_repository" --
 expect_succ pulp rpm repository show --repository "cli_test_rpm_repository"
 test "$(echo "$OUTPUT" | jq -r '.description')" = "null"
 
+# sqlite metadata removal from pulp-cli (pulp/pulp_rpm#3328)
+if pulp debug has-plugin --name "rpm" --specifier ">3.24.0"
+then
+  expect_fail pulp rpm repository update --name "cli_test_rpm_repository" --no-sqlite-metadata
+fi
```
mdellweg, line 49: I think we can do a positive test too.

pulpcore/cli/rpm/repository.py
```diff
@@ -159,7 +159,11 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
             on the repodata."""
         ),
     ),
-    click.option("--sqlite-metadata/--no-sqlite-metadata", default=None),
+    pulp_option(
+        "--sqlite-metadata/--no-sqlite-metadata",
+        needs_plugins=[PluginRequirement("rpm", specifier="<=3.24.0")],
+        default=None,
+    ),
```
mdellweg, line 166: Can you please add a similar guard in the context object, so other consumers than the cli can benefit?

CHANGES/831.removal
```diff
@@ -0,0 +1,2 @@
+Removed "--sqlite-metadata/--no-sqlite-metadata" option from `pulp rpm repository update` command.
+It was phased out from the pulp_rpm plugin.
```
mdellweg, line 2: Can you update the changelog according to the change please?

mdellweg (commented):
This looks much better!
Thank you!

pulpcore/cli/rpm/repository.py
pedro-psb: Hey @mdellweg , I'm stuck with this.
What I understand from this is that it will prevent api-level calls to uses sqlite-metadata option in the payload for the desired version requirement. Then, it look like I need to create a `PulpRpmSqliteMetadataContext` (similar to [docs](https://docs.pulpproject.org/pulp_cli/architecture/#version-dependent-code-paths)) in "pulp_glue/rpm/context.py" and somehow add it in the `update` subcommand context, but this process is still very unclear to me. Could you help me out?

How can I relate the restriction defined in the EntityContext with the payload option? What should happend in the EntityContext when requirement is not met? How to/should I test that? And more broadly, what could be the those other consumers?

mdellweg: First pulp_glue is a library that was split out from the cli. Other consumers are pulp/squeezer and pulpcore (with the replicate feature).
Since this was an additional check for the same, you can even try to only have it over there (I try to get most of the needs_plugin logic handled inside glue because that's the common denominator).
The context class that is effected seems to be `PulpRpmRepositoryContext`, because `sqlite_metadata` is one of it's attributes. The method for such hooks is called `preprocess_entity`.
https://docs.pulpproject.org/pulp_cli/glue_reference/common_context/#pulp_glue.common.context.PulpEntityContext.preprocess_entity

CHANGES/831.removal
pedro-psb: How is that?
> Added a version-dependency guard of `pulp_rpm<=3.4.0` to the `--sqlite-metadata/--no-sqlite-metadata` option from `pulp rpm repository update`  command, as it was phased out from the plugin.

And this is still considered a ".removal", right?

mdellweg: Let's keep it a ".removal", as it's a breaking change for the user.
I would try to make it sound less technical:

> Marked option `--sqlite-metadata` on `pulp rpm repository update` unavailable for `pulp_rpm>=3.4`, as it was removed there.

Another (could be very technical again; it's a different audience) changelog should go into `CHANGES/pulp-glue` then.

pulpcore/cli/rpm/repository.py
```diff
@@ -159,7 +159,10 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
             on the repodata."""
         ),
     ),
-    click.option("--sqlite-metadata/--no-sqlite-metadata", default=None),
+    # version-guard lives in pulp-glue
```
mdellweg, line 162: I don't think that comment is bringing much value.

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -42,6 +42,14 @@ expect_succ pulp rpm repository update --repository "cli_test_rpm_repository" --
 expect_succ pulp rpm repository show --repository "cli_test_rpm_repository"
 test "$(echo "$OUTPUT" | jq -r '.description')" = "null"
 
+# sqlite metadata removal from pulp-cli (pulp/pulp_rpm#3328)
+if pulp debug has-plugin --name "rpm" --specifier ">3.24.0"
```
mdellweg, line 46:
```suggestion
if pulp debug has-plugin --name "rpm" --specifier ">=3.24.0"
```
I cannot imagine it will be removed in a z stream...

pulpcore/cli/rpm/repository.py
```diff
@@ -159,7 +159,10 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
...
+    click.option(
+        "--sqlite-metadata/--no-sqlite-metadata", default=None, help=_("Requires pulp_rpm<=3.24.0")
```
mdellweg, line 164:
```suggestion
        "--sqlite-metadata/--no-sqlite-metadata", default=None, help=_("Requires pulp_rpm<3.24.0")
```

CHANGES/831.removal
```diff
@@ -0,0 +1,2 @@
+Marked option `--sqlite-metadata` on `pulp rpm repository update/create` unavailable
+for `pulp_rpm>3.24.0`, as it was removed there.
```
mdellweg, line 2: No crazy linebreaks please.

pulp-glue/pulp_glue/rpm/context.py
```diff
@@ -281,6 +281,8 @@ def preprocess_entity(self, body: EntityDefinition, partial: bool = False) -> En
                     feature=_("customization of the config.repo file"),
                 )
             )
+        if "sqlite_metadata" in body:
+            self.pulp_ctx.needs_plugin(PluginRequirement("rpm", specifier="<=3.24.0"))
```
mdellweg, line 285:
```suggestion
            self.pulp_ctx.needs_plugin(PluginRequirement("rpm", specifier=">=3.24.0", inverted=True, feature=_("...<add somethig useful here>...")))
```

mdellweg (dismissed):
Please check the version specifiers. I cannot believe that this is removed in a z-release.

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
pedro-psb: Sorry, forget to talk about that. I still need to check when it will be released with Daniel and Ina, I was using it as a placeholder.

pedro-psb: @dralley and @ipanova , wdyt?

pulp-glue/pulp_glue/rpm/context.py
mdellweg: The wording of the error message changes from "needed" to "prevents" when using inverted. XD

(In the past, with `min_version` and `max_version` instead of `specifier` this had a subtle influence on the inclusion of the version at the bounds.)

pedro-psb: Hmm, I was wondering about the invert flag.
Does that error message work?
> Sqlite metadata support was removed from pulp_rpm in version `3.y.z`

mdellweg: if you are referring to "feature", then no. (That's why it's not called "error".) Please take a look at how the message is composed.

pulpcore/cli/rpm/repository.py
pedro-psb: Is there a use case where `pulp_option` could be preferred? I found this a bit confusing.
If there is, we could add a section in docs clarifying that.
And in any case, it would be helpfull to add a small example of using the `preprocess_entity` (as I've used here) for adding version constrains to options (maybe [here](https://docs.pulpproject.org/pulp_cli/architecture/#version-dependent-code-paths)).

pulp-glue/pulp_glue/rpm/context.py
pedro-psb: > The wording of the error message changes from "needed" to "prevents" when using inverted. 
I had misunderstood that. I'll use the API description then

pulpcore/cli/rpm/repository.py
mdellweg: > Is there a use case where `pulp_option` could be preferred? I found this a bit confusing. If there is, we could add a section in docs clarifying that.

`pulp_option` provides a superset of features over `click.option`. Since this project, as many others, grew organically, it is not quite used consistently in all places. To use things like `needs_plugin`, a `pulp_option` is absolutely necessary. I'm still on the fence whether we should recommend using it in (almost; `--config` is an exception for some reason as far as i remember) all places.

> And in any case, it would be helpfull to add a small example of using the `preprocess_entity` (as I've used here) for adding version constrains to options (maybe [here](https://docs.pulpproject.org/pulp_cli/architecture/#version-dependent-code-paths)).

Don't make me stop you. XD

pulp-glue/pulp_glue/rpm/context.py
mdellweg: I urge you to really take a look at how this string snippet is embedded into a sentence. The CLI is all about convenience and UX. A cryptic error message is not helping on that goal.

pulpcore/cli/rpm/repository.py
mdellweg: OK, sorry, `needs_plugin` does not work with `click.option`, but it still works when the layer below is raising the right exception (Hooray to stack unwinding exception handling!).
Declaring the need at the option level was what we were doing before realizing that pulp_glue had it's independent value as a standalone library. For the sake of the cli it almost does not matter where that exception is raised. But for uses of glue outside of the cli, it does.
I currently believe we should try as best as we can to move (or in your case, add) all version dependency in the glue library so all associated projects can benefit. This also makes the glue library the comprehensive guide to all known Pulp API subtleties.
The question of `pulp_option` over `click.option` is therefor in many cases one of best practice and least surprise. I think i'll vote for "use `pulp_option`" because this make `pulpcore.cli.common.generic` the library to the pulp-cli plugins and we have some control over subcommands that we loose when they use `click.option` directly.

pulpcore/cli/rpm/repository.py
dralley: <3.25

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
mdellweg: And 3.25 is certain? Is it merged in pulp_rpm already?

dralley: It's not merged but it will be shortly

pulp-glue/pulp_glue/rpm/context.py
mdellweg: https://github.com/pulp/pulp-cli/blob/main/pulp-glue/pulp_glue/common/context.py#L531

pedro-psb:
I've also added the guard to `PulpRpmPublicationContext` because I've found the option it also exposed in `pulp_rpm` publication API. Even though it's not exposed in the CLI, I though it may be more consistent, as pulp-glue can be used by other means.

mdellweg:
> I've also added the guard to `PulpRpmPublicationContext` because I've found the option it also exposed in `pulp_rpm` publication API. Even though it's not exposed in the CLI, I though it may be more consistent, as pulp-glue can be used by other means.

Sounds like we may want to add it there too. (Probably a new issue and a new PR though.)

pulpcore/cli/rpm/repository.py
pedro-psb: Thanks for the extra clarification. I'll open another issue for this.

pedro-psb:
> > I've also added the guard to `PulpRpmPublicationContext` because I've found the option it also exposed in `pulp_rpm` publication API. Even though it's not exposed in the CLI, I though it may be more consistent, as pulp-glue can be used by other means.
> 
> Sounds like we may want to add it there too. (Probably a new issue and a new PR though.)

Is it necessary open another issue? The reason is basically the same: the drop of support for this option. These are just two different consequences.

pulpcore/cli/rpm/repository.py
mdellweg: Please don't hesitate to capture any of this in the docs.

mdellweg:
> > > I've also added the guard to `PulpRpmPublicationContext` because I've found the option it also exposed in `pulp_rpm` publication API. Even though it's not exposed in the CLI, I though it may be more consistent, as pulp-glue can be used by other means.
> > 
> > 
> > Sounds like we may want to add it there too. (Probably a new issue and a new PR though.)
> 
> Is it necessary open another issue? The reason is basically the same: the drop of support for this option. These are just two different consequences.

I may have been ambiguous here: I thought about adding the option to the cli in the other place too.
At the very least it needs a changelog (Could be an orphaned one like "+<some_name>.feature").
If you feel like it i'm happy to add that with the same pr. Separate the commits though.

pulpcore/cli/rpm/repository.py
```diff
@@ -159,7 +159,14 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
...
         ),
     ),
-    click.option("--sqlite-metadata/--no-sqlite-metadata", default=None),
+    click.option(
+        "--sqlite-metadata/--no-sqlite-metadata",
+        default=None,
+        help=_(
+            """DEPRECATED: An option specifying whether Pulp should generate SQLite metadata.
+            Unavailable for pulp_rpm>=3.25.0"""
+        ),
```
mdellweg, line 168: +1 to help text. :)

mdellweg (commented):
I'd still prefer to wait until the corresponding change is merged upstream.
At that point (actually one nightly later), the tests should start to exercise this case in the container built from sources.

dralley:
@mdellweg This is now merged upstream https://github.com/pulp/pulp_rpm/pull/3328

dralley:
>I may have been ambiguous here: I thought about adding the option to the cli in the other place too

@mdellweg I'm not sure this is worthwhile, we're removing this option because it has little practical purpose and the functionality was nigh-untestable from a functional standpoint.  Katello never used it, so in practice it mostly seems to have been present because Pulp 2 had the functionality rather than because Pulp 3 needed it.  Also adding it in another place might encourage it to be used, which is of course something we don't want.

mdellweg (approved):
I think, that's it then.

dralley:
@mdellweg I think there's a required check that no longer exists as a GHA workflow?

mdellweg:
The check is still there (and it absolutely should be). And I see the workflow run somwhere. But somehow it's no longer associated with the PR. Can you push once more to see the tests running.

pedro-psb:
Inverting the `specifier>="3.25.0"` to `specifier<"3.25.0"` (and inverting the assertions order accordingly) made the test pass. Semantically, both are exactly the same, so can I report that as a bug?

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -42,6 +42,14 @@ expect_succ pulp rpm repository update --repository "cli_test_rpm_repository" --
 expect_succ pulp rpm repository show --repository "cli_test_rpm_repository"
 test "$(echo "$OUTPUT" | jq -r '.description')" = "null"
 
+# sqlite metadata removal from pulp_rpm (pulp/pulp_rpm#3328)
+if pulp debug has-plugin --name "rpm" --specifier "<3.25.0"
+then
+  expect_succ pulp rpm repository update --name "cli_test_rpm_repository" --no-sqlite-metadata
+else
+  expect_fail pulp rpm repository update --name "cli_test_rpm_repository" --no-sqlite-metadata
```
dralley, line 50: This is the same command on both branches, is that intentional?  I wouldn't expect that it can both succeed and fail.

pedro-psb: It succeeds if one is using `pulp_rpm<3.25.0` (which still support sqlite-metadata option) but fails otherwise.

dralley: I misread it, whoops.  My bad

mdellweg:
> Inverting the `specifier>="3.25.0"` to `specifier<"3.25.0"` (and inverting the assertions order accordingly) made the test pass. Semantically, both are exactly the same, so can I report that as a bug?

That sounds strange. Do you have a verified reproducer? Maybe it's a little bit ambiguous when you provide dev versions?
Like "3.25.0.dev.0" is still "<3.25.0" but contains the change already?

mdellweg (approved):

pedro-psb:
I'll try to make a reproducible setup for this.
But how do these test work, do they use actual repo checkouts for the plugins? Because rpm changes are on main, but there is no `3.25.0.dev` branch or anything.

mdellweg:
We build a container with plugins from source the test case here is called nightly.

PR #833, opened 2023-11-23
gpgcheck integer instead of choice

### LINKED ISSUE #677: rpm 'gpgcheck', 'repo_gpgcheck' is expected to be an integer.
## Summary
When running `pulp rpm repository create` or `pulp rpm repository update` with `--gpgcheck` or `--repo-gpgcheck`, I get the error below.

## Steps to reproduce
```console
$ pulp rpm repository update --name yum-docker-el7 --remote yum-docker-el7 --gpgcheck 1 --repo-gpgcheck 1
Error: Validation failed for 'repositories_rpm_rpm_partial_update':
  application/json: 'gpgcheck' is expected to be an integer.
  application/x-www-form-urlencoded: 'gpgcheck' is expected to be an integer.
  multipart/form-data: 'gpgcheck' is expected to be an integer.
```

## Expected behavior
Action to complete successfully.

pulp-ansible                    0.16.1
pulp-cli                        0.18.0
pulp-cli-deb                    0.0.4
pulp-container                  2.14.3
pulp-deb                        2.20.1
pulp-file                       1.12.0
pulp-glue                       0.18.0
pulp-python                     3.8.0
pulp-rpm                        3.19.1
pulpcore                        3.22.3

## Additonal context
Setting the click.option for gpgcheck and repo-gpgcheck to type int seems to fix, but I'm not sure that's the right answer.
https://github.com/pulp/pulp-cli/blob/2cc6dc424c3358ce456f21cd238c2fdadc30d834/pulpcore/cli/rpm/repository.py#L134
```
update_options = [
...
  click.option("--gpgcheck", type=int),
  click.option("--repo-gpgcheck", type=int),
...
]
```

MichalPysik:
Arguments --gpgcheck and --repo-gpgcheck in creating a rpm repository no longer fail to convert to integer.

closes #677

### DIFF
--- a/CHANGES/677.bugfix
+++ b/CHANGES/677.bugfix
@@ -1,0 +1,1 @@
+Arguments --gpgcheck and --repo-gpgcheck in creating a rpm repository no longer fail to convert to integer.
--- a/pulpcore/cli/rpm/repository.py
+++ b/pulpcore/cli/rpm/repository.py
@@ -82,6 +82,12 @@ def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) ->
 CONTENT_LIST_SCHEMA = s.Schema([{"pulp_href": str}])
 
 
+def choice_to_int_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Optional[int]:
+    if value:
+        return int(value)
+    return None
+
+
 @pulp_group()
 @click.option(
     "-t",
@@ -146,6 +152,7 @@ update_options = [
     click.option(
         "--gpgcheck",
         type=click.Choice(("0", "1")),
+        callback=choice_to_int_callback,
         help=_(
             """DEPRECATED:Option specifying whether a client should perform a GPG signature check
             on packages."""
@@ -154,6 +161,7 @@ update_options = [
     click.option(
         "--repo-gpgcheck",
         type=click.Choice(("0", "1")),
+        callback=choice_to_int_callback,
         help=_(
             """DEPRECATED:Option specifying whether a client should perform a GPG signature check
             on the repodata."""
--- a/tests/scripts/pulp_rpm/gpgcheck_test.sh
+++ b/tests/scripts/pulp_rpm/gpgcheck_test.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# shellcheck source=tests/scripts/config.source
+. "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
+
+pulp debug has-plugin --name "rpm" --max-version "3.23.0" || exit 23
+
+cleanup () {
+  pulp rpm repository destroy --name "cli-repo-gpgcheck" || true
+}
+trap cleanup EXIT
+
+expect succ pulp rpm repository create --name "cli-repo-gpgcheck" --gpgcheck 1 --repo-gpgcheck 1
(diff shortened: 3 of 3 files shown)


MichalPysik:
I still think about keeping ```click.Choice``` and converting to int via some callback, which would lead to prettier ```--help``` text imho. Functionally, this solution still works tho.

MichalPysik:
@mdellweg Agreed, the underlying logic wants it to be an integer, but from the user perspective I don't like that at all. I will try the callback solution and keep it a ```click.Choice``` if possible, and also add a test case as both you and @lubosmj suggested.

pulpcore/cli/rpm/repository.py
```diff
@@ -146,6 +146,7 @@ def repository(ctx: click.Context, pulp_ctx: PulpCLIContext, repo_type: str) ->
     click.option(
         "--gpgcheck",
         type=click.Choice(("0", "1")),
+        callback=lambda x, y, z: int(z) if z else None,
```
mdellweg, line 149: Please use proper names.
Also a lambda may not be the right thing to use here.

CHANGES/677.bugfix
```diff
@@ -0,0 +1,2 @@
+Arguments --gpgcheck and --repo-gpgcheck in creating a rpm repository now require an
+integer instead of click.choice(0, 1).
```
mdellweg, line 2: This is not matching the change. Also no crazy line breaks please.

mdellweg: The commit message is similarly outdated.

tests/scripts/pulp_rpm/gpgcheck_test.sh
```diff
@@ -0,0 +1,14 @@
...
+
+pulp debug has-plugin --name "rpm" --max-version "3.23.0" || exit 23
+
+cleanup () {
+  pulp rpm repository destroy --name "cli-repo-gpgcheck" || true
+}
+trap cleanup EXIT
+
+expect succ pulp rpm repository create --name "cli-repo-gpgcheck" --gpgcheck 1 --repo-gpgcheck 1
+expect succ pulp rpm repository update --name "cli-repo-gpgcheck" --gpgcheck 0 --repo-gpgcheck 0
```
mdellweg, line 14: You should verify in a subsequent show that the value is 0.

mdellweg (changes requested):

pulpcore/cli/rpm/repository.py
```diff
@@ -82,6 +82,12 @@ def _content_callback(ctx: click.Context, param: click.Parameter, value: Any) ->
 CONTENT_LIST_SCHEMA = s.Schema([{"pulp_href": str}])
 
 
+def choice_to_int_callback(ctx: click.Context, param: click.Parameter, value: Any) -> Optional[int]:
+    if value:
+        return int(value)
+    return None
+
```
mdellweg, line 89: Thanks for adding the test!
I was sure there was a bug in here, because in python `0` evaluates to `False`. As your test demonstrates, this actually works. And i think i understood now that what actually is passed in here is `"0"`, a non empty string that  evaluates to `True` instead. I would have expected that we needed `if value is None` instead, and i'm still on the fence. But I think we can merge it that way.

mdellweg (approved):

PR #837, opened 2023-11-29
Improve Contribution docs about version guards and pulp-glue

### LINKED ISSUE #836: Improve info on pulp-glue and adding version-constrain to option (on Contributing)
## Summary

When working on the drop of #831 Matthias clarified some history and practical info which would be useful in the Contributing section:
- Introduction to pulp-glue and why it's important
- How-to add option version-constrains (on pulp-glue)
- Recommend `pulp_option` more generally for declaring options (over `click.option`)

## Examples

> pulp_option provides a superset of features over click.option. Since this project, as many others, grew organically, it is not quite used consistently in all places. To use things like needs_plugin, a pulp_option is absolutely necessary. I'm still on the fence whether we should recommend using it in (almost; --config is an exception for some reason as far as i remember) all places.
>
> OK, sorry, needs_plugin does not work with click.option, but it still works when the layer below is raising the right exception (Hooray to stack unwinding exception handling!).
>
> Declaring the need at the option level was what we were doing before realizing that pulp_glue had it's independent value as a standalone library. For the sake of the cli it almost does not matter where that exception is raised. But for uses of glue outside of the cli, it does.
>
> I currently believe we should try as best as we can to move (or in your case, add) all version dependency in the glue library so all associated projects can benefit. This also makes the glue library the comprehensive guide to all known Pulp API subtleties.
>
> The question of pulp_option over click.option is therefor in many cases one of best practice and least surprise. I think i'll vote for "use pulp_option" because this make pulpcore.cli.common.generic the library to the pulp-cli plugins and we have some control over subcommands that we loose when they use click.option directly.

pedro-psb:
Done in the 'Architecture' section of 'Contribution':
- Add 'Pulp Glue' section
- Add clarifications about version-guard option/body-params
- Update docstrings of objects that were referenced.
- Nest 'Context' and 'Version Dependent' section under 'Pulp Glue'
- Fix typos

closes #836

Review Checklist:

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -6,7 +6,7 @@ The Pulp CLI architecture is described in this section.
 
 The Pulp CLI is designed with a plugin structure. Plugins can either live in the pulp-cli package or be shipped independently.
 By convention, all parts of the CLI are packages in the open namespace `pulpcore.cli`.
-A plugin can register itself with the main app by specifying by specifying its main module as a `pulp_cli.plugins` entrypoint in `setup.py`.
+A plugin can register itself with the main app by specifying its main module as a `pulp_cli.plugins` entrypoint in `setup.py`.
 
 ```python
 entry_points={
@@ -16,7 +16,7 @@ entry_points={
 }
 ```
 
-The plugin can then attach subcommands to the `pulpcore.cli.common.main` command by providing a `mount` method in the main module.
+The plugin should then attach subcommands to the `pulpcore.cli.common.main` command by providing a `mount` method in the main module.
 
 ```python
 from pulpcore.cli.common.generic import pulp_command
@@ -30,12 +30,22 @@ def mount(main: click.Group, **kwargs: Any) -> None:
     main.add_command(my_command)
 ```
 
-### Contexts
+### Pulp Glue
+
+Pulp CLI provides the `pulp-glue` library as an abstraction layer that let's you perform high level operations in pulp.
+Its goal is to abstract interacting with the rest api by parsing the api docs and waiting on tasks and task groups.
+It is shipped as a separate python package to allow broad use across multiple projects, such as `pulp-squeezer` and `pulpcore`.
+To this end, `pulp-glue` is the go to place for all known version dependent Pulp API subtleties and their corresponding fixes (see Version dependent codepaths below).
+
+#### Contexts
 
 In `click`, every subcommand is accompanied by a `click.Context`, and objects can be attached to them.
-In this CLI we attach a `PulpContext` to the main command.
-It acts as an abstraction layer that handles the communication to the pulp server through its `api` property.
-Further we encourage the handling of communication with certain endpoints by subclassing the `PulpEntityContext`.
+In this CLI we attach a [`PulpCLIContext`][pulpcore.cli.common.generic.PulpCLIContext] to the main command, which inherits from `pulp-glue`'s [`PulpContext`][pulp_glue.common.context.PulpContext].
+This context handles the communication to the pulp server through its `api` property.
+
+Further we encourage the handling of communication with certain endpoints by subclassing the [`PulpEntityContext`][pulp_glue.common.context.PulpEntityContext] or some of the resource-specific children, such as [PulpRepositoryContext][pulp_glue.common.context.PulpRepositoryContext].
+Some examples of this can be found under `pulp_glue/{plugin-name}/context.py`.
+
 By attaching them to the contexts of certain command groups, they are accessible to commands via the `pass_entity_context` decorator.
 Those entity contexts should provide a common interface to the layer of `click` commands that define the user interaction.
 
@@ -54,14 +64,16 @@ def my_sub_command(entity_ctx):
     entity_ctx.destroy(href)
 ```
 
-### Version dependent code paths
+#### Version dependent code paths
 
 Each Pulp CLI release is designed to support multiple Pulp server versions and the CLI itself is versioned independently of any version of the Pulp server components.
 It is supposed to be able to communicate with different combinations of server component versions at the same time.
 Because of this, it might be necessary to guard certain features and workarounds by checking against the available server plugin version.
-As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects to provide a consistent interface to the command callbacks.
-To facilitate diverting code paths depending on plugin versions, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods.
-While `has_plugin` will evaluete immediately, `needs_plugin` can be seen as a deferred assertion.
+
+As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects.
+To facilitate diverting code paths depending on plugin versions, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods, both of which accept a [`PluginRequirement`][pulp_glue.common.context.PluginRequirement] object to describe dependencies on server components.
+
+While `has_plugin` will evaluate immediately, `needs_plugin` can be seen as a deferred assertion.
 It will raise an error, once the first access to the server is attempted.
 
 ```python
@@ -82,17 +94,28 @@ def my_command(ctx, pulp_ctx):
     ctx.obj = MyEntityContext(pulp_ctx)
 ```
 
-The named tuple `PluginRequirement` is used to describe dependence on server components.
-It needs the `name` of the plugin and accepts optionally an inclusive `min` version,
-an exclusive `max` version, and an `inverted` flag for exclusion of a version range.
+To declare version restrictions on *options*, the [`preprocess_entity`][pulp_glue.common.context.PulpEntityContext.preprocess_entity] method can be used to check if a given option is present in the request body and conditionally apply the requirements to the context.
 
-Additionally, the `PulpOption` provides the `needs_plugins` keyword argument.
-It accepts a list of `PluginRequirements` to error when used.
+In the following example, a guard is added because `my_option` was introduced to `MyPluginRepository` in version 3.24.0 of `"my_plugin"`:
+
+```python
+class PulpMyPluginRepositoryContext(PulpRepositoryContext):
+    ...
+
+    def preprocess_entity(self, body, partial) -> EntityDefinition:
+        body = super().preprocess_entity(body, partial=partial)
+        if "my_option" in body:
+            self.pulp_ctx.needs_plugin(
+                PluginRequirement("my_plugin", specifier=">=3.24.0", feature=_("my feature"))
+            )
+        return body
+```
 
 ### Generics
 
 For certain often repeated patterns like listing all entities of a particular kind,
 we provide generic commands that use the underlying context objects.
+The following example shows the use of the [`show_command`][pulpcore.cli.common.generic.show_command] generic.
 
 ```python
 from pulpcore.cli.common.generic import name_option, show_command,
@@ -101,15 +124,15 @@ lookup_params = [name_option]
 my_command.add_command(show_command(decorators=lookup_params))
 ```
 
-Some of these commands understand extra arguments.
-They can be attached by passing a list of `click.Options` via the `decorators` argument.
+To add options to these subcommands, pass a list of [`PulpOption`][pulpcore.cli.common.generic.PulpOption] objects to the `decorators` argument.
+Preferably these are created using the [`pulp_option`][pulpcore.cli.common.generic.pulp_option] factory.
 
 ```python
 from pulpcore.cli.common.generic import list_command,
 
 filter_params = [
-    click.option("--name"),
-    click.option("--name-contains", "name__contains"),
+    pulp_option("--name"),
+    pulp_option("--name-contains", "name__contains"),
 ]
 my_command.add_command(list_command(decorators=filter_params))
 ```
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -296,6 +296,13 @@ def pulp_group(
 
 
 class PulpOption(click.Option):
+    """
+    Pulp-CLI specific subclass of `click.Option`.
+
+    The preferred way to use this is through the
+    [`pulp_option`][pulpcore.cli.common.generic.pulp_option] factory.
+    """
+
     def __init__(
         self,
         *args: t.Any,
@@ -551,7 +558,26 @@ def null_callback(
 
 
 def pulp_option(*args: t.Any, **kwargs: t.Any) -> t.Callable[[FC], FC]:
-    """Factory that creates a `PulpOption`."""
+    """
+    Factory of [`PulpOption`][pulpcore.cli.common.generic.PulpOption] objects.
+
+    `PulpOption` provides extra features over `click.Option`, namely:
+
+    1. Define version constrains.
+    2. Support for template variables in the help message.
+    3. Limit the use of options to certain entity contexts.
+
+    Examples:
+        Define version constrains and custom help message:
+        ```
+        pulp_option(
+            "--name",
+            needs_plugins=[PluginRequirement("rpm", specifier=">=3.12.0")],
+            help=_("Name of {entity}"),
+            allowed_with_contexts=(PulpRpmRepositoryContext,),
+        )
+        ```
+    """
     kwargs["cls"] = PulpOption
     return click.option(*args, **kwargs)
 
--- a/CHANGES/836.doc
+++ b/CHANGES/836.doc
@@ -1,0 +1,1 @@
+Improved `Developer Material/Architecture` section by adding clarification about `pulp-glue`, version guards and `pulp_option` factory.


docs/architecture.md
```diff
@@ -30,12 +30,21 @@ def mount(main: click.Group, **kwargs: Any) -> None:
     main.add_command(my_command)
 ```
 
-### Contexts
+### Pulp Glue
+
+Pulp CLI uses `pulp-glue` library to manage plugin versions dependent features. Historically, those restrictions were
+defined in the CLI itself at the *option level*, but because there were other consumers interested in that (such as `pulp-smash` and `pulpcore`)
+we extracted this module as an independent library. Thus, `pulp-glue` aims to be the comprehensive guide to all known Pulp API subtleties.
```
mdellweg, line 37: Please use one line per sentence. Everything else makes reading diffs hard.

docs/architecture.md
```diff
@@ -30,12 +30,21 @@ def mount(main: click.Group, **kwargs: Any) -> None:
     main.add_command(my_command)
 ```
 
-### Contexts
+### Pulp Glue
+
+Pulp CLI uses `pulp-glue` library to manage plugin versions dependent features. Historically, those restrictions were
```
mdellweg, line 35: This is not the reason, we introduced glue. Glue is an abstraction layer that let's you perform high level operations in pulp (like repository.update(...)). It abstracts talking to the rest api including parsing the api.json, waiting on tasks and task groups. And it allows multiple projects to use this level of abstraction.
And from that architectural viewpoint it's natural to put all the version dependent fixes there too.

Historically there was no glue and therefore not a good separation of concerns. But i doubt, that this information is really helpful.

"Thus, `pulp-glue` aims to be the comprehensive guide to all known Pulp API subtleties." This is not the selling point for pulp-glue, but rather a corollary.

docs/architecture.md
```diff
@@ -6,7 +6,7 @@ The Pulp CLI architecture is described in this section.
 
 The Pulp CLI is designed with a plugin structure. Plugins can either live in the pulp-cli package or be shipped independently.
 By convention, all parts of the CLI are packages in the open namespace `pulpcore.cli`.
-A plugin can register itself with the main app by specifying by specifying its main module as a `pulp_cli.plugins` entrypoint in `setup.py`.
+A plugin can register itself with the main app by specifying its main module as a `pulp_cli.plugins` entrypoint in `setup.py`.
```
mdellweg, line 9: Good catch, but we should start mentioning the `mount` command.

docs/architecture.md
```diff
@@ -54,14 +63,17 @@ def my_sub_command(entity_ctx):
...
+#### Version dependent code paths
 
 Each Pulp CLI release is designed to support multiple Pulp server versions and the CLI itself is versioned independently of any version of the Pulp server components.
 It is supposed to be able to communicate with different combinations of server component versions at the same time.
 Because of this, it might be necessary to guard certain features and workarounds by checking against the available server plugin version.
+
 As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects to provide a consistent interface to the command callbacks.
 To facilitate diverting code paths depending on plugin versions, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods.
-While `has_plugin` will evaluete immediately, `needs_plugin` can be seen as a deferred assertion.
+Then, the named tuple [`PluginRequirement`][pulp_glue.common.context.PluginRequirement] is used to describe dependence on server components.
```
mdellweg, line 74: This is not a named tuple.

mdellweg: ... both of which accept a `PluginRequirement` object to describe ...
or
Dependencies on server components are passed around as `PluginRequirement` objects.

docs/architecture.md
```diff
@@ -82,17 +94,32 @@ def my_command(ctx, pulp_ctx):
     ctx.obj = MyEntityContext(pulp_ctx)
 ```
 
-The named tuple `PluginRequirement` is used to describe dependence on server components.
-It needs the `name` of the plugin and accepts optionally an inclusive `min` version,
-an exclusive `max` version, and an `inverted` flag for exclusion of a version range.
+To declare version restrictions on *options* (*body parameters* in the API), the [`preprocess_entity`][pulp_glue.common.context.PulpEntityContext.preprocess_entity]
+method can check if a given option is being used and conditionally apply the requirements to the context.
```
mdellweg, line 98: This function is intended more generally. We also fix some inputs there or perform transformations on certain fields.

docs/architecture.md
```diff
@@ -82,17 +94,32 @@ def my_command(ctx, pulp_ctx):
...
+
+Each *plugin* may already have `Contexts` binded to specific [resource_class](/using_the_cli/#general-command-syntax) entities,
+which can be found under `pulp_glue/{plugin-name}/context.py`. In the example, a guard is added to `my_option`, which belongs to the
+`Repository` *resource_class* of `"my_plugin"`.
```
mdellweg, line 102: This is a bit unclear to me. Whenever you are talking to an entity, there should most definitely be a context for it.

docs/architecture.md
```diff
@@ -82,17 +94,32 @@ def my_command(ctx, pulp_ctx):
...
 
-Additionally, the `PulpOption` provides the `needs_plugins` keyword argument.
-It accepts a list of `PluginRequirements` to error when used.
```
mdellweg, line 90: Where's this gone?

docs/architecture.md
```diff
@@ -102,14 +129,15 @@ my_command.add_command(show_command(decorators=lookup_params))
 ```
 
 Some of these commands understand extra arguments.
-They can be attached by passing a list of `click.Options` via the `decorators` argument.
+They can be attached by passing a list of [`pulp_option`][pulpcore.cli.common.generic.pulp_option] factories via the `decorators` argument.
+These factories act as `click.Option` objects, but provides extra functionality specific to this CLI and thus are preferred over the latter.
```
mdellweg, line 133: This is not correct. You don't pass the factories, but the `Option`s. And the decorator accepts `click.Option`, while `PulpOption` is a subclass thereof.

pulpcore/cli/common/generic.py
```diff
@@ -296,6 +296,12 @@ def pulp_group(
 
 
 class PulpOption(click.Option):
+    """
+    Wrapper class for `click.Option`.
+
+    This provides a standard way of defining CLI options and helps keep consistent behavior.
```
mdellweg, line 302: We want to encourage using the factory, so please mention that here.
Also this is a great place to tell, what `PulpOption` provides on top of `click.Option`.

mdellweg (changes requested):

docs/architecture.md
pedro-psb: > Pulp CLI uses `pulp-glue` library as an abstraction layer that let's you perform high level operations in pulp.
> It abstracts talking to the rest api (including parsing the api.json), waiting on tasks and task groups.
> Also, it allows multiple projects to use this level of abstraction, such as `pulp-smash` and `pulpcore`.
> For this, `pulp-glue` is becoming the comprehensive guide to all known Pulp API subtleties, such as version dependent fixes.
>

docs/architecture.md
pedro-psb: You mean invert the order of the information, like first mention the `mount`  part and then the setup of the `entrypoint`?

docs/architecture.md
pedro-psb: The sentences says `preprocess_entity` can be used to achieve a task, but it doesn't imply that it's the only thing it can be used for (and the linked reference description also enforces its  general purpose) nature. Given the context ("Version dependent path" section), I think it's reasonable to just mention this one. Wdyt?

docs/architecture.md
pedro-psb: I see. The relation about entities and context was not very clear. How about (without the bold)
> Each *plugin*  **have** `Contexts` binded to specific [resource_class](/using_the_cli/#general-command-syntax) entities, which can be found under `pulp_glue/{plugin-name}/context.py`. (...)

docs/architecture.md
mdellweg: Sounds good too.
And we can always reiterate...

docs/architecture.md
mdellweg: better for sure.

docs/architecture.md
mdellweg: Oh, right. `mount` is already there. (It was invisible to the diff...)

Can you just replace "can" by "should" in line 19?

docs/architecture.md
pedro-psb: Sorry, I guess I'm in a loophole about this.

When I was asking you about pulp_option vs pulp_glue, I was trying to find out if there was a way of writing:
"use `needs_plugins` at `PulpOption` in this case or at `pulp_glue` in that other case". As version guard can (and generally should) be implemented in `pulp_glue`, I concluded duplicating  `needs_plugin` in `PulpOption` would always be redundant from the CLI perspective, since thats already covered up by pulp_glue.

But if you think it's worth it, I could move this to the "Generics" section, where I mention `pulp_option` (since this section is now a subsection of "Pulp Glue" heading).

docs/architecture.md
pedro-psb: Sorry, I've mixed that up.

mdellweg: It has been a named tuple in the past... Long ago.

docs/architecture.md
mdellweg: Oh, so you are saying: `needs_plugins` is redundant now that we try to move that all to `pulp_glue`. You have a point there. But I'm not sure if that is a transition, or whether we actually need that possibility in both places in the end. I cannot come up with a good reason though. Maybe deprecating these is another 1.0 goal. Let's keep it removed and see where we get.

pulpcore/cli/common/generic.py
```diff
@@ -299,7 +299,10 @@ class PulpOption(click.Option):
     """
     Wrapper class for `click.Option`.
 
-    This provides a standard way of defining CLI options and helps keep consistent behavior.
+    This is the standard way of defining CLI options. In addition to `click.Option` it provides:
+    - `needs_plugin`: specify version constrains for an option
+    
+    The preferred way to use this is through the [`pulp_option`](pulpcore.cli.common.generic.pulp_option) factory.
```
mdellweg, line 305: Can we keep this rather short and put the bulk of documentation with the `pulp_option`? Maybe "Pulp-CLI specific subclass of click.Option" and then the reference to pulp_option.

Also can you figure out what else besides `needs_plugin` this class provides?

docs/architecture.md
```diff
@@ -82,17 +92,30 @@ def my_command(ctx, pulp_ctx):
...
+```python
+class PulpMyPluginRepositoryContext(PulpRepositoryContext):
+    ...
+
+    def preprocess_entity(self, body, partial) -> EntityDefinition:
+        body = super().preprocess_entity(body, partial=partial)
+        if "my_option" in body:
+            self.pulp_ctx.needs_plugin(
+                PluginRequirement( "my_plugin", specifier=">=3.24.0", feature=_("my feature"),
+                )
```
mdellweg, line 109:
```suggestion
                PluginRequirement("my_plugin", specifier=">=3.24.0", feature=_("my feature"))
```

pulpcore/cli/common/generic.py
```diff
@@ -296,6 +296,15 @@ def pulp_group(
 
 
 class PulpOption(click.Option):
+    """
+    Wrapper class for `click.Option`.
```
mdellweg, line 300: A subclass is not a wrapper IMHO.

docs/architecture.md
```diff
@@ -94,12 +92,10 @@ def my_command(ctx, pulp_ctx):
...
 ```
 
-To declare version restrictions on *options* (*body parameters* in the API), the [`preprocess_entity`][pulp_glue.common.context.PulpEntityContext.preprocess_entity]
-method can check if a given option is being used and conditionally apply the requirements to the context.
+To declare version restrictions on *options*, the [`preprocess_entity`][pulp_glue.common.context.PulpEntityContext.preprocess_entity] method can be used to check if a given option is present in the request body and conditionally apply the requirements to the context.
 
-Each *plugin* may already have `Contexts` binded to specific [resource_class](/using_the_cli/#general-command-syntax) entities,
-which can be found under `pulp_glue/{plugin-name}/context.py`. In the example, a guard is added to `my_option`, which belongs to the
-`Repository` *resource_class* of `"my_plugin"`.
+Each *plugin* have `Contexts` binded to specific [resource_class](using_the_cli.md/#general-command-syntax) entities, which can be found under `pulp_glue/{plugin-name}/context.py`.
```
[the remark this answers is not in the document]
pedro-psb: I would say this is a related but a complementary information:
-  line 46 is mostly useful for adding a new plugin, where it will be necessary to subclass `PulpEntityContext` and setup the context.
- this line is mostly useful for someone (like me) who wants to modify an existing Context and won't do a deep dive into plumbing. It maybe implicit from the architecture that existing plugins already have Context bound to resources, and it's possible to guess that lives it there, but IMHO its not so obvious for that audience.

Maybe it would be appropriate to move it there:
> Further we encourage the handling of communication with certain endpoints by subclassing the `PulpEntityContext`.
> Existing plugins already have *Contexts* bound to specific *resource classes*, which can be found under `pulp_glue/{plugin-name}/context.py`.

docs/architecture.md
```diff
@@ -94,12 +92,10 @@ def my_command(ctx, pulp_ctx):
...
+In the example, a guard is added to `my_option`, which belongs to the `Repository` *resource_class* of `"my_plugin"`.
```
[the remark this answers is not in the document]
pedro-psb: Thats better. Can I just replace the second *added* to avoid repeating it?
> because `my_option` was ~added~ **introduced** `MyPluginRepository`

pulpcore/cli/common/generic.py
pedro-psb: Sure, will do.

Hmm, it seems to provide basic templating to the help msg, so it can access the entity name that it belongs to with curly braces.

mdellweg: Yes! It does. And there's another mechanism that allows to limit the use of some options to certain entity contexts. That one is probably the most complicated to explain.

docs/architecture.md
```diff
@@ -30,12 +30,20 @@ def mount(main: click.Group, **kwargs: Any) -> None:
...
+For this, `pulp-glue` is becoming the comprehensive guide to all known Pulp API subtleties, such as version dependent fixes.
+
+#### Contexts
 
 In `click`, every subcommand is accompanied by a `click.Context`, and objects can be attached to them.
-In this CLI we attach a `PulpContext` to the main command.
-It acts as an abstraction layer that handles the communication to the pulp server through its `api` property.
-Further we encourage the handling of communication with certain endpoints by subclassing the `PulpEntityContext`.
+In this CLI we attach a [`PulpContext`][pulp_glue.common.context.PulpContext] to the main command.
+This context object lives in `pulp-glue` and handles the communication to the pulp server through its `api` property.
```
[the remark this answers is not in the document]
pedro-psb: :open_mouth:

docs/architecture.md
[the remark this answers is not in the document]
mdellweg: It's always hard to make any definite claim about coverage in the cli. You see, we add what we need when we need it. This will probably never be complete.

docs/architecture.md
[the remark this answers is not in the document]
mdellweg: Sounds way better.

pulpcore/cli/common/generic.py
pedro-psb: I've added an example in the docstring to provide more straighfoward way of understanding `PulpOption` features. Could you help adding one about that limiting-use-of-context feature?

docs/architecture.md
```diff
@@ -54,14 +64,16 @@ def my_sub_command(entity_ctx):
...
+#### Version dependent code paths
 
 Each Pulp CLI release is designed to support multiple Pulp server versions and the CLI itself is versioned independently of any version of the Pulp server components.
 It is supposed to be able to communicate with different combinations of server component versions at the same time.
 Because of this, it might be necessary to guard certain features and workarounds by checking against the available server plugin version.
+
 As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects to provide a consistent interface to the command callbacks.
-To facilitate diverting code paths depending on plugin versions, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods.
-While `has_plugin` will evaluete immediately, `needs_plugin` can be seen as a deferred assertion.
+To facilitate diverting code paths depending on plugin versions, the `PulpContext` provides the `needs_plugin` and `has_plugin` methods, both of which accept a [`PluginRequirement`]() object to describe dependencies on server components.
```
mdellweg, line 74: Is the PluginRequirement here supposed to be a link? Do we need to define it, or is there some magic to hook it up automatically?

pulpcore/cli/common/generic.py
```diff
@@ -551,7 +557,24 @@ def null_callback(
...
+    2. Support for template variables in the help message.
+    3. Limit the use of options to certain entity contexts.
+
+    Examples:
+        Define version constrains and custom help message:
+        ```
+        pulp_option(
+            "--name", 
+            needs_plugins=[PluginRequirement("rpm", specifier=">=3.12.0")],
+            help=_("Name of {entity}"),
```
mdellweg, line 574: Let's add `allowed_with_contexts` to the example.

docs/architecture.md
```diff
@@ -54,14 +64,16 @@ def my_sub_command(entity_ctx):
...
 ```
 
-### Version dependent code paths
+#### Version dependent code paths
 
 Each Pulp CLI release is designed to support multiple Pulp server versions and the CLI itself is versioned independently of any version of the Pulp server components.
 It is supposed to be able to communicate with different combinations of server component versions at the same time.
 Because of this, it might be necessary to guard certain features and workarounds by checking against the available server plugin version.
+
 As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects to provide a consistent interface to the command callbacks.
```
mdellweg, line 73:
```suggestion
As a rule of thumb, all necessary workarounds should be implemented in the corresponding `Context` objects.
```

We have a whole paragraph for this reason now. ;)

mdellweg (commented):
Probably the last round of asks.

docs/architecture.md
pedro-psb: I've missed that, we have to define.
Later I'll do some research on linters for documentation, they do a better job than us for certain things,

mdellweg:
Please squash the commits for merging (after running black...).

mdellweg (approved):

Issue #842, opened 2023-12-05
CLI does not send the file name in the form data on file uploads.


gerrod3:
## Summary
When serializing the request body for an upload, openapi.py does not specify the file name in the upload tuple `(filename, file_data, content_type)`

## Steps to reproduce
Perform a file upload with `-vvv` and look at the filename for the file field, it'll probably be 'file' (the name of the upload file used by the api)

```
(pulp) [gerrod@localhost pulp_ansible]$ pulp -vvv ansible content upload --file pulp-squeezer-0.0.9.tar.gz
upload_collection : post http://localhost:5001/ansible/collections/
  User-Agent: Pulp-CLI/0.22.0.dev
  Accept-Encoding: gzip, deflate
  Accept: application/json
  Connection: keep-alive
  Content-Length: 36993
  Content-Type: multipart/form-data; boundary=94544652e2e7dcd9cf857ec97c99bb6d
  Authorization: Basic YWRtaW46cGFzc3dvcmQ=
b'--94544652e2e7dcd9cf857ec97c99bb6d\r\nContent-Disposition: form-data; name="file"; filename="file"
```

mdellweg:
What is the actual bug here? What does not work when doing it this way?

mdellweg:
Sorry, wrong button.

gerrod3:
Endpoints that parse the filename of the uploaded file like we do in ansible: https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/serializers.py#L483-L485

mdellweg:
And how come we didn't see this earlier?

gerrod3:
We haven't seen this earlier because I set up ansible collection upload to use the endpoint with the least validation, which I happen to deprecate in pulp_ansible 0.16  (https://github.com/pulp/pulp_ansible/pull/1176). I'm going to submit a PR to switch collection uploads to use the pulp v3 api instead.

PR #843, opened 2023-12-05
Fix filename form data on uploads

gerrod3:
fixes: #842

Review Checklist:

### DIFF
--- a/pulp-glue/pulp_glue/common/openapi.py
+++ b/pulp-glue/pulp_glue/common/openapi.py
@@ -504,20 +504,21 @@ class OpenAPI:
                 elif content_type.startswith("application/x-www-form-urlencoded"):
                     data = body
                 elif content_type.startswith("multipart/form-data"):
-                    uploads: Dict[str, UploadType] = {}
+                    uploads: Dict[str, Tuple[str, UploadType, str]] = {}
                     data = {}
                     # Extract and prepare the files to upload
                     if body:
                         for key, value in body.items():
                             if isinstance(value, (bytes, BufferedReader)):
-                                uploads[key] = value
+                                uploads[key] = (
+                                    getattr(value, "name", key),
+                                    value,
+                                    "application/octet-stream",
+                                )
                             else:
                                 data[key] = value
                     if uploads:
-                        files = [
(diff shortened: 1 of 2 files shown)


pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -504,20 +504,19 @@ def render_request_body(
...
                     data = {}
                     # Extract and prepare the files to upload
                     if body:
                         for key, value in body.items():
-                            if isinstance(value, (bytes, BufferedReader)):
-                                uploads[key] = value
+                            if isinstance(value, bytes):
+                                uploads[key] = (key, value, "application/octet-stream")
+                            elif isinstance(value, BufferedReader):
+                                uploads[key] = (value.name, value, "application/octet-stream")
```
mdellweg, line 515: I don't fine the name attribute in the docs for BufferedReader (i'm also not sure this is the best way to determine we are transferring a file.) How about we do
```suggestion
                                uploads[key] = (getattr(value, "name", key), value, "application/octet-stream")
```
instead of the new if and then look if we can use the openapi-spec to see whether a file is supposed to be transferred?

gerrod3: What do you mean by check the api-spec to see if a file is supposed to be transferred?

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -504,20 +504,21 @@ def render_request_body(
...
-                    uploads: Dict[str, UploadType] = {}
+                    uploads: Dict[str, Tuple[str, UploadType, str]] = {}
                     data = {}
                     # Extract and prepare the files to upload
                     if body:
                         for key, value in body.items():
                             if isinstance(value, (bytes, BufferedReader)):
-                                uploads[key] = value
+                                uploads[key] = (
+                                    getattr(value, "name", key),
```
mdellweg, line 514: I believe the python bindings send files with the key as the filename also. But let's do it this way now.
(Actually I always assumed, DRF or at least Pulp would ignore the filename also.)

mdellweg (approved):

PR #847, opened 2023-12-07
Repository version user-entered options order fix

### LINKED ISSUE #650: It looks like option order can matter in unintuitive ways.
## Summary
While using Pulp CLI, I encountered the following:
```
$ oci-env pulp deb repository version destroy --version 1 --repository=${ENTITIES_NAME}
Error: A apt repository must be specified for this command.
$ oci-env pulp deb repository version destroy --repository=${ENTITIES_NAME} --version 1
Started background task /pulp/api/v3/tasks/51676f37-ea1f-4c97-b0e8-031f93577892/
.Done.
```

## Steps to reproduce

1. Create some repository, with some content in it.
2. Delete a repository version using the `--version` and `--repository` flags in that order
3. You get e.g.: `Error: A apt repository must be specified for this command.`

## Expected behavior
The order the `--version` and `--repository` flags are given to a CLI command in should not matter.

```
# pulp --version
pulp3 command line interface, version 0.16.0
```

```
  "versions": [
    {
      "component": "core",
      "version": "3.22.0.dev",
      "package": "pulpcore"
    },
    {
      "component": "deb",
      "version": "2.21.0.dev",
      "package": "pulp_deb"
    },
    {
      "component": "file",
      "version": "1.12.0.dev",
      "package": "pulp-file"
    }
  ],
```

## Additonal context

I suspect this is nothing to do with `pulp-cli-deb` and will happen equally for other content plugins, but I did not test this.

MichalPysik:
The order in which the user enters options (--version, --repository, ..) no longer matters for
'pulp <plugin> repository version' commands, since all API calls are now deferred to after the
options' callbacks are processed.

closes #650

Review Checklist:

### DIFF
--- a/CHANGES/650.bugfix
+++ b/CHANGES/650.bugfix
@@ -1,0 +1,1 @@
+User-entered order of parameters no longer matters for repository version commands.
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -414,14 +414,9 @@ def href_callback(
 def _version_callback(
     ctx: click.Context, param: click.Parameter, value: t.Optional[int]
 ) -> t.Optional[int]:
-    entity_ctx = ctx.find_object(PulpEntityContext)
-    assert entity_ctx is not None
-    repository_ctx = ctx.find_object(PulpRepositoryContext)
-    assert repository_ctx is not None
-    if value is not None:
-        entity_ctx.pulp_href = f"{repository_ctx.entity['versions_href']}{value}/"
-    else:
-        entity_ctx.pulp_href = repository_ctx.entity["latest_version_href"]
+    repository_version_ctx = ctx.find_object(PulpRepositoryVersionContext)
+    assert repository_version_ctx is not None
+    repository_version_ctx.entity = {"number": value}
     return value
 
 
--- a/pulp-glue/pulp_glue/common/context.py
+++ b/pulp-glue/pulp_glue/common/context.py
@@ -675,9 +675,8 @@ class PulpEntityContext:
             self.pulp_ctx.needs_plugin(plugin_requirement)
 
         self._entity = None
-        if pulp_href is None:
-            self._entity_lookup = entity or {}
-        else:
+        self._entity_lookup = entity or {}
+        if pulp_href is not None:
             self.pulp_href = pulp_href
 
     def call(
@@ -1159,6 +1158,21 @@ class PulpRepositoryVersionContext(PulpEntityContext):
             return {}
         else:
             return {self.repository_ctx.HREF: self.repository_ctx.pulp_href}
+
+    @property
+    def entity(self) -> EntityDefinition:
+        if (
+            self._entity is None
+            and "number" in self._entity_lookup.keys()
+            and self._entity_lookup.get("number") is None
+        ):
+            self.pulp_href = self.repository_ctx.entity["latest_version_href"]
+        return super().entity
+
+    @entity.setter
+    def entity(self, value: Optional[EntityDefinition]) -> None:
+        # ignore needed due to a bug regarding overriding property setter in mypy
+        PulpEntityContext.entity.fset(self, value)  # type: ignore[attr-defined]
 
     def repair(self, href: Optional[str] = None) -> Any:
         """
--- a/tests/scripts/pulp_file/test_sync.sh
+++ b/tests/scripts/pulp_file/test_sync.sh
@@ -46,7 +46,7 @@ expect_succ pulp file repository version repair --repository "cli_test_file_repo
 test "$(echo "$OUTPUT" | jq -r '.state')" = "completed"
 
 # Delete version again
-expect_succ pulp file repository version destroy --repository "cli_test_file_repository" --version 1
+expect_succ pulp file repository version destroy --version 1 --repository "cli_test_file_repository"
 
 # Test autopublish
 expect_succ pulp file repository create --name "$autopublish_repo" --remote "cli_test_file_remote" --autopublish
--- a/tests/test_help_pages.py
+++ b/tests/test_help_pages.py
@@ -1,4 +1,5 @@
 import typing as t
+from functools import reduce
 
 import click
 import pytest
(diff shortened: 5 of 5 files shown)


MichalPysik:
I don't know whether that's needed, but if you want me to add a simple test case, I can insert a ```pulp {plugin} repository version show``` somewhere in the existing tests.

lubosmj:
You may sneak a `repository version destroy` call at the end of an existing test, like: `pulp file repository version destroy --version 1 --repository=repo` and change the order of some of the used resource lookups, like: https://github.com/pulp/pulp-cli/blob/3183a2a9795a2091f106b0c414c87d649a8347a5/tests/scripts/pulp_rpm/test_rpm_sync_publish.sh#L101, https://github.com/pulp/pulp-cli/blob/3183a2a9795a2091f106b0c414c87d649a8347a5/tests/scripts/pulp_container/test_copy.sh#L32.

With this, we will verify whether your fix works globally or not.

lubosmj (approved):

pulpcore/cli/common/generic.py
```diff
@@ -863,11 +863,13 @@ def _type_callback(ctx: click.Context, param: click.Parameter, value: t.Optional
     help=_("Name of the repository"),
     callback=lookup_callback("name", PulpRepositoryContext),
     expose_value=False,
+    is_eager=True,
```
mdellweg, line 866: I'm not convinced, this is a nice solution, because it put's `--repository` on the same level as `--help`.
I believe we have ways to specify repository versions in a single option with the export code.

lubosmj: Would that mean that the help message will not be printed if I use both `--repository` and `--help`?

mdellweg: That's sure something to test. But I'm hesitant to call "--repository" _that_ special.

MichalPysik: @mdellweg I though so based on the documentation, but --help still acts as it should. Even if I use --repository with an invalid repository name, and put --help AFTER --repository, I still only get the error message printed.

Maybe --help is not only eager, but also has some special property in Click.

MichalPysik: "repo" is a repository that actually exists, "repo2" is nonexistent
```
pulp file repository version show --repository repo --help
Usage: pulp file repository version show [OPTIONS]

  Show details of a repository version.

Options:
  --repository-href TEXT  HREF of the repository
  --repository TEXT       A resource to look for identified by <name> or by
                          <href>.
  --version INTEGER       Repository version number
  --help                  Show this message and exit.
  ```

```
pulp file repository version show --repository repo2 --help
Usage: pulp file repository version show [OPTIONS]

  Show details of a repository version.

Options:
  --repository-href TEXT  HREF of the repository
  --repository TEXT       A resource to look for identified by <name> or by
                          <href>.
  --version INTEGER       Repository version number
  --help                  Show this message and exit.
```

File: CHANGES/650.bugfix [hunk 2]
```diff
@@ -0,0 +1 @@
+Option (also resource lookup option) --repository is now an eager (high priority)  parameter.
```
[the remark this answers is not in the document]
MichalPysik: @daviddavis thanks, you're right. Is the new one any better?

pulpcore/cli/common/generic.py
mdellweg: OK, eager is not the solution, because the real issue in disguise is that `_version_callback` is evaluating the repository context prematurely (before _all_ options have been seen), by accessing it's `pulp_href`. Leading to api calls before we even know the command is syntactically correct.
Instead, it should hook the contexts up lazily.

mdellweg (changes requested):

pulpcore/cli/common/generic.py
```diff
@@ -414,14 +414,9 @@ def _href_callback(
...
-    assert entity_ctx is not None
-    repository_ctx = ctx.find_object(PulpRepositoryContext)
-    assert repository_ctx is not None
-    if value is not None:
-        entity_ctx.pulp_href = f"{repository_ctx.entity['versions_href']}{value}/"
-    else:
-        entity_ctx.pulp_href = repository_ctx.entity["latest_version_href"]
+    repository_version_ctx = ctx.find_object(PulpRepositoryVersionContext)
+    assert repository_version_ctx is not None
+    repository_version_ctx.entity = {"number": value}
```
mdellweg, line 419: That is it? So the solution was there ready to be used all along...
Please add that test i gave you?

gerrod3: Interestingly this doesn't break the scenario where you don't supply the `--version` option because versions are always listed highest first. It does add the message `Not all X entries were shown` to the output, but the behavior is still correct.

mdellweg: What are you telling me? We do the lookup with limit=1 and so we incidentally only get the latest by not using the number filter?

gerrod3: Correct. The behavior before was that if you didn't supply the `--version` option the version callback would set the `RepositoryVersionContext` to the repository's latest version href. Now if you don't supply the option the `find` method will just use the first version returned because of `limit=1` and it just so happens that we always return versions newest/latest first.

MichalPysik: @mdellweg I saw your uniqueness check PR, does implicit latest version in my code ({"number": None}) trigger the exception you added for multiple found results?

MichalPysik: @mdellweg by entity getter you're not referring to PulpRepositoryVersionContext's get_version_context() function right? Since that is called only without arguments in this case, making us unable to use either -1 or None to get the latest version.

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -100,7 +100,7 @@ then
 fi
 
 expect_succ pulp rpm repository version list --repository "cli_test_rpm_repository"
-expect_succ pulp rpm repository version repair --repository "cli_test_rpm_repository" --version 1
+expect_succ pulp rpm repository version repair --version 1 --repository "cli_test_rpm_repository"
```
mdellweg, line 103: Did you change __all__ of these?

MichalPysik: What do you mean? Just changed the order for 3 different test cases (also quite some time ago now)

mdellweg: We want the old version to work nonetheless. So we should probably only change one instance.

> also quite some time ago now

This is an open PR, its all a projection into the future.

mdellweg (approved):

PR #857, opened 2024-01-04
Add note about specifier not including .dev versions

### LINKED ISSUE #852: Version constrain `>=x.y.z` doesn't include unreleased plugin code `x.y.z.dev` on tests
## Summary

When writting a test for a plugin that contains a version constrains `>=x.y.z`, the test won't recognize `x.y.z.dev` as being part of that constrain specification.

## Steps to reproduce

Create a test file and run it with pytest:
```python
from pulp_glue.common.context import PluginRequirement

def test_dev_version_gte_base_version():
    plugin_requirement = PluginRequirement("plugin", specifier=">=3.25.0")
    assert "3.25.0" in plugin_requirement # succeeds
    assert "3.25.0.dev" in plugin_requirement # fails
```

## Expected behavior

Unreleased version (developer versions) of those constrains to be recognized as part of the case specification.

pulpcore: `3.42.0.dev`
pulp-cli:  `0.23.0.dev0`

## Additonal context

When some feature will be added or removed from a plugin, often this should be reflected in the CLI. 
It would be convenient to allow testing those changes as if they were released, without having to wait the real release.

If that's not something desirable at all, we could add that in the docs to clarify one should wait the plugin changes to be released before doing in the CLI.

pedro-psb:
When some feature will be added or removed from a plugin, often this should be reflected in the CLI, but for it to be properly tested, that change should be already released. Using .dev versions isn't be reliable.

closes #852

Review Checklist:


docs/architecture.md
```diff
@@ -111,6 +111,8 @@ class PulpMyPluginRepositoryContext(PulpRepositoryContext):
         return body
 ```
 
+Note that specifier `>=x.y.z` doesn't include `x.y.z.dev`, which  means that when adding or removing a feature from a plugin, that change must first be released before it's constrains can be realiably tested in the CLI.
+
```
mdellweg, line 115: Great addition. Let's spice it up a little:
```suggestion
!!! note
    The specifier `>=x.y.z` doesn't include `x.y.z.dev` according to PEP 440.
    Therefore, when adapting to an unreleased feature change from a plugin, you need to specify the prerelease part of the version explicitly.
    However `>=x.y.z.dev` is never unambiguous in the current Pulp versioning practice.
    Once that change is released please reset the constraint to the plain `x.y.z` schema.

```

Note that the meaning is somewhat different, stating that you _can_ test unreleased features in the CLI, but the "dev" version part is never unambiguous.

Maybe it's time to look into `setuptools-scm` for Pulp. We'd get automatic `dev<n>` versions.

pedro-psb: Thanks, that clarifies what/why to do.
The idea of using `setuptools-scm` is to have unique dev identifiers for each commit, so pre-release specifiers can't live unnoticed?

mdellweg: I believe the idea is to derive the current version of the project from git. But i just realized this will not work, since we do not tag our releases on the main branch. So last "tag + n commits" is meaningless. But it would have been nice to make assertions like `specifier>="1.2.3.dev15"`.

mdellweg (approved):

PR #861, opened 2024-01-11
Fix style for reference docs

mdellweg:
[noissue]

Review Checklist:

### DIFF
--- a/docs/css/mkdocstrings.css
+++ b/docs/css/mkdocstrings.css
@@ -1,0 +1,19 @@
+.doc-symbol-attribute::after {
+  content: "attr";
+}
+
+.doc-symbol-function::after {
+  content: "func";
+}
+
+.doc-symbol-method::after {
+  content: "meth";
+}
+
+.doc-symbol-class::after {
+  content: "class";
+}
+
+.doc-symbol-module::after {
+  content: "mod";
+}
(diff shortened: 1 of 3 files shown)


docs/css/mkdocstrings.css
pedro-psb, line 1: I've tried running this locally and it doesn't seem to make any difference, and indeed, there is no CSS classes with those names in the HTML source of a code-ref page. I've also tried using it recently on the docs project and couldn't make it work

I suspect this is a bug with the [show_symbol_type_heading option](https://github.com/mkdocstrings/python/blob/03e53e3964908d77068ed7d8538a046222b56d55/src/mkdocstrings_handlers/python/templates/material/_base/class.html#L28), because it looks like it should work.

pedro-psb: Also, this one works and helps with readability: https://mkdocstrings.github.io/python/usage/customization/#readthedocs

pedro-psb: Actually, this was just a version mismatch. This was released in [1.8.0](https://github.com/mkdocstrings/python/releases/tag/1.8.0) and works properly after upgrade

pedro-psb (approved):

ggainey (approved):

PR #884, opened 2024-01-23
Remove .dev from pulp_rpm 3.25.0 version specifiers

dralley:
This reverts commit e2a10a2b16b0a5d13129c06d367f3ed4d7e980b5.

[noissue]

### DIFF
--- a/tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
+++ b/tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
@@ -43,7 +43,7 @@ expect_succ pulp rpm repository show --repository "cli_test_rpm_repository"
 test "$(echo "$OUTPUT" | jq -r '.description')" = "null"
 
 # sqlite metadata removal from pulp_rpm (pulp/pulp_rpm#3328)
-if pulp debug has-plugin --name "rpm" --specifier "<3.25.0.dev"
+if pulp debug has-plugin --name "rpm" --specifier "<3.25.0"
 then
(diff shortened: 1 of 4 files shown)


dralley:
What's up with this error?  It seems sporadic, it happened on the other PR as well and went away w/o anything changing, but it's seemingly common..

tests/scripts/pulp_rpm/test_rpm_sync_publish.sh
```diff
@@ -127,18 +127,18 @@ then
   expect_fail pulp rpm repository sync --repository "cli_test_rpm_repository" --sync-policy foobar
 fi
 
-if pulp debug has-plugin --name "rpm" --min-version "3.25.0.dev"
+if pulp debug has-plugin --name "rpm" --max-version "3.24.0"
```
mdellweg, line 130:
```suggestion
if pulp debug has-plugin --name "rpm" --specifier "<3.25.0"
```
Please check the version again...

dralley: I'm aware, but I keep hitting the same error unexpectedly (not the same as the sporadic one mentioned above), so I was trying a few different combinations

mdellweg: https://github.com/pulp/pulp-cli/actions/runs/7641882905/job/20820247714?pr=884#step:7:148
The nightly container is reporting to run `"rpm": "3.25.0.dev"` so we need to wait until that nightly container built with "3.26.0.dev" I would say.

ggainey (approved):

PR #885, opened 2024-01-23
Remove test support for pre-pulpcore-3.18 versions

dralley:
It's been long enough that I think we can drop support for some of the older pulpcore releases. At least in tests. Many of these specifiers are even older, and our CI matrix is only testing 3.18+ anyways.

[noissue]


mdellweg (commented):
Not a bad Idea.

Would you mind taking a look at the test matrix and align it with update-ci-branches in pulpcore?

dralley:
Not a problem, updated.

mdellweg (approved):

PR #895, opened 2024-02-02
Check for uniqueness on entity lookup

### LINKED ISSUE #894: Find does not check for uniquenes as advertised.
Interestingly this doesn't break the scenario where you don't supply the `--version` option because versions are always listed highest first. It does add the message `Not all X entries were shown` to the output, but the behavior is still correct.

_Originally posted by @gerrod3 in https://github.com/pulp/pulp-cli/pull/847#discussion_r1475534432_

mdellweg:
In case more than one entity is found for the specified search parameters, an exception is thrown. This does not validate the search parameters beforehand guaranteed to be a natual key set.

fixes #894

Review Checklist:

### DIFF
--- a/pulp-glue/pulp_glue/common/context.py
+++ b/pulp-glue/pulp_glue/common/context.py
@@ -806,14 +806,24 @@ class PulpEntityContext:
         Raises:
             PulpException: if no entity satisfies the search parameters uniquely.
         """
-        search_result = self.list(limit=1, offset=0, parameters=kwargs)
-        if len(search_result) != 1:
+        payload: Dict[str, Any] = kwargs.copy()
+        payload.update(self.scope)
+        payload["offset"] = 0
+        payload["limit"] = 1
+        result: Mapping[str, Any] = self.call("list", parameters=payload)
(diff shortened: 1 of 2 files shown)


pulp-glue/pulp_glue/common/context.py
```diff
@@ -806,14 +806,24 @@ def find(self, **kwargs: Any) -> Any:
...
             raise PulpException(
                 _("Could not find {entity} with {kwargs}.").format(
                     entity=self.ENTITY, kwargs=kwargs
                 )
             )
-        return search_result[0]
+        if result["count"] > 1:
+            raise PulpException(
+                _("Multiple {entities} found with {kwargs}.").format(
+                    entity=self.ENTITIES, kwargs=kwargs
```
MichalPysik, line 823: {entities} in the message doesn't match ```entity``` in the format function

MichalPysik, line 823:
```suggestion
                    entities=self.ENTITIES, kwargs=kwargs
```

gerrod3 (approved):

PR #902, opened 2024-02-14
Add header option

### LINKED ISSUE #889: Support custom Authorization header
## Summary
As a user I can submit a custom header auth token.

## Background
There is an [auth style used on console.redhat.com](https://source.redhat.com/groups/public/consoledot/consoledot_blog/hcc__service_accounts_with_rbac_support_as_an_alternative_to_basic_auth) which involves using long-lived credentials to then fetch a hours-long credential that is a `Bearer` style token. That `Bearer` token would need to be used by `pulp-cli` to access a Pulp install on console.redhat.com

## Examples
It would be great if somehow the CLI could be configured with a Bearer token to submit a header like:  `Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c`

We'll wrap the CLI took with a script to do the fetching of this and save it in an environment variable. It would be great if we could just pass this header name and value (or just value?) to the CLI with each call. We want to avoid having to save it into a config file because we'll have to update that file every few hours then.

mdellweg:
This allows to pass arbitrary header values along with the requests.

fixes #889

Review Checklist:

### DIFF
--- a/CHANGES/889.feature
+++ b/CHANGES/889.feature
@@ -1,0 +1,1 @@
+Added `--header` parameter to allow passing an arbitrary number of custom headers along with every request.
--- a/pulp-glue/pulp_glue/common/openapi.py
+++ b/pulp-glue/pulp_glue/common/openapi.py
@@ -52,6 +52,7 @@ class OpenAPI:
         base_url: The base URL inlcuding the HTTP scheme, hostname and optional subpaths of the
             served api.
         doc_path: Path of the json api doc schema relative to the `base_url`.
+        headers: Dictionary of additional request headers.
         username: Username used for basic auth.
         password: Password used for basic auth.
         cert: Client certificate used for auth.
@@ -68,6 +69,7 @@ class OpenAPI:
         self,
         base_url: str,
         doc_path: str,
+        headers: Optional[Dict[str, str]] = None,
         username: Optional[str] = None,
         password: Optional[str] = None,
         cert: Optional[str] = None,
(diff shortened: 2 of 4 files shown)


decko:
What happens if we use --headers with an `Authorization` header and the --username/password at the same time? Is it possible we're gonna end with two `Authorization` headers? One gonna overwrite the other?

mdellweg:
> What happens if we use --headers with an `Authorization` header and the --username/password at the same time? Is it possible we're gonna end with two `Authorization` headers? One gonna overwrite the other?

Yes, but just don't provide a username at this point.

CHANGES/889.feature
```diff
@@ -0,0 +1 @@
+Added `--header` parameter to allow passing a custom header along with every request.
```
ggainey, line 1:
```suggestion
Added `--header` parameter to allow passing an arbitrary number of custom headers along with every request.
```

ggainey (changes requested):

ggainey:
> > What happens if we use --headers with an `Authorization` header and the --username/password at the same time? Is it possible we're gonna end with two `Authorization` headers? One gonna overwrite the other?
> 
> Yes, but just don't provide a username at this point.

Yeah - in this instance, "don't do that" is a perfectly valid response to any problems that arise from duplicate-auth-headers...

CHANGES/889.feature
mdellweg: You can even override the user agent string that way now.

ggainey (approved):

PR #915, opened 2024-03-14
Cast

mdellweg:
Review Checklist:


CHANGES/pulp-glue/+cast.feature
gerrod3, line 1: Can you remind me again how these named changelog entries work? Why not file an issue for them?

pulp-glue/tests/conftest.py
```diff
@@ -0,0 +1,26 @@
+import typing as t
+
+import pytest
+
+from pulp_glue.common.context import PulpContext
+from pulp_glue.common.openapi import BasicAuthProvider
+
+
+class PulpTestContext(PulpContext):
+    # TODO check if we can just make the base class ignore echo.
```
gerrod3, line 10: What does this mean, doesn't the base class raise a `NotImplemented` error?

pulp-glue/tests/test_api_quirks.py
```diff
@@ -1,27 +1,18 @@
-import typing as t
 from copy import deepcopy
 
 import pytest
+
 from pulp_glue.common.context import _REGISTERED_API_QUIRKS, PulpContext
 
+pytestmark = pytest.mark.glue
```
gerrod3, line 7: Is this another pytest quirk? How does it work?

pytest_pulp_cli/__init__.py
```diff
@@ -92,11 +92,18 @@ def pulp_cli_settings(tmp_path_factory: pytest.TempPathFactory) -> t.Tuple[pathl
     if os.environ.get("PULP_API_ROOT"):
         for key in settings:
             settings[key]["api_root"] = os.environ["PULP_API_ROOT"]
+    return settings
+
+
+@pytest.fixture(scope="session")
+def pulp_cli_settings_path(
```
gerrod3, line 99: Nice refactor!

pulp-glue/pulp_glue/common/__init__.py
```diff
@@ -1 +1,32 @@
...
+        enabled_plugins: Optional list of plugins to consider for loading.
+    """
+    for entry_point in entry_points(group="pulp_glue.plugins"):
+        name = entry_point.name
+        if (
+            enabled_plugins is None or entry_point.name in enabled_plugins
+        ) and entry_point.name not in loaded_plugins:
+            plugin = entry_point.load()
+            if hasattr(plugin, "mount"):
+                plugin.mount()
```
gerrod3, line 31: Currently none of the pulp_glue plugins have a `mount` method, do you see the need for one in the future?

pulp-glue/pyproject.toml
```diff
@@ -33,6 +33,15 @@ documentation = "https://docs.pulpproject.org/pulp_cli/"
 repository = "https://github.com/pulp/pulp-cli"
 changelog = "https://docs.pulpproject.org/pulp_cli/CHANGES/"
 
+[project.entry-points."pulp_glue.plugins"]
```
gerrod3, line 36: Have you opened up the corresponding PRs in our other cli-plugin repositories? They won't be registered as loaded until they have made this change.

gerrod3 (commented):
Looks good, mainly have clarifying questions before approval.

pulp-glue/tests/conftest.py
mdellweg: It does, and that means, the PulpContext is unusable as is. I think that is a bad thing.

pulp-glue/tests/test_api_quirks.py
mdellweg: It marks all tests in this scope. (Putting it in conftest did not work sadly.)

pulp-glue/pulp_glue/common/__init__.py
mdellweg: I'm not sure. This would be better "have" than "need".

pulp-glue/pyproject.toml
mdellweg: That is right, but i think it should be safe to fade this in slowly.

PR #916, opened 2024-03-14
Add Token and OAuth2 authentication methods

decko:
Review Checklist:


pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -53,36 +55,79 @@ class AuthProviderBase:
     Returned auth objects need to be compatible with `requests.auth.AuthBase`.
     """
 
-    def basic_auth(self) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.AuthBase]]:
```
mdellweg, line 56: No, you don't get to change the interface.

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -53,36 +55,79 @@ class AuthProviderBase:
...
-    ) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.AuthBase]]:
-        for proposal in security:
-            if [security_schemes[name] for name in proposal] == [
-                {"type": "http", "scheme": "basic"}
-            ]:
-                result = self.basic_auth()
-                if result:
-                    return result
-        raise OpenAPIError(_("No suitable auth scheme found."))
+    def __call__(self, request: requests.PreparedRequest) -> requests.PreparedRequest:
```
mdellweg, line 64: This is an Auth provider. It's supposed to provide appropriate AuthBase objects. It's not authenticating anything.

pulp-glue/pyproject.toml
```diff
@@ -25,6 +25,7 @@ classifiers = [
 dependencies = [
   "packaging>=20.0,<24",
   "requests>=2.24.0,<2.32",
+  "requests-oauthlib>=1.4.0,<1.5.0"
```
dkliban, line 28: you are missing a comma at the end.

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -53,36 +55,79 @@ class AuthProviderBase:
...
+        self.token = token
+
+    def auth(self, request) -> requests.PreparedRequest:
+        request.headers["Authorization"] = self.token
+        return request
+
+
+class ThirdPartyAuthProvider(AuthProviderBase):
+
+    auth_type = "third_party"
```
mdellweg, line 114: Why can we not just call it oauth2?
Also isn't there a requests-oauth out there?

pulp_cli/config.py
```diff
@@ -58,6 +61,9 @@
     ),
     click.option("--username", default=None, help=_("Username on pulp server")),
     click.option("--password", default=None, help=_("Password on pulp server")),
+    click.option("--client_id", default=None, help=_("client_id on Identity Provider")),
```
mdellweg, line 64: Username?

pulp_cli/config.py
```diff
@@ -58,6 +61,9 @@
...
+    click.option("--client_secret", default=None, help=_("client_secret on Identity Provider")),
```
mdellweg, line 65: password?

pulp_cli/config.py
```diff
@@ -58,6 +61,9 @@
...
+    click.option("--token_url", default=None, help=_("URL to reach the Identity Provider")),
```
mdellweg, line 66: The user does not know that. It must be extracted from the www-authenticate header.

mdellweg (changes requested):
This is heading in the wrong direction.

pulp-glue/pulp_glue/common/openapi.py
decko: Renamed it, and yes, we're using `request-oauthlib`. Here https://github.com/pulp/pulp-cli/blob/0738418f7247b7e901c03ea69b6997cb6842653e/pulp-glue/pyproject.toml#L28

pulp-glue/pulp_glue/common/openapi.py
decko: It's returning an `AuthBase` compatible object, like stated [here](https://github.com/pulp/pulp-cli/blob/375f8b0952e33c409eed58c396fe0feedb7e43eb/pulp-glue/pulp_glue/common/openapi.py#L55) and based on [this](https://requests.readthedocs.io/en/latest/_modules/requests/auth/#AuthBase).
If this is about the name of the `auth` method it was based on the previous one `basic_auth`, which I think it's related only to HTTP Basic Authentication method.

pulp-glue/pulp_glue/common/openapi.py
decko: Yeah, but the way the `AuthProviderBase` is written the `__call__` method is incompatible with the original interface and also only calls the `basic_auth` method. This way every time we need to write a new AuthProvider we have to override the `__call__` method to: first, be compliant with `AuthBase` objects, and second: to call another method that would make more send for different authentication method.

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -133,6 +178,7 @@ def __init__(
             if cert or key:
                 raise OpenAPIError(_("Cannot use both 'auth' and 'cert'."))
             self.auth_provider = auth_provider
+            self._session.auth = auth_provider
```
decko, line 181: need to check for a better way to provide the authentication if needed.

pulp_cli/config.py
decko: Although the OAuth2 RFC presents a `WWW-Authenticate` section [here](https://www.rfc-editor.org/rfc/rfc6750#section-3), it's not clear that it should contain the issuer URL where we could ask for a new token, for example. 
Also, I could be wrong here.

pulp-glue/pulp_glue/common/openapi.py
mdellweg: It is supposed to _return_, not _be_ an AuthBase.

decko:
> This is heading in the wrong direction.

Ok, starting over on another PR.

mdellweg:
> > This is heading in the wrong direction.
> 
> Ok, starting over on another PR.

Please don't. That way we loose all the discussion. Please drop the new commits in the current PR.

PR #929, opened 2024-03-20
Check only for schema and netloc on base_url.

### LINKED ISSUE #928: config command stopped working after 41538e6
## Summary
After updating pulp-cli and pulp-glue to the latest version of `main` branch, the config command keeps complaining about the `base_url` field.

## Steps to reproduce

Running the command:
```
pulp config create --username admin --password password --base-url http://localhost:5001/pulp/api/v3/
```

always return the following prompt:
```
Config file failed to parse. ('base_url' should be of the form '<schema>://<netloc>').
Continue without config? [y/N]:
```

## Expected behavior

It should run without issues and create or update the config file.

Only happens after 41538e6. 8437376 and before is not affected by this.

decko:
Closes #928

### DIFF
--- a/pulp_cli/config.py
+++ b/pulp_cli/config.py
@@ -112,14 +112,7 @@ def validate_config(config: t.Dict[str, t.Any], strict: bool = False) -> None:
     if "base_url" in config:
         try:
             parsed_url = urlparse(config["base_url"])
-            if (
-                not parsed_url.scheme
-                or not parsed_url.netloc
-                or parsed_url.path
-                or parsed_url.params
-                or parsed_url.query
-                or parsed_url.fragment
-            ):
+            if not parsed_url.scheme or not parsed_url.netloc:
                 errors.append(_("'base_url' should be of the form '<schema>://<netloc>'"))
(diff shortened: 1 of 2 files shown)


pulp_cli/config.py
```diff
@@ -112,14 +112,7 @@ def validate_config(config: t.Dict[str, t.Any], strict: bool = False) -> None:
...
-                or not parsed_url.netloc
-                or parsed_url.path
-                or parsed_url.params
-                or parsed_url.query
-                or parsed_url.fragment
-            ):
+            if not parsed_url.scheme or not parsed_url.netloc:
                 errors.append(_("'base_url' should be of the form '<schema>://<netloc>'"))
         except ValueError:
             errors.append(_("Failed to parse the 'base_path'."))
```
mdellweg, line 118: Hah, this is the only erroneous line i can spot.
```suggestion
            errors.append(_("Failed to parse the 'base_url'."))
```

mdellweg:
Do you want to use this PR to improve the error message, or should we just close it?

PR #930, opened 2024-03-21
Fix changelog parser

mdellweg:
The original version was not coping well when the version happened to appear twice in the title of a changelog.

[noissue]

Review Checklist:

### DIFF
--- a/.ci/scripts/collect_changes.py
+++ b/.ci/scripts/collect_changes.py
@@ -21,8 +21,12 @@ START_STRING = tc_settings.get(
 TITLE_FORMAT = tc_settings.get("title_format", "{name} {version} ({project_date})")
 
 
+# Build a regex to find the header of a changelog section.
+# It must have a single capture group to single out the version.
+# see help(re.split) for more info.
 NAME_REGEX = r".*"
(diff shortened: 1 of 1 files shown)


.ci/scripts/collect_changes.py
```diff
@@ -22,14 +22,16 @@
...
+VERSION_CAPTURE_REGEX = rf"({VERSION_REGEX})"
 DATE_REGEX = r"[0-9]{4}-[0-9]{2}-[0-9]{2}"
 TITLE_REGEX = (
     "("
     + re.escape(
         TITLE_FORMAT.format(name="NAME_REGEX", version="VERSION_REGEX", project_date="DATE_REGEX")
     )
     .replace("NAME_REGEX", NAME_REGEX)
+    .replace("VERSION_REGEX", VERSION_CAPTURE_REGEX, 1)
     .replace("VERSION_REGEX", VERSION_REGEX)
```
ggainey, line 35: I don't think I grok why we're replacing VERSION_REGEX twice here - can you explain?

mdellweg: We replace the first instance with (capture-)paranthesies and the rest with the plain regex that "just matches".

ggainey (approved):

PR #937, opened 2024-04-02
Split architecture doc

mdellweg:
Review Checklist:

### DIFF
--- a/staging_docs/dev/learn/architecture.md
+++ b/staging_docs/dev/learn/architecture.md
@@ -1,55 +1,39 @@
-# Architecture
-
-The Pulp CLI architecture is described in this section.
+# Pulp CLI Architecture
 
 ## Pulp Glue
 
-Pulp CLI provides the `pulp-glue` library as an abstraction layer that lets you perform high-level operations in pulp.
-Its goal is to abstract interacting with the REST api by parsing the api docs, and waiting on tasks and task groups.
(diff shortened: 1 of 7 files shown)


staging_docs/dev/learn/architecture.md
pedro-psb, line 1: Internal links are missing a `docs` node: `site:pulp-glue/docs/learn/...`.

pedro-psb (changes requested):

staging_docs/dev/learn/architecture.md
mdellweg: Can you help me create a docs test building workflow?

mdellweg: More strict would definitely be helpful.
If I understand correctly, pulp-docs will always (on each subprojects PRs) build the whole docs site, right? We should be as strict as possible in order not to break the global docs building constantly.

pedro-psb: Yes it builds the whole doc everytime.

But the effect on CI stability is quite the opposite, IMO:
If we are strict about broken links, for example, and some repository merges a change with broken links (somehow bypassing PR checks), then every build will fail because of that, whereas a less strict policy on broken links wouldn't prevent building.

mdellweg: OK, that's why we need to be at least as strict in PR review.
(bypassing such a check can be considered deliberately evil behaviour)

ggainey (approved):

PR #939, opened 2024-04-10
Adjust the cookiecutters

mdellweg:
This will help with the transition to bump-my-version and the new staging docs.

[noissue]

Review Checklist:


ggainey (approved):
looks reasonable

PR #941, opened 2024-04-18
Has cli

mdellweg:
Review Checklist:


ggainey (changes requested):
Can we get tests for the new command and new --version output?

mdellweg:
Tests in place now.

PR #945, opened 2024-04-22
Format jaml

mdellweg:
Review Checklist:


mdellweg:
This format should really live in a library to be used here.

PR #949, opened 2024-04-23
Add docs migration warning and redirects from old pages

### LINKED ISSUE #932: Add redirects from old docs to new docs
## Summary

The old docs publishing was deactivated and its content was moved to [the new docs](https://staging-docs.pulpproject.org/) (see #910).

When the moved content is live on the staging-docs, we should:
- publish the last archive version of the old docs. Should contain a migration note on the index.
- add redirect routes to the new docs (preferably mapping individual pages).

We might use https://github.com/mkdocs/mkdocs-redirects.

## Examples
- If someone enters `https://docs.pulpproject.org/pulp_cli/` no redirects take place, but the index page contains the migration note in the spotlight.
- If someone enters `https://docs.pulpproject.org/pulp_cli/installation/`, the page is redirected to the corresponding one in `https://staging-docs.pulpproject.org/`.

pedro-psb:
See https://github.com/pulp/pulp-docs/issues/33

This adds a single page to the old documentation site with a note to the docs migration. Other pages are redirected to this, so people can understand what is going on before going to the new site.

The idea is that this gets published and stays freezed until July, when we should "deactivate" all legacy docs and add broad-redirects to the production site (under pulpproject.org)

Closes #932 

Review Checklist:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,0 +1,9 @@
+# Pulp 3 CLI Docs
+
+The documentation for pulp-cli and pulp-glue has moved to a new place.
+Go to the new site:
+
+- [pulp-cli section](https://staging-docs.pulpproject.org/pulp-cli/docs/user/)
+- [pulp-glue section](https://staging-docs.pulpproject.org/pulp-glue/docs/dev/learn/architecture/)
+
+Learn more about the new documentation project [here](https://discourse.pulpproject.org/t/unified-documentation-part-2/1188/2)
--- a/CHANGES/932.misc
+++ b/CHANGES/932.misc
@@ -1,0 +1,1 @@
+Added warning about the new docs location on the legacy docs and included some redirects.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -17,39 +17,20 @@ markdown_extensions:
   - "pymdownx.tabbed":
       alternate_style: false
 plugins:
-  - "search"
-  - mkdocstrings:
-      default_handler: "python"
-      handlers:
-        python:
-          paths:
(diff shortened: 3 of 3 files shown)


pedro-psb:
Hey @mdellweg, I would like some general feedback on this approach before undrafting.

docs/index.md
```diff
@@ -0,0 +1,8 @@
+# Pulp 3 CLI Docs
+
+The documentation for pulp-cli and pulp-glue has moved to a new place.
+
+Learn more about the new documentation project [here](https://discourse.pulpproject.org/t/unified-documentation-part-2/1188/2) or go to the new site:
+
+- [pulp-cli section](https://staging-docs.pulpproject.org/pulp-cli/docs/user/)
+- [pulp-glue section](https://staging-docs.pulpproject.org/pulp-glue/docs/dev/learn/architecture/)
```
mdellweg, line 8: Looks good in general.
I'd point the users to the new location first and then provide a link to the why however.

pedro-psb: You mean, move the user directly to staging-docs (like [here](https://staging-docs.pulpproject.org/pulp-cli/docs/user/)) and edit the page there to contain the migration note?

mdellweg: No, i just meant to shuffle the lines in this file around.

Whoever got here wants to know where it's now. Only a tiny portion of visitors may be interested in the why. So move the important links to the top.

pedro-psb:
I'm not sure how this will get published, but it will be required install `mkdocs-redirects`.

mdellweg (approved):
I take it you have tested this... XD

PR #977, opened 2024-06-04
Add support for new pulp_rpm "prune-packages" feature.

ggainey:


ggainey:
Needed:

- [ ] rpm feature to get merged (https://github.com/pulp/pulp_rpm/pull/3536)
- [ ] Issue
- [ ] Tests

pulpcore/cli/rpm/prune.py
```diff
@@ -0,0 +1,109 @@
...
+)
+
+
+@pulp_command()
+@multi_repository_option
+@click.option(
+    "--repository-href",
+    multiple=True,
+    help=_("Repository HREF of a Reppository to prune. Can be called multiple times."),
+)
```
mdellweg, line 37: Don't do this. The resource option accepts hrefs already.

pulpcore/cli/rpm/prune.py
```diff
@@ -0,0 +1,109 @@
...
+    pass_pulp_context,
+    pulp_command,
+    resource_option,
+)
+
+_ = gettext.gettext
+
+multi_repository_option = resource_option(
+    "--repository",
+    default_plugin="rpm",
```
mdellweg, line 19: Give that variable a meaningful name.
```suggestion
    "--repository",
    "repositories",
    default_plugin="rpm",
```

pulpcore/cli/rpm/prune.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    if all_repositories:
+        repo_hrefs = ["*"]
+    else:
+        repo_hrefs = [
+            repository_ctx.pulp_href
+            for repository_ctx in repository
+            if isinstance(repository_ctx, PulpRpmRepositoryContext)
+        ] + list(repository_href)
```
mdellweg, line 100: This transformation is unnecessary, as it will be carried out in pulp_glue.

pulpcore/cli/rpm/prune.py
```diff
@@ -0,0 +1,100 @@
...
+            repository_ctx.pulp_href
+            for repository_ctx in repositories
+            if isinstance(repository_ctx, PulpRpmRepositoryContext)
+        ]
+
+    params = {
+        "repo_hrefs": repo_hrefs,
+        "keep_days": keep_days,
+        "dry_run": dry_run,
+    }
```
mdellweg, line 98: Please try this:
```suggestion
    if all_repositories:
        repositories = ["*"]

    params = {
        "repo_hrefs": repositories,
        "keep_days": keep_days,
        "dry_run": dry_run,
    }
```

ggainey: I had tried that and it failed with "'repo_hrefs' is expected to be a list." I just did a little deeper experimentation,. and discovered that the problem is that, coming into this call, repositories is a generator. If I list(repositories), the correct validation happens in the correct place and All Is Well :)  next update will address.

ggainey: Nope - you can't "just pass" the generator, and you  can't pass `list(repositories)` because mypy complains `error: List comprehension has incompatible type List[str | PulpEntityContext | None]; expected List[str]` and the lint-check fails.

Restoring to original comprehension.

mdellweg: Hmm. This raises the question whether it's good to allow `["*"]` as a "list of hrefs" on the server side. How is this represented in the openapi spec? If it were a "list of repository_hrefs" i'd expect to see `{"type": "array", "items": {"type": "str", "format": "uri"}`.

mdellweg: What happens if you provided an empty list?
What, if you don't pass it at all?

The output of `pulp debug openapi operation --id rpm_prune_prune` or so should help us. The referenced (body) schema is of interest too.

ggainey: > What happens if you provided an empty list? What, if you don't pass it at all?
```
$ pulp rpm prune-packages 
Error: at least one --repository, or --all-repositories, must be specified
```

> The output of `pulp debug openapi operation --id rpm_prune_prune` or so should help us. The referenced (body) schema is of interest too.
```
$ pulp debug openapi schema --name PrunePackages  
{
  "type": "object",
  "description": "Serializer for prune-old-Packages operation.",
  "properties": {
    "repo_hrefs": {
      "type": "array",
      "items": {
        "type": "string",
        "minLength": 1
      },
      "description": "Will prune old packages from the specified list of repos. Use ['*'] to specify all repos. Will prune based on the specified repositories' latest_versions."
    },
    "keep_days": {
      "type": "integer",
      "format": "int64",
      "default": 14,
      "description": "Prune packages introduced *prior-to* this many days ago. Default is 14. A value of 0 implies 'keep latest package only.'",
      "minimum": 0
    },
    "dry_run": {
      "type": "boolean",
      "default": false,
      "description": "Determine what would-be-pruned and log the list of packages. Intended as a debugging aid."
    }
  },
  "required": [
    "repo_hrefs"
  ]
}
```

mdellweg: So repo_hrefs is required and must contain at least one href or "*".
Thanks, that gives some insight.

ggainey: See current refactoring - took some pointers from https://github.com/pulp/pulp-cli/pull/749 , now simpler and keeps mypy happy.

pulpcore/cli/rpm/prune.py
```diff
@@ -0,0 +1,89 @@
...
+    type=bool,
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help=_("Evaluate the prune-status of the specified repositories but DO NOT make any changes."),
+)
+@pass_pulp_context
+def prune_packages(
+    pulp_ctx: PulpCLIContext,
+    repositories: t.Iterable[str],
```
mdellweg, line 59: I think, this should be:
```suggestion
    repositories: t.Iterable[PulpRpmRepositoryContext],
```
And I'm sad, click's annotation does not catch these.

mdellweg (approved):

PR #980, opened 2024-06-05
Introduce ViewSetContext

mdellweg:
Review Checklist:

### DIFF
--- a/pulpcore/cli/core/orphan.py
+++ b/pulpcore/cli/core/orphan.py
@@ -1,15 +1,15 @@
 import typing as t
 
 import click
-from pulp_glue.common.context import PluginRequirement, PulpEntityContext
+from pulp_glue.common.context import PluginRequirement, PulpViewSetContext
 from pulp_glue.common.i18n import get_translation
 from pulp_glue.core.context import PulpOrphanContext
 
 from pulpcore.cli.common.generic import (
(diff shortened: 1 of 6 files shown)


pulpcore/cli/core/orphan.py
```diff
@@ -47,7 +47,7 @@ def orphan(ctx: click.Context, pulp_ctx: PulpCLIContext) -> None:
     ),
     needs_plugins=[PluginRequirement("core", specifier=">=3.15.0")],
 )
-@pass_entity_context
+@pass_view_set_context
 @pass_pulp_context
 def cleanup(pulp_ctx: PulpCLIContext, orphan_ctx: PulpEntityContext, **kwargs: t.Any) -> None:
```
ggainey, line 52:
```suggestion
def cleanup(pulp_ctx: PulpCLIContext, orphan_ctx: PulpViewSetContext, **kwargs: t.Any) -> None:
```

Am I right?

ggainey (changes requested):

pulpcore/cli/core/orphan.py
mdellweg: Indeed. In fact, if this still works something else must be wrong!?

ggainey (approved):

PR #1000, opened 2024-06-25
Added --directory to `rpm content -t package upload`.

### LINKED ISSUE #994: Provide easier workflow for uploading rpms into a repository
## Summary
Today with pulp-cli, if a user wants to upload 3 rpms, they have to run:

pulp artifact upload --file "$PKG1"
pulp artifact upload --file "$PKG2"
pulp artifact upload --file "$PKG3"

Then for each of those you have to grab the href of each artifact and fetch the sha256:

pulp show --href "${ARTIFACT_HREF_1}" | jq -r '.sha256')
pulp show --href "${ARTIFACT_HREF_2}" | jq -r '.sha256')
pulp show --href "${ARTIFACT_HREF_3}" | jq -r '.sha256'

Then for each of those, they have to use the sha256 to create a content unit and grab the unit href:

PACKAGE_HREF1=$(pulp rpm content create  --sha256 "${ARTIFACT_SHA256}"  | jq -r '.pulp_href')
PACKAGE_HREF2=$(pulp rpm content create  --sha256 "${ARTIFACT_SHA256}"  | jq -r '.pulp_href')
PACKAGE_HREF3=$(pulp rpm content create  --sha256 "${ARTIFACT_SHA256}"  | jq -r '.pulp_href')

Finally you can add these to a repository:

TASK_HREF=$(pulp rpm repository content modify \
            --repository "${REPO_NAME}" \
            --add-content "[{\"pulp_href\": \"${PACKAGE_HREF1}\"}, {\"pulp_href\": \"${PACKAGE_HREF2}\"}, {\"pulp_href\": \"${PACKAGE_HREF3}\"}]" \
            2>&1 >/dev/null | awk '{print $4}')

This is a LOT of commands to run for a user and there isn't an easy way for me to easily give a couple commands to a user to run to upload an arbitrary list of rpms.  Having a command that handled everything and simply took in a list of rpms and outputted a repo version would be much simpler.

An example of what i'm thinking of:

```
# pulp rpm repository content upload  --repository "${REPO_NAME}"  --files=*.rpm

A new repository version has been created at  /pulp/api/v3/repositories/rpm/abcdef/versions/5
```

This command would need to handle things such as:
* An rpm already exists (such as if i interrupt the command half way through and restart it)

ggainey:
This finds all *.rpm in the specified directory and arranges for them to be added to the specified --repository as a single repository-version.

closes #994.

Review Checklist:

### DIFF
--- a/pulpcore/cli/rpm/content.py
+++ b/pulpcore/cli/rpm/content.py
@@ -1,4 +1,6 @@
+import glob
 import typing as t
+from uuid import uuid4
 
 import click
 from pulp_glue.common.context import PulpEntityContext
@@ -6,6 +8,7 @@ from pulp_glue.common.i18n import get_translation
 from pulp_glue.core.context import PulpArtifactContext
 from pulp_glue.rpm.context import (
     PulpRpmAdvisoryContext,
+    PulpRpmCopyContext,
     PulpRpmDistributionTreeContext,
     PulpRpmModulemdContext,
     PulpRpmModulemdDefaultsContext,
@@ -264,9 +267,18 @@ content.add_command(
 @pulp_option(
     "--file",
     type=click.File("rb"),
-    required=True,
-    help=_("An RPM binary"),
+    help=_("An RPM binary. One of --file or --directory is required."),
     allowed_with_contexts=(PulpRpmPackageContext,),
+    required=False,
+)
+@pulp_option(
+    "--directory",
+    type=click.Path(exists=True, readable=True, file_okay=False, dir_okay=True),
+    help=_(
+        "A directory containing RPM binaries named .rpm. One of --file or --directory is required."
+    ),
+    allowed_with_contexts=(PulpRpmPackageContext,),
+    required=False,
 )
 @pulp_option(
     "--file",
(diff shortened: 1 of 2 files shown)


ggainey:
POC - entire approach may not be viable based on expected-user-scenario...

pulpcore/cli/rpm/content.py
```diff
@@ -285,10 +297,81 @@ def upload(
...
+            body: t.Dict[str, t.Any] = {
+                "name": f"uploadtmp_{dest_repo_ctx.entity['name']}_{uuid4()}",
+                "retain_repo_versions": 1,
+                "autopublish": False,
+            }
+            result = tmp_repo_ctx.create(body=body)
+            # Do the deed
+            try:
+                rpms_path = f"{directory}/*.rpm"
+                rpm_names = glob.glob(rpms_path)
```
mdellweg, line 322: I think, we could determine this list first and decide on whether which upload method we use depending on 0-1-many files.

pulpcore/cli/rpm/content.py
```diff
@@ -285,10 +297,81 @@ def upload(
...
+        else:
+            # Create a temp-repo
+            dest_repo_ctx = kwargs["repository"]
+            tmp_repo_ctx = PulpRpmRepositoryContext(pulp_ctx)
+            body: t.Dict[str, t.Any] = {
+                "name": f"uploadtmp_{dest_repo_ctx.entity['name']}_{uuid4()}",
+                "retain_repo_versions": 1,
+                "autopublish": False,
+            }
+            result = tmp_repo_ctx.create(body=body)
```
mdellweg, line 318: Obviously, we could provide a `--tmp-repository` parameter to circumvent the need for a creation permission. But that does not sound like a good idea. And running multiple of these operations with the same tmp storage won't give the desired results.
If we had a stable naming scheme, (e.g. `^tmp[/a]{8}$` or maybe `^tmp/.*$`) we could add a regex based access policy helper to allow anyone to create those repositories.

mdellweg (commented):
Thanks for putting this together.
Some instant ideas follow:

ggainey:
Closing - the functionallity here has been subsumed into #1003 (q.v.)

PR #1013, opened 2024-07-11
Add OAuth2 ClientCredentials flow support

### LINKED ISSUE #926: Add support for issue a token from an IdentityProvider using OAuth2 or OpenIDConnect `client_credential` grant.
## Summary
Using OAuth2 or OpenIDConnect `client_credential` grant, we can issue a token from an third party Identity Provider to be used to authenticate against Pulp or an API Gateway in front of it. This `access_token` would be sent as an `Authorization` header along the request to the Pulp instance.

## Examples
As an example we could use:
```pulp --client_id <client_id> --client_secret <client_secret> --token_url <https://identityprovider.com/oidc/token> status```
to request pulp status.

### PR DESCRIPTION
**Pull request by decko**
This work adds OAuth2 ClientCredentials grant flow support to the Pulp-CLI.
Closes #926 

***
**We're going with the short-term pragmatic approach, and reserving the right to completely refactor this whole area soon**
***

As test strategy, we have two options here:

1- Use the `responses` lib (link [here](https://github.com/getsentry/responses)), a library that mocks the `requests` lib. With that we can intercept calls to a IdP mock to obtain a token, and then intercept the call to a Pulp instance and assert that there's the `access_token` being used with the `Authorization` header.

2- Mock an IdP using nginx rewrite module to return a 200 Ok message with an appropriate response to the token request. Together we need to configure the pulp server to provide the proper URL to allow pulp-cli to request a token and test an authorized operation on it.

### DIFF
--- a/pulp-glue/pulp_glue/common/authentication.py
+++ b/pulp-glue/pulp_glue/common/authentication.py
@@ -1,0 +1,91 @@
+import typing as t
+from datetime import datetime, timedelta
+
+import requests
+
+
+class OAuth2ClientCredentialsAuth(requests.auth.AuthBase):
+    """
+    This implements the OAuth2 ClientCredentials Grant authentication flow.
+    https://datatracker.ietf.org/doc/html/rfc6749#section-4.4
+    """
+
+    def __init__(
+        self,
+        client_id: str,
+        client_secret: str,
+        token_url: str,
+        scopes: t.List[str],
+    ):
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.token_url = token_url
+        self.scopes = scopes
+
+        self.access_token: t.Optional[str] = None
+        self.expire_at: t.Optional[datetime] = None
+
+    def __call__(self, request: requests.PreparedRequest) -> requests.PreparedRequest:
+        if self.expire_at is None or self.expire_at < datetime.now():
+            self.retrieve_token()
+
+        assert self.access_token is not None
+
+        request.headers["Authorization"] = f"Bearer {self.access_token}"
+
+        # Call to untyped function "register_hook" in typed context
+        request.register_hook("response", self.handle401)  # type: ignore[no-untyped-call]
+
+        return request
+
+    def handle401(
+        self,
+        response: requests.Response,
+        **kwargs: t.Any,
+    ) -> requests.Response:
+        if response.status_code != 401:
+            return response
+
+        # If we get this far, probably the token is not valid anymore.
+
+        # Try to reach for a new token once.
+        self.retrieve_token()
+
+        assert self.access_token is not None
+
+        # Consume content and release the original connection
+        # to allow our new request to reuse the same one.
+        response.content
+        response.close()
+        prepared_new_request = response.request.copy()
+
+        prepared_new_request.headers["Authorization"] = f"Bearer {self.access_token}"
+
+        # Avoid to enter into an infinity loop.
+        # Call to untyped function "deregister_hook" in typed context
+        prepared_new_request.deregister_hook(  # type: ignore[no-untyped-call]
+            "response", self.handle401
+        )
+
+        # "Response" has no attribute "connection"
+        new_response: requests.Response = response.connection.send(prepared_new_request, **kwargs)
+        new_response.history.append(response)
+        new_response.request = prepared_new_request
+
+        return new_response
+
+    def retrieve_token(self) -> None:
+        data = {
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+            "scope": " ".join(self.scopes),
+            "grant_type": "client_credentials",
+        }
+
+        response: requests.Response = requests.post(self.token_url, data=data)
+
+        response.raise_for_status()
+
+        token = response.json()
+        self.expire_at = datetime.now() + timedelta(seconds=token["expires_in"])
+        self.access_token = token["access_token"]
--- a/pulp-glue/pulp_glue/common/openapi.py
+++ b/pulp-glue/pulp_glue/common/openapi.py
@@ -24,6 +24,7 @@ UploadType = t.Union[bytes, t.IO[bytes]]
 SAFE_METHODS = ["GET", "HEAD", "OPTIONS"]
 ISO_DATE_FORMAT = "%Y-%m-%d"
 ISO_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+AUTH_PRIORITY = ("oauth2", "basic")
 
 
 class OpenAPIError(Exception):
@@ -44,6 +45,30 @@ class UnsafeCallError(OpenAPIError):
     pass
 
 
+class OpenAPISecurityScheme:
+    def __init__(self, security_scheme: t.Dict[str, t.Any]):
+        self.security_scheme = security_scheme
+
+        self.security_type = self.security_scheme["type"]
+        self.description = self.security_scheme.get("description", "")
+
+        if self.security_type == "oauth2":
+            try:
+                self.flows: t.Dict[str, t.Any] = self.security_scheme["flows"]
+                client_credentials: t.Optional[t.Dict[str, t.Any]] = self.flows.get(
+                    "clientCredentials"
+                )
+                if client_credentials:
+                    self.flow_type: str = "clientCredentials"
+                    self.token_url: str = client_credentials["tokenUrl"]
+                    self.scopes: t.List[str] = list(client_credentials["scopes"].keys())
+            except KeyError:
+                raise OpenAPIValidationError
+
+        if self.security_type == "http":
+            self.scheme = self.security_scheme["scheme"]
+
+
 class AuthProviderBase:
     """
     Base class for auth providers.
@@ -57,18 +82,39 @@ class AuthProviderBase:
         """Implement this to provide means of http basic auth."""
         return None
 
+    def oauth2_client_credentials_auth(
+        self, flow: t.Any
+    ) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.AuthBase]]:
+        """Implement this to provide other authentication methods."""
+        return None
+
     def __call__(
         self,
         security: t.List[t.Dict[str, t.List[str]]],
         security_schemes: t.Dict[str, t.Dict[str, t.Any]],
     ) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.AuthBase]]:
+
+        authorized_schemes = []
+        authorized_schemes_types = set()
+
         for proposal in security:
-            if [security_schemes[name] for name in proposal] == [
-                {"type": "http", "scheme": "basic"}
-            ]:
-                result = self.basic_auth()
-                if result:
-                    return result
+            for name in proposal:
+                authorized_schemes.append(security_schemes[name])
+                authorized_schemes_types.add(security_schemes[name]["type"])
+
+        if "oauth2" in authorized_schemes_types:
+            for flow in authorized_schemes:
+                if flow["type"] == "oauth2":
+                    oauth2_flow = OpenAPISecurityScheme(flow)
+
+            if oauth2_flow.flow_type == "client_credentials":
+                result = self.oauth2_client_credentials_auth(oauth2_flow)
+            if result:
+                return result
+        elif "http" in authorized_schemes_types:
+            result = self.basic_auth()
+            if result:
+                return result
         raise OpenAPIError(_("No suitable auth scheme found."))
 
 
--- a/pulpcore/cli/common/generic.py
+++ b/pulpcore/cli/common/generic.py
@@ -10,6 +10,7 @@ import click
 import requests
 import schema as s
 import yaml
+from pulp_glue.common.authentication import OAuth2ClientCredentialsAuth
 from pulp_glue.common.context import (
     DATETIME_FORMATS,
     DEFAULT_LIMIT,
@@ -230,6 +231,21 @@ class PulpCLIAuthProvider(AuthProviderBase):
             else:
                 self.pulp_ctx.password = click.prompt("Password", hide_input=True)
         return (self.pulp_ctx.username, self.pulp_ctx.password)
+
+    def oauth2_client_credentials_auth(
+        self, oauth2_flow: t.Any
+    ) -> t.Optional[requests.auth.AuthBase]:
+        if self.pulp_ctx.username is None:
+            self.pulp_ctx.username = click.prompt("Username/ClientID")
+        if self.pulp_ctx.password is None:
+            self.pulp_ctx.password = click.prompt("Password/ClientSecret")
+
+        return OAuth2ClientCredentialsAuth(
+            client_id=self.pulp_ctx.username,
+            client_secret=self.pulp_ctx.password,
+            token_url=oauth2_flow.token_url,
+            scopes=oauth2_flow.scopes,
+        )
 
 
 ##############################################################################
--- a/CHANGES/926.feature
+++ b/CHANGES/926.feature
@@ -1,0 +1,2 @@
+Added support to OAuth2 ClientCredentials grant flow as authentication method.
+This is tech preview and may change without previous warning.
--- a/docs/user/reference/_SUMMARY.md
+++ b/docs/user/reference/_SUMMARY.md
@@ -1,2 +1,3 @@
 * [Using the CLI](using_the_cli.md)
 * [Supported Workflows](supported_workflows.md)
+* [Authentication Methods](authentication.md)
--- a/docs/user/reference/authentication.md
+++ b/docs/user/reference/authentication.md
@@ -1,0 +1,15 @@
+# Supported Authentication Methods
+
+Pulp-CLI support some authentication methods to authenticate against a Pulp instance.
+Some very simple and common, like HTTP Basic Auth, and some more complex like OAuth2.
+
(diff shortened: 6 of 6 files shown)

### CODE REVIEW

**Comment by dkliban** [comment]
@decko @mdellweg I used this PR to interact with our deployment that uses oauth2 for authentication. I set the username to client_id and the password to client_secret. I was able to create a domain, a repository, a remote, and sync the repository. Thanks for this awesome feature!

File: pulp-glue/pulp_glue/common/authentication.py [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+    def is_token_expired(self) -> bool:
+        if self.token:
+            issued_at = datetime.fromisoformat(self.token["issued_at"])
+            expires_in = timedelta(seconds=self.token["expires_in"])
+            token_timedelta = issued_at + expires_in - timedelta(seconds=5)
+
+            if token_timedelta >= datetime.now():
+                return False
+
+        return True
```
mdellweg, line 67: Instead of this rather long check, let's persist `expires_at` on the object and replace the call to this function with `if expires_at is None or datetime.now() > expires_at`

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -57,18 +84,36 @@ def basic_auth(self) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.Auth
         """Implement this to provide means of http basic auth."""
         return None
 
+    def auth(self, flow: t.Any) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.AuthBase]]:
+        """Implement this to provide other authentication methods."""
+        return None
+
```
mdellweg, line 90: in search for a better name...
And the docstring should reflect what authentication this is for.

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -85,6 +130,16 @@ def basic_auth(self) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.Auth
         return (self.username, self.password)
 
 
+class OAuth2AuthProvider(AuthProviderBase):
```
mdellweg, line 133: Looks like dead code.

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -603,6 +658,7 @@ def render_request(
         security: t.List[t.Dict[str, t.List[str]]] = method_spec.get(
             "security", self.api_spec.get("security", {})
         )
+        auth: t.Optional[t.Union[tuple[str, str], requests.auth.AuthBase]] = None
```
mdellweg, line 661: Is this needed? Why?

pulpcore/cli/common/generic.py
```diff
@@ -231,6 +232,19 @@ def basic_auth(self) -> t.Optional[t.Union[t.Tuple[str, str], requests.auth.Auth
                 self.pulp_ctx.password = click.prompt("Password", hide_input=True)
         return (self.pulp_ctx.username, self.pulp_ctx.password)
 
+    def auth(self, flow: t.Any) -> t.Optional[requests.auth.AuthBase]:
```
mdellweg, line 235: Still the wrong name here, but it looks about right being that easy, no?

mdellweg (dismissed):

pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -44,6 +49,28 @@ class UnsafeCallError(OpenAPIError):
...
+        if self.type == "oauth2":
+            self.flows: OAuth2Flows = self.security_scheme.get("flows")
+            client_credentials: ClientCredentials = self.flows["clientCredentials"]
+            if client_credentials:
+                self.flow_type: t.Optional[str] = "clientCredentials"
+                self.token_url: str = client_credentials["tokenUrl"]
+                self.scopes: OAuth2FlowsScopes = list(client_credentials.get("scopes").keys())
+
+        if self.type == "http":
+            self.scheme = self.security_scheme["scheme"]
```
mdellweg, line 71: This is all kind of hazy.
Can you explain why the logic didn't fit in the authproviderbase call method?

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+import requests
+
+
+class OAuth2ClientCredentialsAuth(requests.auth.AuthBase):
+    def __init__(
+        self,
+        client_id: str = "",
+        client_secret: str = "",
+        token_url: str = "",
+        scopes: t.List[str] = [],
```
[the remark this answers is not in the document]
mdellweg: A link to the specification would be good too.

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+        *args: t.List[t.Any],
+        **kwargs: t.Dict[t.Any, t.Any],
+    ):
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.token_url = token_url
+        self.scopes = scopes
+        self.token: t.Dict[t.Any, t.Any] = {}
```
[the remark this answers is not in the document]
mdellweg: It looks like the token combined with metadata.
So really token_info.

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __call__(self, request: requests.PreparedRequest) -> requests.PreparedRequest:
+        self.retrieve_local_token()
+
+        access_token = self.token.get("access_token")
+        request.headers["Authorization"] = f"Bearer {access_token}"
```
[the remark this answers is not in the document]
mdellweg: get can return None. So the server will fail when you could have failed before sending it.

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+
+        request.register_hook("response", self.handle401)  # type: ignore
+
+        return request
+
+    def handle401(
+        self, response: requests.Response, **kwargs: t.Dict[t.Any, t.Any]
```
mdellweg, line 37: This is not how you annotate kwargs:
```suggestion
        self, response: requests.Response, **kwargs: t.Any
```

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+            "scope": " ".join(self.scopes),
+            "grant_type": "client_credentials",
+        }
+
+        response: requests.Response = requests.post(self.token_url, data=data)
+
+        self.token = response.json() if response.status_code == 200 else None
+        self.token["issued_at"] = datetime.now().isoformat()
```
mdellweg, line 93: This is the point where you add the lifetime to only save the expiration date.
Also there's no use in formatting the datetime as a string.

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def retrieve_token(self) -> None:
+        data = {
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+            "scope": " ".join(self.scopes),
+            "grant_type": "client_credentials",
+        }
+
+        response: requests.Response = requests.post(self.token_url, data=data)
```
mdellweg, line 90: This response needs a throw_for_status (or similar).
After that there is no need to check for 200 anymore.

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def handle401(
+        self, response: requests.Response, **kwargs: t.Dict[t.Any, t.Any]
+    ) -> requests.Response:
+        if response.status_code != 401:
+            return response
+
+        self.retrieve_local_token()
+        if self.is_token_expired():
+            self.retrieve_token()
```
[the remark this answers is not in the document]
mdellweg: Talking to Grant we established that at this point it's safe to just unhook the hook from the copied request.

pulp-glue/pulp_glue/common/authentication.py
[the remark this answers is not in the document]
decko: > I don't think any of these can reasonably have a default, can they?

Nope. :(

pulp-glue/pulp_glue/common/openapi.py
decko: Here's the thing: we will have to add a new method every time we want to add a new authentication method here? With this aren't we changing the interface of this class?

pulp-glue/pulp_glue/common/openapi.py
decko: Maybe mypy was complaining about it ages ago. Removed.

pulp-glue/pulp_glue/common/authentication.py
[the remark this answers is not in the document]
decko: Yeah, but here's the thing. The first request will always fail, since we don't have the token yet.
Then we will request one on the `handle401` method. 
Make sense to just call it once, and if it doesn't work we just finish it and call it a day?

dkliban:
I accidentally used wrong credentials and got the following traceback:

```
Traceback (most recent call last):
  File "/home/dkliban/insights-venv/bin/pulp", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulpcore/cli/common/generic.py", line 326, in invoke
    return super().invoke(ctx)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulpcore/cli/common/generic.py", line 326, in invoke
    return super().invoke(ctx)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulpcore/cli/common/generic.py", line 326, in invoke
    return super().invoke(ctx)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulpcore/cli/common/generic.py", line 326, in invoke
    return super().invoke(ctx)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)   
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^   
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/decorators.py", line 92, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/decorators.py", line 92, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulpcore/cli/common/generic.py", line 1289, in callback
    result = entity_ctx.list(limit=limit, offset=offset, parameters=kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulp-glue/pulp_glue/common/context.py", line 807, in list
    result: t.Mapping[str, t.Any] = self.call("list", parameters=payload)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulp-glue/pulp_glue/common/context.py", line 609, in call
    return self.pulp_ctx.call(
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulp-glue/pulp_glue/common/context.py", line 375, in call
    result = self.api.call(
             ^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulp-glue/pulp_glue/common/openapi.py", line 778, in call
    response: requests.Response = self._session.send(request)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/requests/sessions.py", line 710, in send
    r = dispatch_hook("response", hooks, r, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/insights-venv/lib64/python3.12/site-packages/requests/hooks.py", line 30, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dkliban/devel/pulp-cli/pulp-glue/pulp_glue/common/authentication.py", line 44, in handle401
    self.retrieve_token()
  File "/home/dkliban/devel/pulp-cli/pulp-glue/pulp_glue/common/authentication.py", line 93, in retrieve_token
    self.token["issued_at"] = datetime.now().isoformat()
    ~~~~~~~~~~^^^^^^^^^^^^^
TypeError: 'NoneType' object does not support item assignment

```

pulp-glue/pulp_glue/common/openapi.py
mdellweg: That'd be an additive change.

pulp-glue/pulp_glue/common/authentication.py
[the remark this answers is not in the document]
mdellweg: I meant that even `""` is probably not reasonable default, so there should not be any.

pulp-glue/pulp_glue/common/authentication.py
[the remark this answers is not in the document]
mdellweg: Then again, maybe all we need to persist is the token and it expiry date.

pulp-glue/pulp_glue/common/authentication.py
```diff
@@ -0,0 +1,94 @@
...
+
+        response.content
+        prep = response.request.copy()
+
+        access_token = self.token.get("access_token")
+        prep.headers["Authorization"] = f"Bearer {access_token}"
+
+        _response: requests.Response = response.connection.send(prep, **kwargs)  # type: ignore
```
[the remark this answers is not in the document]
mdellweg: Here's an example.
https://github.com/pulp/pulp-cli/blob/56aefa9c1d1b71a928387e4476a415d7d2403dea/pulpcore/cli/rpm/content.py#L297
It helps a lot to judge if the ignore is warranted in the first place.

pulp-glue/pulp_glue/common/openapi.py
decko: Does the AuthProviderBase need to understand how to properly parse an OpenAPI SecurityScheme for each of those authentication methods? I was thinking on abstracting this to AuthProviderBase and exposes the minimum needed to choose which method to call.

pulp-glue/pulp_glue/common/authentication.py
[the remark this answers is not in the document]
decko: Done

decko:
Thanks @dkliban. We worked on that case and CLI is handling it fine now. Thanks.

ggainey (approved):
Review suggests we really want to refactor how AuthBase deals with handling/prioritizing multiple-arbitrary-schemes. However, that will have to wait. The PR in its current state answers a pressing existing need, and is "correct enough" to be useful. Approving, while recognizing this is truly "tech preview" and we are likely going to make significant changes in this code flow.

PR #1022, opened 2024-07-25
Fix return value of create

mdellweg:
When creating entities via tasks, we need to select one from the created_resources. This will use the HREF_PATTERN to find the right one.


pulp-glue/pulp_glue/common/context.py
```diff
@@ -903,8 +903,22 @@ def create(
...
+                            (
+                                h
+                                for h in result["created_resources"]
+                                if re.match(
+                                    re.escape(self.pulp_ctx.api_path) + self.HREF_PATTERN, h
+                                )
+                            )
+                        )
+                    else:
+                        self.pulp_href = result["created_resources"][0]  # YOLO
```
gerrod3, line 918: YOLO indeed

gerrod3 (approved):

PR #1027, opened 2024-07-29
Validate copy

mdellweg:


ggainey (approved):
Approving - this is aimed at solving https://github.com/pulp/pulp-oci-images/actions/runs/10172388150/job/28135537940#step:4:2029

PR #1051, opened 2024-08-27
Refactor the auth mechanism selection algorithm

mdellweg:
This fixes some regressions for wrongly identified auth mechanisms when parsing the security_schemes from the openapi document. Also the whole selection decision tree is mapped onto a better function call dependency chain.

### DIFF
--- a/pulp-glue/pulp_glue/common/openapi.py
+++ b/pulp-glue/pulp_glue/common/openapi.py
@@ -6,6 +6,7 @@ import datetime
 import json
 import os
 import typing as t
+from collections import defaultdict
 from contextlib import suppress
 from io import BufferedReader
 from urllib.parse import urljoin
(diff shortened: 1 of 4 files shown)


pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -78,61 +54,89 @@ class AuthProviderBase:
...
-        if "oauth2" in authorized_schemes_types:
-            for flow in authorized_schemes:
-                if flow["type"] == "oauth2":
-                    oauth2_flow = OpenAPISecurityScheme(flow)
-
-            # We "know" we'll have an outh2-flow here
-            if oauth2_flow.flow_type == "client_credentials":
-                result = self.oauth2_client_credentials_auth(oauth2_flow)
+            if len(proposal) == 0:
+                # Empty proposal: No auhentication needed. Shortcut return.
```
ggainey, line 106:
```suggestion
                # Empty proposal: No authentication needed. Shortcut return.
```

ggainey (changes requested):

ggainey (approved):

PR #1053, opened 2024-09-01
Test cert setup

mdellweg:


.ci/run_container.sh
```diff
@@ -60,10 +61,16 @@ else
...
   PORT="443"
   PULP_CONTENT_ORIGIN="https://localhost:8080/"
-  python3 -m trustme -d "${TMPDIR}/settings/certs"
-  export PULP_CA_BUNDLE="${TMPDIR}/settings/certs/client.pem"
-  ln -fs server.pem "${TMPDIR}/settings/certs/pulp_webserver.crt"
-  ln -fs server.key "${TMPDIR}/settings/certs/pulp_webserver.key"
+  python3 "${BASEPATH}/gen_certs.py" -d "${PULP_CLI_TEST_TMPDIR}/settings/certs"
+  export PULP_CA_BUNDLE="${PULP_CLI_TEST_TMPDIR}/settings/certs/ca.pem"
+  ln -fs server.pem "${PULP_CLI_TEST_TMPDIR}/settings/certs/pulp_webserver.crt"
+  ln -fs server.key "${PULP_CLI_TEST_TMPDIR}/settings/certs/pulp_webserver.key"
```
gerrod3, line 67: I'm confused on how these links find the right files. I though the server.pem and server.key are being stored in `PULP_CLI_TEST_TMPDIR/settings/certs`, what directory are the `ln` commands running from?

.ci/settings/settings.py
```diff
@@ -1,4 +1,3 @@
 ALLOWED_EXPORT_PATHS = ["/tmp"]
-ORPHAN_PROTECTION_TIME = 0
```
gerrod3, line 2: Are we sure this isn't going to break some tests?

gerrod3 (commented):
Some questions before I approve.

.ci/settings/settings.py
mdellweg: Sure? Well we put it in a loooong time ago as a stop gap. I don't think we do that in the repositories CI anymore.

gerrod3 (approved):

PR #1085, opened 2024-09-20
WIP Remove python-toml dependency

### LINKED ISSUE #1084: Make sure any dependencies are also available as RPMs?
- [ ] python-toml is a deprecated package and must not be depeneded on.

mdellweg:
Python 3.11 ships tomllib instead. For older versions we can use tomli with the same interface, and tomli-w allows dumping toml files.

fixes #1084


mdellweg:
Backporting this is probably a bad idea. It introduces a new dependency.

gerrod3 (approved):

PR #1106, opened 2024-11-06
Refactor internals of openapi class

mdellweg:
This will mask attributes of the class private and add a bit of abstraction from requests.

### DIFF
--- a/pulp-glue/pulp_glue/common/openapi.py
+++ b/pulp-glue/pulp_glue/common/openapi.py
@@ -310,7 +310,7 @@ class OpenAPI:
             param_spec = {k: v for k, v in param_spec.items() if v.get("required", False)}
         return param_spec
 
-    def extract_params(
+    def _extract_params(
         self,
         param_in: str,
         path_spec: t.Dict[str, t.Any],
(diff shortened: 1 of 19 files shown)


pulp-glue/pulp_glue/common/openapi.py
```diff
@@ -252,24 +260,26 @@ def _parse_api(self, data: bytes) -> None:
...
     def _download_api(self) -> bytes:
         try:
-            response: requests.Response = self._session.get(urljoin(self.base_url, self.doc_path))
+            response: requests.Response = self._session.get(urljoin(self._base_url, self._doc_path))
         except requests.RequestException as e:
             raise OpenAPIError(str(e))
         response.raise_for_status()
-        if "Correlation-ID" in response.headers:
-            self._set_correlation_id(response.headers["Correlation-ID"])
+        if "Correlation-Id" in response.headers:
```
ggainey, line 267: Having recently been burned by this - is there a good way to make this a case-insensitive comparison?

ggainey (commented):
Looks good - the "can we do case-insensitive?" comment is really just asking for clarification - if "we" own both sides of the conversation, we should be fine. But the spec allows for headers to be not-case-exact, yeah?

pulp-glue/pulp_glue/common/openapi.py
mdellweg: I'll look into CIMultiDict.

mdellweg: Instead of reworking it here, i decided to add some more ready commits that already included this improvement.

ggainey (approved):

PR #1118, opened 2025-01-17
Update base_url in test if content_origin is null

git-hyagi:
If `content_origin` is not set, the `distribution.base_url` will be defined with the relative path and the curl command in the test script will fail to run because of the missing hostname/address.

ref: https://github.com/pulp/pulpcore/pull/6175

### DIFF
--- a/tests/scripts/pulp_file/test_distribution.sh
+++ b/tests/scripts/pulp_file/test_distribution.sh
@@ -59,6 +59,13 @@ fi
 expect_succ pulp file distribution list --base-path "cli_test_file_distro"
 test "$(echo "$OUTPUT" | jq -r length)" -eq 1
 base_url="$(echo "$OUTPUT" | jq -r .[0].base_url)"
+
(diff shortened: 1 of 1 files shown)


tests/scripts/pulp_file/test_distribution.sh
```diff
@@ -59,6 +59,13 @@ fi
 expect_succ pulp file distribution list --base-path "cli_test_file_distro"
 test "$(echo "$OUTPUT" | jq -r length)" -eq 1
 base_url="$(echo "$OUTPUT" | jq -r .[0].base_url)"
+
+# if base_url starts with "/" (relative path) we need to prepend it with the server address
+if [[ $base_url == /* ]]
+then
+  base_url="http://localhost:8080${base_url}"
```
mdellweg, line 66: It's probably better to use the base url from the config instead of a hardcoded address here.
OTOH, we are not testing the cli with it. We should just remove this part.

PR #1123, opened 2025-01-22
Remove aspects of tests not testing the cli

mdellweg:
Apparently we can no longer assume that base_url is an url. But also this really is not testing CLI behaviour and should not be here in the first place.

### DIFF
--- a/tests/scripts/pulp_file/test_distribution.sh
+++ b/tests/scripts/pulp_file/test_distribution.sh
@@ -13,18 +13,6 @@ cleanup() {
 }
 trap cleanup EXIT
 
-if [ "$VERIFY_SSL" = "false" ]
-then
-  curl_opt=("-k")
-else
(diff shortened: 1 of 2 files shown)


tests/scripts/pulp_file/test_distribution.sh
```diff
@@ -58,17 +46,8 @@ fi
 
 expect_succ pulp file distribution list --base-path "cli_test_file_distro"
 test "$(echo "$OUTPUT" | jq -r length)" -eq 1
-base_url="$(echo "$OUTPUT" | jq -r .[0].base_url)"
-
-# if base_url starts with "/" (relative path) we need to prepend it with the server address
-if [[ $base_url == /* ]]
-then
-  base_url="https://pulp:443${base_url}"
```
mdellweg, line 66: Hardcoding a server address in these tests was never going to work properly.

ggainey (approved):

PR #1142, opened 2025-03-07
Fix headers with special characters in config

mdellweg:
This adjusts the regular expression for headers to allow all characters in the value. It also adds the same validation to the header option and merges headers from config and options.

### DIFF
--- a/CHANGES/1139.bugfix
+++ b/CHANGES/1139.bugfix
@@ -1,0 +1,2 @@
+Fix a bug when header values weren't allowed to contain other than alphanumeric characters.
+Also combine the headers from setting and the commandline.
--- a/pulp_cli/config.py
+++ b/pulp_cli/config.py
@@ -47,6 +47,21 @@ SETTINGS = [
     "verbose",
     "plugins",
 ]
+HEADER_REGEX = r"^[-a-zA-Z0-9_]+:.+$"
+
+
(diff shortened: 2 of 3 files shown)


CHANGES/1139.bugfix
```diff
@@ -0,0 +1,2 @@
+Fix a bug when header values contain more than alphanumeric characters.
```
ggainey, line 1:
```suggestion
Fix a bug when header values weren't allowed to contain other than alphanumeric characters.
```

pulp_cli/config.py
```diff
@@ -47,6 +47,20 @@
...
     "plugins",
 ]
+HEADER_REGEX = r"^[-a-zA-Z0-9_]+:.+$"
+
+
+def headers_callback(
+    ctx: click.Context, param: click.Parameter, value: t.Iterable[str]
+) -> t.Iterable[str]:
+    if not all((re.match(HEADER_REGEX, item) for item in value)):
+        raise click.BadParameter("format is <header-name>:<value>")
```
ggainey, line 57:
```suggestion
        raise click.BadParameter(f"format requires <header-name>:<value> : {item}")
```

What if we do this?

ggainey (changes requested):

ggainey (approved):

PR #1151, opened 2025-03-19
Move pulpcore.cli.common.generic to pulp_cli.generic

mdellweg:
The idea of using "pulpcore" as an open namespace inspired by the python bindings was not a good one after all. Specifically it collides with the way we build our docs. And due to the use of entrypoints it's not even needed here. The symbols are imported at the old location for compatibility for now.


mdellweg:
I expect to move more modules before removing the compatibility paths.

ggainey (approved):

PR #1182, opened 2025-04-30
Add support for new replicate features

gerrod3:


pulpcore/cli/core/upstream_pulp.py
```diff
@@ -4,6 +4,7 @@
 from pulp_glue.core.context import PulpUpstreamPulpContext
 
 from pulp_cli.generic import (
+    PluginRequirement,
```
mdellweg, line 7: This should come from `pulp_glue.common.context`.

PR #1184, opened 2025-05-07
a basic howto on creating a pulp-cli module

YasenT:
I went through the exercise of creating a separate pulp-cli module. And this is some basic howto. Does it make sense as a structure and text?


mdellweg:
I'm sorry you had to go through this.
I wish you had asked upfront and learned how to use the cookiecutter templates we have in place.
In any case this document describes a way we don't want to teach developers and also is misleading by containing too many wrong conclusions.
I appreciate the effort and agree on the benefit of a plugin writing guide.

PR #1188, opened 2025-05-12
Remove two small just-added tests

dralley:
We decided on a different behavior, no-op instead of outright failure.


.ci/scripts/check_click_for_mypy.py
```diff
@@ -1,9 +1,12 @@
 #!/bin/env python3
 
-import click
+from importlib.metadata import version
+
 from packaging.version import parse
 
-if parse(click.__version__) < parse("8.1.1") or parse(click.__version__) >= parse("8.2"):
```
dralley, line 6: click 8.2 was just released, our pyproject.toml only requests for <9, so this fails now

dralley:
Not sure what's up with the unittests.  Maybe the recent click update broke them too.

.ci/scripts/check_click_for_mypy.py
dralley: tbf they didn't actually remove it, but the warning alone caused our CI to fail

mdellweg (approved):

PR #1230, opened 2025-08-14
Fix cookiecutter's application of the test-matrix for plugins

gerrod3:
This should properly apply pulp-cli's test-matrix to the plugins.


gerrod3:
Not the right change.

PR #1241, opened 2025-09-10
Add ability to run tests in parallel [PULP-753]

gerrod3:
Need to figure out how to handle tests that call orphan cleanup and make sure they run last and in serial. Most of this is ensuring each test uses unique names to prevent collisions.


pulpcore/cli/container/distribution.py
```diff
@@ -115,6 +116,10 @@ def update(
 
     distribution: EntityDefinition = distribution_ctx.entity
     body: EntityDefinition = {}
+    non_blocking = not (
+        base_path is not None
+        or distribution_ctx.pulp_ctx.has_plugin(PluginRequirement("core", specifier="<3.24.0"))
+    )
```
mdellweg: This sounds like 
```suggestion
    non_blocking = (
        base_path is None
        and distribution_ctx.pulp_ctx.has_plugin(PluginRequirement("core", specifier=">=3.24.0"))
    )
```

What changed in that version?

PR #1249, opened 2025-09-23
when no repo, use synchronous upload for rpm

YasenT:
Extending the rpm uploads to use synchronous endpoint

### DIFF
--- a/pulpcore/cli/rpm/content.py
+++ b/pulpcore/cli/rpm/content.py
@@ -389,14 +389,17 @@ def upload(
                 _("You must specify one (and only one) of --file or --directory.")
             )
 
-        # Sanity: If directory, repository required
+        # Sanity: If using temp repository, both directory and repository are required
         final_dest_repo_ctx = kwargs["repository"]
-        if directory and not final_dest_repo_ctx:
+        if kwargs["use_temp_repository"] and not (directory and final_dest_repo_ctx):
             raise click.ClickException(
-                _("You must specify a --repository to use --directory uploads.")
+                _(
+                    "You must specify both --directory and --repository "
+                    "to use --use-temp-repository."
+                )
             )
 
-        # Sanity: ignore publish|use_temp unless directory has been specified
+        # Sanity: ignore publish|use_temp unless destination-repository has been specified
         use_tmp = final_dest_repo_ctx and kwargs["use_temp_repository"]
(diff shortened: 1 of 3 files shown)


pulp-glue/pulp_glue/common/context.py
```diff
@@ -1632,21 +1632,34 @@ def upload(
...
+                    body["upload"] = upload_href
+                else:
+                    from pulp_glue.core.context import PulpArtifactContext
+
+                    artifact_href = PulpArtifactContext(self.pulp_ctx).upload(file, chunk_size)
+                    body["artifact"] = artifact_href
+
+        # For rpm plugin >= 3.32.5, use synchronous upload endpoint when no repository is provided
+        # For older versions, always use the create endpoint (backward compatibility)
+        if repository is None and self.pulp_ctx.has_plugin(
```
ggainey: This is in common-context and be called with any kind-of PulpRepositoryContext. All you're checking here is that the pulp-instance being talked-to has rpm>=3.32.5. What if I'm invoking this from, say, `pulp python content upload`? You need to check the kind-of RepositoryContext to see that it's for a plugin that supports the new call *in addition to* making sure that plugin is the right version or higher.

YasenT: So from all the plugins, just 2x had issues
pulp-file, which doesn't seem to support the synchronous uploads, so i am overwriting this context in the file context.
And rpm which had an issue, which is resolved in the latest version of the plugin
The rest are passing the tests just fine, their synchronous implementations seems to be older and working as expected with the updated context.

pulpcore/cli/rpm/content.py
```diff
@@ -416,17 +419,20 @@ def upload(
...
+                if use_tmp and dest_repo_ctx:
                     result = _copy_to_final(dest_repo_ctx, final_dest_repo_ctx, pulp_ctx)
             finally:
                 if use_tmp and dest_repo_ctx:
                     # Remove the tmp-upload repo
                     dest_repo_ctx.delete()
 
-        final_version_number = _latest_version_number(final_dest_repo_ctx)
+        final_version_number = (
+            _latest_version_number(final_dest_repo_ctx) if final_dest_repo_ctx else None
```
ggainey: Look at _latest_version_number() - it already does this.

PR #1289, opened 2025-12-17
Add support for pulp-python attestations feature

gerrod3:


pulpcore/cli/python/remote.py
```diff
@@ -98,6 +98,12 @@ def remote(ctx: click.Context, pulp_ctx: PulpCLIContext, /, remote_type: str) ->
         callback=load_json_callback,
         needs_plugins=[PluginRequirement("python", specifier=">=3.2.0")],
     ),
+    pulp_option(
+        "--provenance",
+        type=click.BOOL,
```
mdellweg: Shouldn't this be "--provenance/--no-provenance" with default=None,
so it uses server default values and does not change the value without being specified?

tests/scripts/pulp_python/test_content.sh
```diff
@@ -33,7 +33,17 @@ fi
 
 expect_succ pulp python repository create --name "cli_test_python_repository"
 HREF="$(echo "$OUTPUT" | jq -r '.pulp_href')"
-expect_succ pulp python repository content add --repository "cli_test_python_repository" --sha256 "$sha256" --filename "shelf-reader-0.1.tar.gz"
-expect_succ pulp python repository content add --repository "$HREF" --sha256 "$sha256" --filename "shelf-reader-0.1.tar.gz" --base-version 0
-expect_succ pulp python repository content remove --repository "cli_test_python_repository" --sha256 "$sha256" --filename "shelf-reader-0.1.tar.gz"
-expect_succ pulp python repository content remove --repository "$HREF" --sha256 "$sha256" --filename "shelf-reader-0.1.tar.gz" --base-version 1
+expect_succ pulp python repository content add --repository "cli_test_python_repository" --sha256 "$sha256"
```
mdellweg: So pulp_python reads the filename (better package name and version) from the artifact now? :+1:

mdellweg (approved):

PR #1292, opened 2026-01-06
Switch linting and formatting to ruff

mdellweg:


ggainey (approved):
Let's see how this works for the CLI, before we turn it on for the rest of Pulp!

PR #1306, opened 2026-02-03
Disable chunk uploads for the sync rpm upload endpoint to improve performance on console

YasenT:
Disable chunk uploads
And retrigger tests to see if it fails at same place. As I see that the nighties run just fine


jobselko:
@YasenT I just re-ran the tests. Could you please add a changelog entry, since this changes the functionality?

mdellweg (approved):
Close enough. Thank you!

PR #1318, opened 2026-02-18
Add a chunk_size setting

mdellweg:
Users experiencing issues when uploading huge files can now define a default chunk size to use per server profile.

### DIFF
--- a/pulp-glue/src/pulp_glue/core/context.py
+++ b/pulp-glue/src/pulp_glue/core/context.py
@@ -45,12 +45,15 @@ class PulpArtifactContext(PulpEntityContext):
     ID_PREFIX = "artifacts"
 
     def upload(
-        self, file: t.IO[bytes], chunk_size: int = 1000000, sha256: str | None = None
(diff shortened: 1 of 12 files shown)


pulp-glue/src/pulp_glue/core/context.py
```diff
@@ -45,12 +45,15 @@ class PulpArtifactContext(PulpEntityContext):
...
+        self,
+        file: t.IO[bytes],
+        chunk_size: int | None = None,
+        sha256: str | None = None,
     ) -> t.Any:
         size = os.path.getsize(file.name)
 
         sha256_hasher = hashlib.sha256()
-        for chunk in iter(lambda: file.read(chunk_size), b""):
+        for chunk in iter(lambda: file.read(10_000_000), b""):
```
ggainey: Any particular reason for 10Mb here? Or is it just "a reasonable amount at a time" to hash?

ggainey (approved):

PR #1336, opened 2026-03-10
[PULP-1254] Serve Python content from a repository version

### LINKED ISSUE #1324: Add CLI support for serving python content form a repository version
The `pulp python distribution create` command should support specifying a repository version so that content can be served from that version instead of always using the latest repository version.

jobselko:
closes #1324

### DIFF
--- a/src/pulpcore/cli/python/distribution.py
+++ b/src/pulpcore/cli/python/distribution.py
@@ -39,8 +39,8 @@ repository_option = resource_option(
     default_type="python",
     context_table={"python:python": PulpPythonRepositoryContext},
     help=_(
-        "Repository to be used for auto-distributing."
-        " When set, this will unset the 'publication'."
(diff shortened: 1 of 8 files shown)


src/pulpcore/cli/python/distribution.py
```diff
@@ -41,6 +41,7 @@
     help=_(
         "Repository to be used for auto-distributing."
         " When set, this will unset the 'publication'."
+        " When 'version' is set, repository will no longer be auto-distributed."
         " Specified as '[[<plugin>:]<type>:]<name>' or as href."
```
mdellweg:
```suggestion
        "Repository to be distributed."
        " When used, this will replace any attached 'publication'."
        " Specified as '[[<plugin>:]<type>:]<name>' or as href."
```

mdellweg (approved):

PR #1348, opened 2026-04-03
Add pulp_rust CLI impl + tests

dralley:
Assisted-By: claude-opus-4.6

### DIFF
--- a/pulp-glue/src/pulp_glue/rust/context.py
+++ b/pulp-glue/src/pulp_glue/rust/context.py
@@ -1,0 +1,70 @@
+from pulp_glue.common.context import (
+    PluginRequirement,
+    PulpContentContext,
+    PulpDistributionContext,
+    PulpEntityContext,
+    PulpRemoteContext,
+    PulpRepositoryContext,
+    PulpRepositoryVersionContext,
+)
+from pulp_glue.common.i18n import get_translation
+
+translation = get_translation(__package__)
+_ = translation.gettext
+
+
+class PulpRustCargoTokenContext(PulpEntityContext):
+    ENTITY = _("cargo token")
+    ENTITIES = _("cargo tokens")
+    HREF = "rust_rust_cargo_token_href"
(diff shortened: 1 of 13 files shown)


src/pulpcore/cli/rust/content.py
```diff
@@ -0,0 +1,50 @@
...
+
+lookup_options = [
+    href_option,
+]
+
+content.add_command(
+    list_command(
+        decorators=[
+            click.option("--name"),
+            click.option("--vers"),
```
mdellweg: I'd prefer this to be called "version".

src/pulpcore/cli/rust/content.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@pulp_group()
+@click.option(
+    "-t",
+    "--type",
+    "content_type",
+    type=click.Choice(["rust"], case_sensitive=False),
+    default="rust",
+)
```
mdellweg: For less boilerplate, can you use the type_option generic here?

pulp-glue/src/pulp_glue/rust/context.py
```diff
@@ -0,0 +1,61 @@
...
+translation = get_translation(__package__)
+_ = translation.gettext
+
+
+class PulpRustContentContext(PulpContentContext):
+    PLUGIN = "rust"
+    RESOURCE_TYPE = "packages"
+    ENTITY = _("rust package")
+    ENTITIES = _("rust packages")
+    HREF = "rust_rust_package_content_href"
```
dralley, line 29: Presuming we don't cut a release immediately, can we still keep the option open to rename the models?would

The `package_content` bit is a bit weird, maybe something to do with https://github.com/pulp/pulp_rust/blob/main/pulp_rust/app/viewsets.py#L64

I'm working on some renamings here: https://github.com/pulp/pulp_rust/pull/32

PR #1354, opened 2026-04-09
Make widespead use of type_option

mdellweg:


src/pulpcore/cli/container/namespace.py
```diff
@@ -28,20 +28,10 @@
...
-    default="container",
-)
 @pass_pulp_context
 @click.pass_context
-def namespace(ctx: click.Context, pulp_ctx: PulpCLIContext, /, namespace_type: str) -> None:
-    if namespace_type == "container":
-        ctx.obj = PulpContainerNamespaceContext(pulp_ctx)
-    else:
-        raise NotImplementedError()
+def namespace(ctx: click.Context, pulp_ctx: PulpCLIContext, /) -> None:
```
ggainey, line 33: Doesn't this need to be-a @type_option?

ggainey (changes requested):

src/pulpcore/cli/container/namespace.py
mdellweg: container namespaces are not a one of the master-detail models.
This type option was a c'n'p artifact forever.

ggainey: Ahhhhh, OK, so `--type` being allowed here is a  bug, that we are fixing. 👍

ggainey (approved):

PR #1362, opened 2026-04-23
Apply ruff UP fixes

mdellweg:


pulp-glue/src/pulp_glue/common/i18n.py
```diff
@@ -1,10 +1,10 @@
 import gettext
-from functools import lru_cache
+from functools import cache
 from importlib.resources import files
 
 
 # Need to call lru_cache() before using it as a decorator for python 3.7 compatibility
```
ggainey: (unrelated to the PR - we can prob drop this comment now that we require py3.10+)

ggainey (dismissed):
Looks like it even caught a few subtle L10N bugs (i.e. changing _("".format()) to _("").format()

pulp-glue/src/pulp_glue/common/i18n.py
mdellweg: Specifically now that we don't use lru_cache at all anymore.

ggainey (approved):

PR #1370, opened 2026-04-26
[Demo] Add a justfile, roughly equivalent to the makefile

dralley:
I learned about this from pysequoia

Just is a command runner, it has many similarities to Make but does a few things differently.
https://github.com/casey/just

The main benefits are:

* `just --list` shows documentation for commands and groups
* Groups are runnable without needing to define new commands
* No .PHONY boilerplate
* multi-line scripts work properly, so `_inner` commands are not needed
* dependencies work better (see `setup-test-config`)
* `just` commands can be run while in a subdirectory, doesn't have to be the directory of the `justfile`

Downside is there is no equivalent to the i18n pattern rules so it becomes a script

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,125 @@
+LANGUAGES := "de"
+GLUE_PLUGINS := `ls pulp-glue/src/pulp_glue/ | tr '\n' ' '`
+CLI_PLUGINS := `ls src/pulpcore/cli/ | tr '\n' ' '`
+
+# Show discovered plugins
+info:
+    @echo "Pulp glue"
+    @echo "plugins: {{ GLUE_PLUGINS }}"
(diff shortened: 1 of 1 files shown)


mdellweg:
What is the equivalent of the "not .PHONY" targets then?

The nice thing (for me) about make is that it is basically a functional programming language for the filesystem.

dralley:
I believe it doesn't, because it's a command runner not a build tool.  At least I don't see an equivalent, though the docs mention that every recipe in just is equivalent to .PHONY

mdellweg:
Since we are actually using "proper Make targets" here, I conclude that we do not want to go forward with this idea.

Thank you for bringing it up anyway.

PR #1376, opened 2026-05-04
[PULP-1670] Add allow_package_substitution to Python repos

### LINKED ISSUE #1375: Add CLI support for allow_package_substitution on python repositories
The `allow_package_substitution` field was added to `PythonRepository` in pulp-python 3.28 but is not exposed in pulp-cli. Currently it can only be set via direct HTTP calls.

Python repositories allow package substitution by default - uploading, syncing, or adding a package with the same filename but a different checksum silently replaces the existing one. Setting `allow_package_substitution` to False rejects such operations, while re-adding a package with an identical checksum is always accepted.

jobselko:
closes #1375

### DIFF
--- a/src/pulpcore/cli/python/repository.py
+++ b/src/pulpcore/cli/python/repository.py
@@ -80,6 +80,15 @@ update_options = [
         needs_plugins=[PluginRequirement("python", specifier=">=3.3.0")],
         default=None,
     ),
+    pulp_option(
+        "--allow-package-substitution/--block-package-substitution",
+        needs_plugins=[PluginRequirement("python", specifier=">=3.28.0")],
+        default=None,
+        help=_(
+            "Allow replacing existing packages with packages of the same filename but"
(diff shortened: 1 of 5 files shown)


src/pulpcore/cli/python/repository.py
```diff
@@ -80,6 +80,16 @@ def repository() -> None:
...
         default=None,
     ),
+    pulp_option(
+        "--allow-package-substitution/--block-package-substitution",
+        needs_plugins=[PluginRequirement("python", specifier=">=3.28.0")],
+        default=None,
+        help=_(
+            "Allow replacing existing packages with packages of the same filename but"
+            " different checksum. When blocked, such operations are rejected."
+            " [default: enabled]"
```
mdellweg: I think, this is misleading. (As always, when not specified, the cli does not change the value or let's the server decide about the default.) Not mentioning this at all is the safest bet.
```suggestion
```

mdellweg (approved):

PR #1392, opened 2026-05-07
[PULP-1693] Add package blocklist to Python repos

### LINKED ISSUE #1391: Add CLI support for package blocklist on python repositories
A package blocklist functionality for PythonRepository was added in pulp-python 3.29 but is not exposed in pulp-cli. Currently it can only be managed via direct HTTP calls.

A repository blocklist prevents specific packages from being added, matching by package name (all versions), name with an exact version, or exact filename. Each entry records the creating user's PRN in the added_by field.

jobselko:
closes #1391

### DIFF
--- a/pulp-glue/src/pulp_glue/python/context.py
+++ b/pulp-glue/src/pulp_glue/python/context.py
@@ -4,7 +4,9 @@ from pulp_glue.common.context import (
     EntityDefinition,
     PluginRequirement,
     PulpContentContext,
+    PulpContext,
     PulpDistributionContext,
+    PulpEntityContext,
     PulpPublicationContext,
     PulpRemoteContext,
     PulpRepositoryContext,
@@ -165,3 +167,24 @@ class PulpPythonRepositoryContext(PulpRepositoryContext):
     def repair_metadata(self) -> t.Any:
         self.needs_capability("repair_metadata")
         return self.call("repair_metadata", parameters={self.HREF: self.pulp_href})
+
+
+class PulpPythonBlocklistEntryContext(PulpEntityContext):
+    ENTITY = _("blocklist entry")
+    ENTITIES = _("blocklist entries")
+    HREF = "python_python_python_blocklist_entry_href"
+    ID_PREFIX = "repositories_python_python_blocklist_entries"
+    NEEDS_PLUGINS = [PluginRequirement("python", specifier=">=3.30.0")]
+    repository_ctx: PulpPythonRepositoryContext
+
+    def __init__(
+        self,
+        pulp_ctx: PulpContext,
+        repository_ctx: PulpPythonRepositoryContext,
+    ) -> None:
+        super().__init__(pulp_ctx)
+        self.repository_ctx = repository_ctx
+
+    @property
+    def scope(self) -> dict[str, t.Any]:
+        return {self.repository_ctx.HREF: self.repository_ctx.pulp_href}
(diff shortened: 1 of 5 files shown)


pulp-glue/src/pulp_glue/python/context.py
```diff
@@ -165,3 +167,24 @@ def preprocess_entity(self, body: EntityDefinition, partial: bool = False) -> En
     def repair_metadata(self) -> t.Any:
         self.needs_capability("repair_metadata")
         return self.call("repair_metadata", parameters={self.HREF: self.pulp_href})
+
+
+class PulpPythonRepositoryBlocklistEntryContext(PulpEntityContext):
```
mdellweg:
```suggestion
class PulpPythonBlocklistEntryContext(PulpEntityContext):
```
I don't think the "repository" in there adds any information or helps distinguishing things.

src/pulpcore/cli/python/repository.py
```diff
@@ -152,6 +155,74 @@ def repository() -> None:
...
+) -> None:
+    """
+    Manage blocklist entries for a Python repository.
+    """
+    assert isinstance(repository_ctx, PulpPythonRepositoryContext)
+    ctx.obj = PulpPythonRepositoryBlocklistEntryContext(pulp_ctx, repository_ctx)
+
+
+blocklist.add_command(list_command(decorators=nested_lookup_options))
+blocklist.add_command(show_command(decorators=nested_lookup_options + [href_option]))
```
mdellweg: This doesn't look right. How do you identify a blocklist on a repository?

Can a repository have more than one blocklist?

src/pulpcore/cli/python/repository.py
```diff
@@ -152,6 +155,74 @@ def repository() -> None:
...
+blocklist.add_command(
+    destroy_command(name="remove", decorators=nested_lookup_options + [href_option])
```
mdellweg: This also looks like suboptimal ux.
You should be able to remove entries on the blocklist by specifying the name and version IMO.

src/pulpcore/cli/python/repository.py
jobselko: Right now the Python plugin API has no way to filter entries by `name`/`version`/`filename` (only href is supported). This will require a change to the plugin, so I am changing this PR to a draft.

A repository has always one implicit blocklist - `list` gives you all blocklist entries, `show` only the one you specify. Does it make sense?

mdellweg: > A repository has always one implicit blocklist - list gives you all blocklist entries, show only the one you specify. Does it make sense?

Yes, that makes sense. But still i am missing the "natural key interface" here.

So we have one blocklist (exactly per respository). And all we care about is listing all entries (maybe filtered/paginated) adding entries and removing entries. And maybe show an individual one, but there is nothing more to an entry than what the natural key `name/version/filename` already encodes.
I think we should really be able to say (paraphrasing):
`... blocklist repo1 block package1`
`... blocklist repo2 unblock package2 2.3.5`
`... blocklist repo3 list`
(apparently we need to turn these arguments in to cli options...)

jobselko: Currently, cli can be used this way:
```
pulp python repository blocklist add --repository foo --name pkg
pulp python repository blocklist remove --repository foo --name pkg
pulp python repository blocklist list --repository foo
```
I think `blocklist add` and `blocklist remove` (destroy?) are clear enough. `blocklist block` and `blocklist unblock` sounds quite repetitive.

mdellweg: Yes, I agree that looks like a good interface.

mdellweg (approved):

PR #1399, opened 2026-05-19
[PULP-1693] Fix blocklist lookup for name-only entries

jobselko:
follow-up to #1391

### DIFF
--- a/src/pulpcore/cli/python/repository.py
+++ b/src/pulpcore/cli/python/repository.py
@@ -30,7 +30,7 @@ from pulp_cli.generic import (
     list_command,
     lookup_callback,
     name_option,
-    pass_entity_context,
+    option_group,
     pass_pulp_context,
     pass_repository_context,
     pulp_group,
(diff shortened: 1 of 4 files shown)


src/pulpcore/cli/python/repository.py
```diff
@@ -181,33 +181,61 @@ def blocklist(
...
+    /,
+    name: str | None,
+    version: str | None,
+    filename: str | None,
+) -> None:
+    """
+    Show a blocklist entry.
+    """
+    assert isinstance(entity_ctx, PulpPythonBlocklistEntryContext)
+    lookup = _blocklist_lookup(name, version, filename)
```
mdellweg: This looks like you might want to use the `generic.option_group`.

jobselko: As discussed, not sure if this can help me

jobselko:
The failing CI test is unrelated - the nightly image was built before pulp_python 3.30.2 was released

mdellweg:
> The failing CI test is unrelated - the nightly image was built before pulp_python 3.30.2 was released

I disagree. It is absolutely related. Since the nightly image isn't yet recent enough to demonstrate this code working we should wait for it to land before proceeding here. (All the other runners just skip this test completely to begin with.)

jobselko:
@mdellweg You are right. I did not notice that all tests were skipped because only nightly was red.

mdellweg (approved):

src/pulpcore/cli/python/repository.py
jobselko: Reworked to use `generic.option_group`

PR #1401, opened 2026-05-20
Stringify objects when posting a form-encoded body

mdellweg:


pulp-glue/src/pulp_glue/common/openapi.py
```diff
@@ -421,7 +426,10 @@ def _render_request_body(
                 if content_type.startswith("application/json"):
                     data = encode_json(body)
                 elif content_type.startswith("application/x-www-form-urlencoded"):
-                    data = body
+                    if isinstance(body, dict):
+                        data = {k: encode_stringify(v) for k, v in body.items()}
+                    else:
+                        data = encode_param(body)
```
gerrod3, line 432: What happens if body is a list of dicts? Is this possible?

pulp-glue/tests/test_schema.py
```diff
@@ -8,6 +8,7 @@
 from pulp_glue.common import oas
 from pulp_glue.common.exceptions import SchemaError, ValidationError
 from pulp_glue.common.schema import (
+    encode_html,
```
gerrod3, line 11: Are you going to add a test for encode_stringify?

pulp-glue/src/pulp_glue/common/openapi.py
mdellweg: Not yet in our api. And I'd need to reread the specification in that case.

pulp-glue/tests/test_schema.py
mdellweg: I can, I guess...

gerrod3 (approved):

PR #1412, opened 2026-06-04
Allow setting `retain_repo_versions` to `null`

alexjfisher:
Setting is already `NULLABLE` in
`pulp-glue/src/pulp_glue/common/context.py` line 1586

```
NULLABLES = {"description", "remote", "retain_repo_versions"}
```

but cli didn't allow an empty string to be used.


mdellweg:
Please add a changelog entry to this bugfix.

alexjfisher:
> Please add a changelog entry to this bugfix.

Done!  (At least I think I've done it right! :) )

mdellweg (approved):

PR #1419, opened 2026-06-08
Allow filtering livetest runs via PYTEST_MARK

gerrod3:
## Summary
- Add a `PYTEST_MARK` make variable to the `livetest` target, defaulting to `live`
- Plugin CI can now run only the live tests for a single component, e.g. `make livetest PYTEST_MARK="live and pulp_container"`

## Test plan
- [ ] `make livetest` still runs the full live suite (default unchanged)
- [ ] `make livetest PYTEST_MARK="live and pulp_container"` runs only container live tests
- [ ] Update plugin CI `script.sh` to use the filtered invocation once this is released

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,9 +59,11 @@ _test: | tests/cli.toml
 test:
 	uv run $(MAKE) _test
 
+PYTEST_MARK ?= live
+
 .PHONY: _livetest
 _livetest: | tests/cli.toml
-	pytest -v tests pulp-glue/tests -m live
+	pytest -v tests pulp-glue/tests -m "$(PYTEST_MARK)"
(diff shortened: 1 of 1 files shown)


gerrod3:
The AI also suggested we could add this rule:

If you want shorter commands for common cases:

```
.PHONY: livetest-%
livetest-%:
	uv run $(MAKE) _livetest PYTEST_MARK="live and pulp_$*"
```

Then:

```
make livetest-container   # → live and pulp_container
make livetest-rpm         # → live and pulp_rpm
```

ggainey (approved):

PR #1437, opened 2026-06-17
Added support for --api-version to the CLI.

ggainey:
Helpful in talking to https://github.com/pulp/pulpcore/pull/7802 - this works but is likely "incomplete" - and certainly needs tests and such.

### DIFF
--- a/pulp-glue/src/pulp_glue/common/openapi.py
+++ b/pulp-glue/src/pulp_glue/common/openapi.py
@@ -728,7 +728,6 @@ class OpenAPI:
         rel_url = path
         for name, value in rendered_parameters["path"].items():
             rel_url = path.replace("{" + name + "}", value)
-
         query_params = rendered_parameters["query"]
 
         url = urljoin(self._base_url, rel_url)
--- a/src/pulp_cli/config.py
+++ b/src/pulp_cli/config.py
@@ -37,6 +37,7 @@ CONFIG_LOCATIONS = [
     str(Path(click.utils.get_app_dir("pulp"), "cli.toml")),
 ]
 FORMAT_CHOICES = list(REGISTERED_OUTPUT_FORMATTERS.keys())
+VERSIONS = ["v3", "v4"]
 REQUIRED_SETTINGS = {
     "base_url",
     "api_root",
@@ -57,6 +58,7 @@ OPTIONAL_SETTINGS = {
     "key",
     "chunk_size",
     "plugins",
+    "api_version",
 }
 SETTINGS = REQUIRED_SETTINGS | OPTIONAL_SETTINGS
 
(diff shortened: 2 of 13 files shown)


pulp-glue/src/pulp_glue/common/openapi.py
mdellweg, line 1: I would not expect this file to be affected at all.

src/pulp_cli/config.py
```diff
@@ -37,6 +37,7 @@
     str(Path(click.utils.get_app_dir("pulp"), "cli.toml")),
 ]
 FORMAT_CHOICES = list(REGISTERED_OUTPUT_FORMATTERS.keys())
+VERSIONS = ["v3", "v4", "v5"]
```
mdellweg: Is v5 for experimenting? Or really taking the future into account?

pulp-glue/src/pulp_glue/common/openapi.py
ggainey: Correct - remains of an earlier experiment.

src/pulp_cli/config.py
ggainey: "v5" is there for doing some what-ifs.  VERSIONS itself is a placeholder. Post-v3, the status-call will tell us what API versions are supported in a given instance, and we can reject 'anything else' based on what an initial status]call tells us. I haven't figured out the "right way" to do that, yet.

tests/scripts/pulpcore/test_task.sh
```diff
@@ -32,7 +32,7 @@ repository_href="$(echo "$OUTPUT" | jq -r '.pulp_href')"
 if pulp debug has-plugin --name "core" --specifier ">=3.21.0"
 then
   expect_succ pulp --background file repository sync --name "cli_test_core_task_repository" --remote "cli_test_core_task_large_remote"
-  task="$(echo "$ERROUTPUT" | grep -E -o "${PULP_API_ROOT}([-_a-zA-Z0-9]+/)?api/v3/tasks/[-[:xdigit:]]*/")"
+  task="$(echo "$ERROUTPUT" | grep -E -o "${PULP_API_ROOT}([-_a-zA-Z0-9]+/)?api/v[[:xdigit:]]+/tasks/[-[:xdigit:]]*/")"
```
mdellweg: Do you really mean hex-digit?

tests/scripts/pulpcore/test_task.sh
```diff
@@ -50,9 +50,9 @@ fi
 
 expect_fail pulp --dry-run task cancel --all
 
-# Test waiting for a task
+## Test waiting for a task
```
mdellweg: Probably unintended change...

mdellweg (approved):

PR #1440, opened 2026-07-07
Add pulp_ansible git remotes

quba42:
I am hoping it is easier to add this, than it is to talk directly to the API for more than 5 minutes...

### DIFF
--- a/CHANGES/+ansible_git_remote.feature
+++ b/CHANGES/+ansible_git_remote.feature
@@ -1,0 +1,1 @@
+Added support for pulp_ansible git remotes.
--- a/CHANGES/pulp-glue/+ansible_git_remote.feature
+++ b/CHANGES/pulp-glue/+ansible_git_remote.feature
@@ -1,0 +1,1 @@
+Added the pulp_ansible git remote context.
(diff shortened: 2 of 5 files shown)


quba42:
I have manually tested this change against the workflow described here: https://github.com/pulp/pulp_ansible/issues/2565

All commands using `pulp ansible remote --type git ...` that I have tried so far have worked.

mdellweg (approved):
If i read this correctly, it does not depend on the pulp_ansible pr, you opened, but it's supposed to help dealing with it.

quba42:
> If i read this correctly, it does not depend on the pulp_ansible pr, you opened, but it's supposed to help dealing with it.

You read that correctly. What this PR does is to expand the `pulp ansible remote --type` flag to accept the argument `git`. What my PR does is expand the functionality of pulp_ansible git remotes so they no longer ignore what the user supplied for "username" and "password". Pulp CLI does not have to care whether pulp_ansible ultimately does anything useful with the API parameters, it just needs to use the API interface correctly. Which it will do without my PR in pulp_ansible.

PR #1443, opened 2026-07-09
Add --distribution filter to container content list.

gerrod3:
Resolve the served repository version from a distribution so users can list tags and other content for an image without manual lookup steps.

Ref: https://github.com/pulp/pulp_container/issues/2164

### DIFF
--- a/src/pulpcore/cli/container/content.py
+++ b/src/pulpcore/cli/container/content.py
@@ -3,21 +3,31 @@ import typing as t
 
 import click
 
-from pulp_glue.common.context import PluginRequirement, PulpContentContext
+from pulp_glue.common.context import (
+    EntityDefinition,
+    PluginRequirement,
+    PulpContentContext,
+    PulpDistributionContext,
+)
 from pulp_glue.container.context import (
     PulpContainerBlobContext,
+    PulpContainerDistributionContext,
     PulpContainerManifestContext,
+    PulpContainerRepositoryContext,
     PulpContainerTagContext,
 )
 
 from pulp_cli.generic import (
+    PulpCLIContext,
     content_filter_options,
     href_option,
     label_command,
     list_command,
     option_group,
+    option_processor,
     pulp_group,
     pulp_option,
+    resource_option,
     show_command,
     type_option,
 )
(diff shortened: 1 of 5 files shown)


src/pulpcore/cli/container/content.py
```diff
@@ -34,6 +46,44 @@ def _content_callback(ctx: click.Context, value: dict[str, t.Any]) -> None:
...
+            _(
+                "Distribution '{name}' is not associated with a repository or repository version."
+            ).format(name=distribution.get("name", ""))
+        )
+    if "/container-push/" in repository_href:
+        repo_ctx: PulpRepositoryContext = PulpContainerPushRepositoryContext(
+            pulp_ctx, pulp_href=repository_href
+        )
+    else:
+        repo_ctx = PulpContainerRepositoryContext(pulp_ctx, pulp_href=repository_href)
```
mdellweg: The common base class wouldn't let us do this, right?

src/pulpcore/cli/container/distribution.py
```diff
@@ -127,18 +127,16 @@ def update(
             repository = t.cast(PulpEntityContext, repository)
             if version is not None:
                 if distribution["repository"]:
-                    distribution_ctx.update(body={"repository": ""}, non_blocking=non_blocking)
+                    distribution_ctx.update(body={"repository": ""}, non_blocking=False)
```
mdellweg: Does the server still not allow to update these in one go?

src/pulpcore/cli/container/content.py
```diff
@@ -34,6 +46,44 @@ def _content_callback(ctx: click.Context, value: dict[str, t.Any]) -> None:
...
+
+    return t.cast(str, repo_ctx.entity["latest_version_href"])
+
+
+def _process_distribution_filter(ctx: click.Context) -> None:
+    distribution = ctx.params.pop("distribution", None)
+    if not distribution:
+        return
```
mdellweg: How do you think about inverting the if here?
This currently looks like an if with only a disguised else branch.

mdellweg (commented):
You should rebase and probably squash the last two commits.

src/pulpcore/cli/container/distribution.py
gerrod3: I believe it does, just my AI agent still struggling to test the code against my dev env and trying random stuff till CI passes. 

Do we need the CLI to support pulpcore<3.24? I've removed the inbetween updates with the correct approach the API accepts.

mdellweg: No, we strive to support what is a supported branch. And that means >= 3.49 atm.

mdellweg: But we probably need to raise the guard in `pulp_glue.common.context.PulpContext.__init__` before we just drop the one here.

gerrod3: Should I bump it to 3.49 or to 3.24?

mdellweg: I started to look what "cleanup" to 3.49 looks like.
https://github.com/pulp/pulp-cli/pull/1445/

It should be good to go with either 3.24 or 3.49 here. Just let's not make it false claims at any point.

mdellweg (approved):

PR #1448, opened 2026-07-23
Add --version option to rpm and file distribution commands

dralley:
Allow setting repository_version on distributions for publication-based plugins (rpm, file), matching the capability added in pulpcore 3.106.0.

Assisted-By: Claude Opus 4.6

### DIFF
--- a/pulp-glue/src/pulp_glue/file/context.py
+++ b/pulp-glue/src/pulp_glue/file/context.py
@@ -12,6 +12,7 @@ from pulp_glue.common.context import (
     PulpRepositoryVersionContext,
     api_spec_quirk,
 )
+from pulp_glue.common.exceptions import PulpException
 from pulp_glue.common.i18n import get_translation
 
 translation = get_translation(__package__)
@@ -87,6 +88,23 @@ class PulpFileDistributionContext(PulpDistributionContext):
                 body["publication"] = None
             if "repository" not in body and "publication" in body:
                 body["repository"] = None
+
+        version = body.pop("version", None)
+        if version is not None:
+            self.pulp_ctx.needs_plugin(PluginRequirement("core", specifier=">=3.106.0"))
+            repository_href = body.pop("repository", None)
+            if repository_href is None and partial:
+                repository_href = self.entity.get("repository")
+            if repository_href is None:
+                raise PulpException(_("--repository must be provided"))
+            body["repository_version"] = f"{repository_href}versions/{version}/"
+            body["repository"] = None
+            body["publication"] = None
+        elif "repository" in body and self.pulp_ctx.has_plugin(
+            PluginRequirement("core", specifier=">=3.106.0")
+        ):
+            body["repository_version"] = None
+
         return body
 
 
--- a/pulp-glue/src/pulp_glue/rpm/context.py
+++ b/pulp-glue/src/pulp_glue/rpm/context.py
@@ -15,6 +15,7 @@ from pulp_glue.common.context import (
     PulpViewSetContext,
     api_spec_quirk,
 )
+from pulp_glue.common.exceptions import PulpException
 from pulp_glue.common.i18n import get_translation
 
 translation = get_translation(__package__)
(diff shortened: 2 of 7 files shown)


mdellweg (commented):
The cli option part looks nice and clean.
The preprocess_entity in glue has a lot of branches i need to trace for a verdict.
Do you think we should unittest this very function here?

But in any case can you please add a changelog entry, ideally one for cli and glue each?

pulp-glue/src/pulp_glue/rpm/context.py
```diff
@@ -81,11 +82,33 @@ class PulpRpmDistributionContext(PulpDistributionContext):
 
     def preprocess_entity(self, body: EntityDefinition, partial: bool = False) -> EntityDefinition:
         body = super().preprocess_entity(body, partial=partial)
+        if "version" in body:
+            self.pulp_ctx.needs_plugin(PluginRequirement("core", specifier=">=3.106.0"))
```
mdellweg: So `{"version": None}` is supposed to fail on pulp_rpm < 3.106.0?

I wonder if we should just move this statements after line 94.

pulp-glue/src/pulp_glue/rpm/context.py
```diff
@@ -81,11 +82,33 @@ class PulpRpmDistributionContext(PulpDistributionContext):
...
+        if version is not None:
+            repository_href = body.pop("repository", None)
+            if not repository_href and partial:
+                current_entity = self.entity
+                repository_href = current_entity.get("repository")
+                if not repository_href and (
+                    repository_version_href := current_entity.get("repository_version")
+                ):
+                    repository_href = f"{repository_version_href.rsplit('/', 3)[0]}/"
+            if not repository_href:
```
mdellweg: This could be
```python
if repository_href is None:
    if partial:
        ...
        repository_href = ...
    else:
        raise ...
```

pulp-glue/src/pulp_glue/rpm/context.py
```diff
@@ -81,11 +82,33 @@ class PulpRpmDistributionContext(PulpDistributionContext):
...
+                raise PulpException(_("--repository must be provided"))
+            body["repository_version"] = f"{repository_href}versions/{version}/"
+            body["repository"] = None
+        elif "repository" in body and self.pulp_ctx.has_plugin(
```
mdellweg, line 102: Is the intention here that when specifying the repository but not the version, the version should be nullified?
So we _need_ the `"repository" in body` part of that condition?

dralley: There's 2 fields in the CLI (repository and version) which controls which of the 3 fields on distribution get set.  If both are present then it sets the repository version, otherwise it sets the repository only.

pulp-glue/src/pulp_glue/rpm/context.py
dralley: Setting repository version on a distribution for a publication based plugin should fail if <3.106

dralley:
@mdellweg How's this?  I simplified it a bit and removed the ability to switch version without re-specifying repository.

Commits should be squashed before merging

pulp-glue/src/pulp_glue/file/context.py
```diff
@@ -87,6 +88,23 @@ def preprocess_entity(self, body: EntityDefinition, partial: bool = False) -> En
...
+                repository_href = self.entity.get("repository")
+            if repository_href is None:
+                raise PulpException(_("--repository must be provided"))
+            body["repository_version"] = f"{repository_href}versions/{version}/"
+            body["repository"] = None
+            body["publication"] = None
+        elif "repository" in body and self.pulp_ctx.has_plugin(
+            PluginRequirement("core", specifier=">=3.106.0")
+        ):
+            body["repository_version"] = None
```
mdellweg, line 106: Now that is kind of easy to follow.

pulp-glue/tests/test_distribution_preprocess.py
```diff
@@ -0,0 +1,107 @@
...
+    return PulpContext.from_config(settings)
+
+
+@pytest.mark.parametrize(
+    "ctx_cls",
+    [PulpRpmDistributionContext, PulpFileDistributionContext],
+    ids=["rpm", "file"],
+)
+class TestDistributionPreprocess:
+    def test_version_with_repository(self, modern_pulp_ctx: PulpContext, ctx_cls: type) -> None:
```
mdellweg: Don't they share a common parent class?

```suggestion
    def test_version_with_repository(self, modern_pulp_ctx: PulpContext, ctx_cls: t.Type[PulpDistributionContext]) -> None:
```

mdellweg (commented):
Just one tiny suggestion to try.

mdellweg (approved):

PR #1455, opened 2026-08-12
Update cookiecutter makefile for paralleltest new PYTEST_MARK

gerrod3:
I once again made a change to the Makefile without updating the cookiecutter.


mdellweg:
There's no use backporting cookiecutter changes. They should only apply from main anyway.

mdellweg (approved):

PR #1456, opened 2026-08-12
Note that Makefile is cookiecutter-managed

gerrod3:
This is the behavior we want, right?

### DIFF
--- a/cookiecutter/ci/{{ cookiecutter.__project_name }}/Makefile
+++ b/cookiecutter/ci/{{ cookiecutter.__project_name }}/Makefile
@@ -1,4 +1,7 @@
 {%- set src_infix = "src/" if cookiecutter.src_layout else "" %}
+# This file is managed by the cookiecutter templates.
+# Do not edit manually
+
 {%- if cookiecutter.translations %}
 LANGUAGES=de
 {%- endif %}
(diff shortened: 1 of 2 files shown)


cookiecutter/ci/{{ cookiecutter.__project_name }}/Makefile
```diff
@@ -1,4 +1,6 @@
 {%- set src_infix = "src/" if cookiecutter.src_layout else "" %}
+# This file is managed by cookiecutter. Any changes should originate in the template and be applied through cookiecutter/apply_templates.py.
```
mdellweg: We have a lot of this in the pyproject.toml, because it's rather unintuitive there.
`# This section is managed by the cookiecutter templates.`

I would keep the language similar:

```suggestion
# This file is managed by the cookiecutter templates.
```

or

```suggestion
# This file is managed by the cookiecutter templates.
# Do not edit manually.
```

Should we make that a macro and add to more files?
The whole .ci directory for example should be managed, but at the moment we don't delete additional files there.
The `.github/workflows/test.yaml` is extra-special. It is templated but keeps the test matrix intact.

mdellweg (approved):

PR #1464, opened 2026-08-20
Harden orphans test against parallel test running

gerrod3:
CLI tests on pulpcore are a bit flaky when running in parallel since pulpcore CI sets ORPHAN_PROTECTION_TIME to zero. This should help.

### DIFF
--- a/tests/scripts/pulpcore/test_orphans.sh
+++ b/tests/scripts/pulpcore/test_orphans.sh
@@ -4,9 +4,8 @@ set -eu
 # shellcheck source=tests/scripts/config.source
 . "$(dirname "$(dirname "$(realpath "$0")")")"/config.source
 
-expect_succ pulp orphan cleanup
-
-test "$(echo "${OUTPUT}" | jq -r '.state' )" = "completed"
+expect_fail pulp --dry-run orphan cleanup
+test "${ERROUTPUT}" = "Error: Call aborted due to safe mode"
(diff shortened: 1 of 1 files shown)


tests/scripts/pulpcore/test_orphans.sh
```diff
@@ -4,16 +4,28 @@ set -eu
...
 pulp debug has-plugin --name "file" || exit 23
 
 dd if=/dev/urandom of=test_1.txt bs=2MiB count=1
 
-expect_succ pulp file content upload --file test_1.txt --relative-path orphan_test/test_1.txt
+expect_succ pulp --domain "$DOMAIN" file content upload --file test_1.txt --relative-path orphan_test/test_1.txt
 content_href=$(echo "${OUTPUT}" | jq -r .pulp_href)
 
-expect_succ pulp orphan cleanup --content-hrefs "[\"$content_href\"]"
-expect_succ pulp orphan cleanup --protection-time 10
```
mdellweg, line 19: What is the issue? As these two commands are meant to be very careful about the stuff that gets cleaned up, are we sure this is not exposing an actual issue with cleanup on the server side?

What if instead, we stop configuring anything for this test, add zero protection time to the first and increase the number on the second to a day? (This test really does not care what the server does, other than responding correctly.)

mdellweg (approved):

